PR #9, opened 2019-03-20
fix dockerfile

phoracek:
Several bugs were introduced with multi-stage dockerfile PR. This PR should fix them.


phoracek:
@SchSeba could you please take a look? With this change, operator pod starts successfully.

PR #12, opened 2019-03-22
automatically detect multus configuration

phoracek:
Multus is not capable of creating its own config file with delegators
pointing to detected default network. Let's use this option instead
of requiring administrator to pass delegates explicitly.

With this change, testing on both OpenShift and Kubernetes will become
easier as we can use the same NetworkAddonsConfig on both. Also integration
with https://github.com/kubevirt/hyperconverged-cluster-operator should
be straightforward.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,52 +16,15 @@ kind: NetworkAddonsConfig
 metadata:
   name: cluster
 spec:
-  multus:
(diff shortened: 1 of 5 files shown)


phoracek:
@SchSeba @booxter could you please take a look?

PR #13, opened 2019-03-22
Fixed build warning

booxter:
cmd/manager/main.go:98: Println call has possible formatting directive %v


booxter:
I know this fix is part of another PR but we should not wait for it to polish.

phoracek:
LGTM

PR #14, opened 2019-03-22
SR-IOV support

booxter:


booxter:
@phoracek this is mostly ready but there are open questions to consider (see commit message).

phoracek:
> Open questions:
> - who creates sriov CRD?
For start, I'd stick to whatever we did before with kubevirt-ansible. How was it done there?
> - how do we manage heterogeneous environments with custom configs for
> each node?
Again, can we start with creating an empty default config per each node? Then we can try to figure out a better way. Any chance we can copy from openshift/cluster-network-operator?
> - we should probably reformat SRIOV_ROOT_DEVICES envvar to the format of
> .json config file instead of forcing users to list it in config format.
Sounds good

booxter:
> > Open questions:
> > 
> > * who creates sriov CRD?
> 
> For start, I'd stick to whatever we did before with kubevirt-ansible. How was it done there?
> 

`kubevirt-ansible` created it hardcoded. Should we do the same here?

> > * how do we manage heterogeneous environments with custom configs for
> >   each node?
> 
> Again, can we start with creating an empty default config per each node? Then we can try to figure out a better way. Any chance we can copy from openshift/cluster-network-operator?
> 

They use a different (older) version of the device plugin that doesn't require configuration. The problem here is that there is no way to modify manifest for ConfigMap carrying per node configuration file for the plugin. We never made it so that it would be easy to specify custom configuration files per node (it's not exposed via a config option.)

But at least admin could do it manually afterwards. My understanding is that the operator is different in that it will detect changes to objects it tracks and revert any local modifications back to where they were. Is it the case? If so, we may need to move the ConfigMap object from under operator control so that admins still have a way to apply per node configuration.

booxter:
The latest upload handles local config changes scenario; accepts a more user friendly format for SRIOV_ROOT_DEVICES envvar. It does not deploy CRD which will be in a separate PR. (We'll need to decide how to create CRD objects here, since there is no code for that just yet; and whether there are config knobs to expose via config resource / envvars.)

deploy/cluster-network-addons-operator_03_deployment.yaml
```diff
@@ -18,7 +18,7 @@ spec:
       containers:
       - name: cluster-network-addons-operator
         # Replace this with the built image name
-        image: quay.io/kubevirt/cluster-network-addons-operator
+        image: quay.io/booxter/cluster-network-addons-operator
```
phoracek: please revert this

pkg/network/sriov.go
```diff
@@ -0,0 +1,54 @@
...
+	"reflect"
+	"strings"
+
+	"github.com/kubevirt/cluster-network-addons-operator/pkg/render"
+	"github.com/pkg/errors"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+
+	opv1alpha1 "github.com/kubevirt/cluster-network-addons-operator/pkg/apis/networkaddonsoperator/v1alpha1"
+)
+
```
phoracek: One piece is missing, we should check for SRIOV against OpenShift network operator. In case user enabled SRIOV in network-addons operator, we need to check that it is indeed enabled also in OpenShift.

We already have this logic in Multus module.

deploy/cluster-network-addons-operator_03_deployment.yaml
booxter: Sorry, my bad. :)

pkg/network/sriov.go
booxter: @phoracek I think the assumption should be that their operator is NOT used for SR-IOV deployment, no? Because these two pieces are not compatible.

phoracek: What we do in Multus is: If openshift network operator is deployed and Multus is requested on addons operator, just make sure it is enabled on openshift operator (we don't deploy anything ourselves).

Since we cannot run both our and openshift SR-IOV plugins at the same time, I though we could follow the same logic. If we are deployed on OpenShift 4 and our user requests SR-IOV, make sure it is enabled on OpenShift (if not, then fail).

I know it is not perfect, do you have any better flow in mind?

booxter: @phoracek do you mean 'Make sure it's NOT enabled on OpenShift'? Because if it's enabled there, we can't reasonably use what they deploy for SR-IOV support for KubeVirt, so arguably, in this case I would expect our operator to return an error and not proceed with deployment.

phoracek: Oh I see! Sounds good to me :)

booxter:
@phoracek hope this one is better. I now handle conflict with openshift as well as I reverted the image name back to kubevirt since it was a simple mistake where I left my local testing environment bits in the PR.

phoracek:
Thanks @booxter :)

PR #16, opened 2019-03-25
WIP keep all components in single operators namespace

phoracek:
Let's group all components under a single namespace. It should make
it easier to track components deployed by the operator and it was
requested by hyperconverged-cluster-operator maintainers.

There will be a follow-up PR making namespace of the operato itself
configurable.

### DIFF
--- a/data/linux-bridge/000-ns.yaml
+++ b/data/linux-bridge/000-ns.yaml
@@ -1,6 +1,0 @@
-apiVersion: v1
-kind: Namespace
-metadata:
-  name: linux-bridge
-  labels:
-    name: linux-bridge
--- a/data/linux-bridge/000-rbac.yaml
+++ b/data/linux-bridge/000-rbac.yaml
@@ -4,7 +4,7 @@ apiVersion: v1
 kind: ServiceAccount
 metadata:
   name: linux-bridge
-  namespace: linux-bridge
+  namespace: {{ .Namespace }}
 ---
 apiVersion: security.openshift.io/v1
 kind: SecurityContextConstraints
@@ -17,5 +17,5 @@ runAsUser:
 seLinuxContext:
   type: RunAsAny
 users:
-- system:serviceaccount:linux-bridge:linux-bridge
+- system:serviceaccount:{{ .Namespace }}:linux-bridge
 {{ end }}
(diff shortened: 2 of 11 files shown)


phoracek:
@SchSeba @booxter could you please review?

booxter:
You can track components with labels, so that's probably not the only way to fix it. Why is this approach better than existing one? The existing separated approach seems to be more in line with e.g. openshift where each component is split in its own service namespace.

phoracek:
It is per kubevirt/hyperconverged-cluster-operator maintainers request: "They are standalone projects, but they are deployed under the larger *hyperconverged-cluster-operator* project. I would say it's really a best practice to keep things together in the same NS.  It's to respect the k8s multi-tenancy as well as enhance the UX."

First I wanted to make it configurable under which namespace will be Multus and Linux Bridge deployed, default would be cluster-network-addons-operators, but it would be possible to override it. However, I'd prefer not to maintain too many parameters. So unless someone requests this feature, I'd stick to a single namespace.

Even with this change, cluster-network-addons-operator will not be deployed under hyperconverged-cluster-operator namespaces because of we cannot create ClusterRoleBinding unless we explicitly specify namespace in manifests. That will be (hopefully) fixed once we integrate with OLM.

booxter:
@phoracek how is putting all components into a single namespace == "respect the k8s multi-tenancy"? If anything, it's exactly not respecting options available to us to split components into logically separate namespaces. Again, on the other side keeping them separated is more in line with what OpenShift does for other components, so I still wonder how this makes for a better UX at least in case of OpenShift.

I guess what I am saying is that `hyperconverged-cluster-operator` requests should not necessarily drive decisions made here. I am not saying that keeping everything in a single namespace is too bad, I am just trying to understand what's the real motivation to make a change here. (I understand the project just started and the cost of change is quite low now, but there must be some rationale besides "these other people told us to."; I am almost sure there is one, I just don't see it yet.)

phoracek:
@booxter sure https://github.com/kubevirt/hyperconverged-cluster-operator/pull/10

phoracek:
@booxter see the Slack discussion I sent to you on mail. That's all I got. I just wanted to follow what hyperconverged-cluster-operator developers think make sense from the higher layer.

ovirt-infra:
All tests passed

phoracek:
Let's close this for now. I don't hear requests for this behavior anymore.

PR #17, opened 2019-03-28
configure ImagePullPolicy via operator CR

phoracek:
Allow administrator to set desired ImagePullPolicy through
NetworkAddonsConfig CR. Until now, it was possible only using
operator's environment variables.

This PR will align us with KubeVirt and CDI operators.

### DIFF
--- a/pkg/network/image-pull-policy.go
+++ b/pkg/network/image-pull-policy.go
@@ -1,0 +1,54 @@
+package network
+
+import (
+	"github.com/pkg/errors"
+	v1 "k8s.io/api/core/v1"
+
+	opv1alpha1 "github.com/kubevirt/cluster-network-addons-operator/pkg/apis/networkaddonsoperator/v1alpha1"
+)
+
+const defaultImagePullPolicy = string(v1.PullIfNotPresent)
+
+func validateImagePullPolicy(conf *opv1alpha1.NetworkAddonsConfigSpec) []error {
+	if conf.ImagePullPolicy == "" {
+		return []error{}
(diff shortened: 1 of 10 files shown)


phoracek:
@SchSeba @booxter could you please take a look?

pkg/network/image-pull-policy.go
```diff
@@ -0,0 +1,54 @@
+package network
```
SchSeba: can you add license here?

pkg/network/image-pull-policy.go
```diff
@@ -0,0 +1,54 @@
...
+
+import (
+	"github.com/pkg/errors"
+	v1 "k8s.io/api/core/v1"
+
+	opv1alpha1 "github.com/kubevirt/cluster-network-addons-operator/pkg/apis/networkaddonsoperator/v1alpha1"
+)
+
+const defaultImagePullPolicy = "IfNotPresent"
```
booxter: should you use `v1.PullIfNotPresent` instead of a literal here?

pkg/network/image-pull-policy.go
phoracek: we don't have the license header in any of the other modules. we should add it in another PR

pkg/network/image-pull-policy.go
phoracek: Fixed

phoracek:
I fixed the constant. License headers should be added in a separate PR since they are missing in the whole project.

PR #19, opened 2019-03-29
use OpenShift Network definition from openshift/api

### LINKED ISSUE #18: Update to OpenShift Network object
Currently we check for existing NetworkConfig object (created by cluster-network-operator) for existing Multus configuration. With latest openshift/cluster-network-operator, this object has been renamed to Network. Sync with this change.

phoracek:
Latest openshift/cluster-network-operator renamed its resource from
NetworkConfig to Config and moved its definition under openshift/api.

This patch reflects this change in cluster-network-addons-operator.

Resolves #18


phoracek:

booxter:
@phoracek I would merge if I would have access.

PR #27, opened 2019-04-01
Change the linux deployment yaml

SchSeba:
Every file that contains a postfix of bridge and tuning will be copy

CNV is Red Hat product based on KubeVirt.
We plan to ship these plugins with cnv-* prefix to make sure they don't break regular openshift plugins if they are also deployed on the same node.

---
This change is [Reviewable](https://reviewable.io/reviews/kubevirt/cluster-network-addons-operator/27)

### DIFF
--- a/data/linux-bridge/002-linux-bridge.yaml
+++ b/data/linux-bridge/002-linux-bridge.yaml
@@ -31,8 +31,8 @@ spec:
             - /bin/bash
             - -c
             - |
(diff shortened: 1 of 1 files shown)


phoracek:
@booxter I'm all for nit-picking, that's why I reverted the original one. Also, I felt guilty not waiting for your input :)

phoracek:
@SchSeba can we merge?

SchSeba:
@phoracek yes this is ready

PR #28, opened 2019-04-01
Document that SR-IOV network is not created by the operator

booxter:
At least, as of now.


booxter:
@phoracek yes restart is needed. We should actually come up with a better solution than that for the future, I don't think it's a good final point. I will update as you implied.

/hold

booxter:
@phoracek is it better?

phoracek:
@booxter yup! Thanks :)

PR #31, opened 2019-04-02
update README with some forgotten fields

phoracek:
We forgot to update some parts of the README with added features. Let's fix it.


phoracek:
@SchSeba @booxter could you please take a look? Just a small change for the sake of consistency.

PR #32, opened 2019-04-02
Generate release manifests

phoracek:
This will make releasing of a new version and testing on local
cluster easier.


SchSeba:
question: can you maybe use https://github.com/kubernetes-sigs/kustomize ?

phoracek:
Maybe later if it becomes more complex. However, now with a couple of parameters, I prefer few lines of `sed` over another tool dependency (hello operator-sdk) and several yamls.

phoracek:
@booxter @SchSeba could you please review?

phoracek:
In order to make nesting of manifests possible (needed for SVC), we need to keep definitions in code.

PR #41, opened 2019-04-04
keep applied config in the same namespace as the operator

phoracek:
This removes hardcoded namespace for applied config configmap. This
unblocks changing of operator's namespace just by changing its manifests.


phoracek:

PR #46, opened 2019-04-10
Fix local cluster manifests

phoracek:
`make cluster-sync` and `make cluster-clean` has been broken with
b3c6fcc. In this patch, we use newly introduced `make manifests`
to generate manifests with local cluster registry and then use
them to deploy the operator.

On top of that, `WATCH_NAMESPACE` has been changed from `""` to operator's
own namespace. That causes controllers to reconcile only objects from
our namespace and therefore ignore components deployed to their respective
namespaces. This bug prevented NetworkAddonsConfig.Status field to be
updated with state of deployed DaemonSets and Deployments.


phoracek:

ovirt-infra:
All tests passed

phoracek:
@SchSeba  @booxter ready to be reviewed now, master is saved!

ovirt-infra:
All tests passed

cluster/clean.sh
```diff
@@ -2,9 +2,9 @@
 
 echo 'Cleaning up ...'
 
-./cluster/kubectl.sh delete --ignore-not-found -f deploy/cluster-network-addons-operator_01_crd.yaml
-./cluster/kubectl.sh delete --ignore-not-found -f deploy/cluster-network-addons-operator_02_rbac.yaml
-./cluster/kubectl.sh delete --ignore-not-found -f deploy/cluster-network-addons-operator_00_namespace.yaml
+./cluster/kubectl.sh delete --ignore-not-found -f _out/namespace.yaml
```
SchSeba: I think we need to first remove the network-addons-config maybe wait so time for all the related object (multus, macpool, linux bridge, sriov) to be remove then the operator and then the namespace?

deploy/olm-catalog/cluster-network-addons/0.0.0/cluster-network-addons-operator.v0.0.0.clusterserviceversion.yaml
```diff
@@ -169,9 +169,6 @@ spec:
                       fieldRef:
                         fieldPath: metadata.name
                   - name: WATCH_NAMESPACE
-                    valueFrom:
```
SchSeba: there is no value here it's fine?

cluster/clean.sh
phoracek: You are not wrong. This functionality is not implemented yet (we need to add finalizer that will wait with removal of NetworkAddonsConfig until all components are removed). However, it is not a part of this PR.

deploy/olm-catalog/cluster-network-addons/0.0.0/cluster-network-addons-operator.v0.0.0.clusterserviceversion.yaml
phoracek: AFAIK it is. Value is of type `string` and if no value is provided it defaults to `""`.

cluster/clean.sh
SchSeba: its ok if we don't wait on this PR but this way we remove the namespace(and the operator inside of it) I think for now we can just remove the CR wait for like 5-10 second so the operator can delete the render manifest and then remove the operator without waiting for all the components to be removed and implement the finalizer in a separate PR

phoracek: I changed it so order of manifest removals is the same as it was before this PR, hence there is no change related to this PR other than changing location of those manifests.

Synchronization is not a top priority now.

ovirt-infra:
All tests passed

PR #49, opened 2019-04-10
Use kubernetes core PullPolicy for imagePullPolicy

phoracek:
Instead of a string, we should use Kubernetes Core v1 PullPolicy
type to specify image pull policy for components' pods.

This is done per request from https://github.com/kubevirt/hyperconverged-cluster-operator/pull/23#discussion_r273713084.


phoracek:
@SchSeba  @booxter  could you please review?

ovirt-infra:
All tests passed

PR #54, opened 2019-04-11
Support OpenShift 4

phoracek:
With OpenShift 4, CNI config and bin directories has been moved from
/opt/cni/bin to /var/lib/cni/bin and from /etc/cni/net.d to
/etc/kubernetes/cni/net.d.

With this patch, we check whether we are running on OpenShift 4 by
looking up operator objects introduced in OpenShift 4 and making
CNI paths in components' manifests based on that.


phoracek:
@SchSeba @booxter please review

PR #57, opened 2019-04-11
Unit Test for validateKubeMacPool & fillDefaultsKubeMacPool

alonSadan:
Tests that validation is done correctly, and that the correct error message is sent.


ovirt-infra:
Hello contributor, thanks for submitting a PR for this project! 

I am the bot who triggers "standard-CI" builds for this project.
As a security measure, I will not run automated tests on PRs that are not from white-listed contributors.

In order to allow automated tests to run, please ask one of the project maintainers to review the code and then do one of the following:

1. Type `ci test please` on this PR to trigger automated tests for it.
2. Type `ci add to whitelist` on this PR to trigger automated tests for it and also add you to the contributor white-list so that your future PRs will be tested automatically. ( keep in mind this list might be overwritten if the job XML is refreshed, for permanent whitelisting, please follow #3 option )
3. If you are planning to contribute to more than one project, maybe it's better to ask them to add you to the project organization, so you'll be able to run tests for all the organization's projects.

pkg/network/network_test.go
```diff
@@ -0,0 +1,66 @@
+package network
```
phoracek: please rename the module to kubemacpool_test.go, so we don't mix all tests together later

pkg/network/network_test.go
```diff
@@ -0,0 +1,66 @@
...
+
+import (
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	opv1alpha1 "github.com/kubevirt/cluster-network-addons-operator/pkg/apis/networkaddonsoperator/v1alpha1"
+)
+
+var _ = Describe("Network", func() {
+	Describe("Internal Functions", func() {
```
phoracek: Once you move it to a test module dedicated to kubemacpool, don't forget to rename these. You can drop the inner Describe and rename outer

pkg/network/network_test.go
```diff
@@ -0,0 +1,66 @@
...
+
+		It("should return an error because only end range is configured", func() {
+			clusterConfig := &opv1alpha1.NetworkAddonsConfigSpec{
+				KubeMacPool: &opv1alpha1.KubeMacPool{StartPoolRange: "", EndPoolRange: "02:00:00:FF:FF:FF"}}
+			errorList := validateKubeMacPool(clusterConfig)
+			Expect(len(errorList)).To(Equal(1))
+			Expect(errorList[0].Error()).To(Equal("both or none of the KubeMacPool ranges needs to be configured"))
+
+		})
+		It("should return an error because the start range mac address is invalid", func() {
```
phoracek: empty line above this

pkg/network/network_test.go
```diff
@@ -0,0 +1,66 @@
...
+			Expect(errorList[0].Error()).To(Equal("failed to parse startPoolRange invalid mac address"))
+		})
+
+		It("should return an error because the end range mac address is invalid", func() {
+			clusterConfig := &opv1alpha1.NetworkAddonsConfigSpec{
+				KubeMacPool: &opv1alpha1.KubeMacPool{StartPoolRange: "02:00:00:00:00:00", EndPoolRange: "FF:FF:FF:FF:FF:FF:FF"}}
+			errorList := validateKubeMacPool(clusterConfig)
+			Expect(len(errorList)).To(Equal(1))
+			Expect(errorList[0].Error()).To(Equal("failed to parse endPoolRange invalid mac address"))
+
```
phoracek: please drop this empty line

pkg/network/network_test.go
phoracek: Please describe what is invalid here (start > end, right?)

pkg/network/network_test.go
```diff
@@ -0,0 +1,66 @@
...
+		})
+
+		It("should NOT return an error mac address is valid", func() {
+			clusterConfig := &opv1alpha1.NetworkAddonsConfigSpec{
+				KubeMacPool: &opv1alpha1.KubeMacPool{StartPoolRange: "02:00:00:00:00:00", EndPoolRange: "02:FF:FF:FF:FF:FF"}}
+			errorList := validateKubeMacPool(clusterConfig)
+			Expect(len(errorList)).To(Equal(0))
+
+		})
+
```
phoracek: please add a test with invalid MAC address

pkg/network/network_test.go
```diff
@@ -0,0 +1,66 @@
...
+			Expect(len(errorList)).To(Equal(1))
+			Expect(errorList[0].Error()).To(Equal("both or none of the KubeMacPool ranges needs to be configured"))
+		})
+
+		It("should return an error because only end range is configured", func() {
+			clusterConfig := &opv1alpha1.NetworkAddonsConfigSpec{
+				KubeMacPool: &opv1alpha1.KubeMacPool{StartPoolRange: "", EndPoolRange: "02:00:00:FF:FF:FF"}}
+			errorList := validateKubeMacPool(clusterConfig)
+			Expect(len(errorList)).To(Equal(1))
+			Expect(errorList[0].Error()).To(Equal("both or none of the KubeMacPool ranges needs to be configured"))
```
phoracek: To make test logs more readable, please add explanation **what** failed to each Expect. e.g.:

```
Expect(errorList[0].Error()).To(Equal("both or none of the KubeMacPool ranges needs to be configured"), "Validation failed with unexpected error message: %s", errorList[0].Error())
```

Similar in other tests. It really makes a big difference while working with failing tests.

pkg/network/network_test.go
```diff
@@ -0,0 +1,66 @@
...
+
+	opv1alpha1 "github.com/kubevirt/cluster-network-addons-operator/pkg/apis/networkaddonsoperator/v1alpha1"
+)
+
+var _ = Describe("Network", func() {
+	Describe("Internal Functions", func() {
+		It("should NOT return an error because kube mac pool can be nil", func() {
+			clusterConfig := &opv1alpha1.NetworkAddonsConfigSpec{KubeMacPool: &opv1alpha1.KubeMacPool{}}
+			errorList := validateKubeMacPool(clusterConfig)
+			Expect(len(errorList)).To(Equal(0))
```
phoracek: I believe you can do:

```
Expect(errorList).To(BeEmpty())
```

pkg/network/network_test.go
```diff
@@ -0,0 +1,66 @@
...
+		It("should return an error because the end range mac address is invalid", func() {
+			clusterConfig := &opv1alpha1.NetworkAddonsConfigSpec{
+				KubeMacPool: &opv1alpha1.KubeMacPool{StartPoolRange: "02:00:00:00:00:00", EndPoolRange: "FF:FF:FF:FF:FF:FF:FF"}}
+			errorList := validateKubeMacPool(clusterConfig)
+			Expect(len(errorList)).To(Equal(1))
+			Expect(errorList[0].Error()).To(Equal("failed to parse endPoolRange invalid mac address"))
+
+		})
+
+		It("should NOT return an error mac address is valid", func() {
```
dankenigsberg: should NOT return an error *because the* mac address is valid

pkg/network/network_test.go
dankenigsberg: @phoracek do you mean a MAC with the *multicast* bit on? I agree that it should be tested.

pkg/network/network_test.go
```diff
@@ -0,0 +1,66 @@
...
+			clusterConfig := &opv1alpha1.NetworkAddonsConfigSpec{
+				KubeMacPool: &opv1alpha1.KubeMacPool{StartPoolRange: "", EndPoolRange: "02:00:00:FF:FF:FF"}}
+			errorList := validateKubeMacPool(clusterConfig)
+			Expect(len(errorList)).To(Equal(1))
+			Expect(errorList[0].Error()).To(Equal("both or none of the KubeMacPool ranges needs to be configured"))
+
+		})
+		It("should return an error because the start range mac address is invalid", func() {
+			clusterConfig := &opv1alpha1.NetworkAddonsConfigSpec{
+				KubeMacPool: &opv1alpha1.KubeMacPool{StartPoolRange: "FF:FF:FF:FF:FF:FF:FF", EndPoolRange: "02:FF:FF:FF:FF:FF"}}
```
dankenigsberg: I'm not a native English speaker, but I think 'StartPoolRange' is not proper. I'd prefer to call it 'RangeStart'; it is quite clear that it is an attribute of the Pool.

pkg/network/network_test.go
dankenigsberg: yes, 'invalid' is a bit vague here; this specific address fails because it is too long (more than 6 octets). We should add another test for negative-length ranges.

pkg/network/network_test.go
phoracek: @phoracek then we need two tests! One for multicast, one for a string that doesn't resemble a MAC address at all, "foo:bar".

pkg/network/network_test.go
alonSadan: Hi @phoracek . Thank you for the comments. 
The error message in errorList[0] would be quite similar to the explanation that is in the "It" statement.
would you like me to add that anyway?
Thanks.

`	It("should return an error because only the end range is configured", func() {
`

phoracek: Good point. You are right, that if there is only a single Except within It block, it is not really needed. Please add this "explanation" message only in cases where it would be valuable to know, what the value is. e.g. if the value is unexpected, print it in Exception message.

alonSadan: Thanks. Will do.

SchSeba:
@alonSadan please update the kubemacpool image to use `quay.io/kubevirt/kubemacpool:v0.2.0`

pkg/network/kubemacpool.go
```diff
@@ -40,6 +41,19 @@ func validateKubeMacPool(conf *opv1alpha1.NetworkAddonsConfigSpec) []error {
 		return []error{errors.Errorf("failed to parse rangeEnd because the mac address is invalid")}
 	}
 
+	if err := checkRange(net.HardwareAddr(conf.KubeMacPool.RangeStart), net.HardwareAddr(conf.KubeMacPool.RangeEnd)); err != nil {
+		return []error{errors.Errorf("faild to set mac address range. Range value is negative")}
+	}
+
+	rangeStart, _ := net.ParseMAC(conf.KubeMacPool.RangeStart)
+	if err := checkCast(rangeStart); err != nil {
+		return []error{errors.Errorf("faild to set mac address range. The first octet of RangeStart is not 02, 06, 0A or 0E")}
```
phoracek: failed

pkg/network/kubemacpool.go
```diff
@@ -40,6 +41,19 @@ func validateKubeMacPool(conf *opv1alpha1.NetworkAddonsConfigSpec) []error {
 		return []error{errors.Errorf("failed to parse rangeEnd because the mac address is invalid")}
 	}
 
+	if err := checkRange(net.HardwareAddr(conf.KubeMacPool.RangeStart), net.HardwareAddr(conf.KubeMacPool.RangeEnd)); err != nil {
```
phoracek: Could you mention in the commit message that you added this validation, not only the test for it?

pkg/network/kubemacpool.go
```diff
@@ -40,6 +41,19 @@ func validateKubeMacPool(conf *opv1alpha1.NetworkAddonsConfigSpec) []error {
...
+		return []error{errors.Errorf("faild to set mac address range. Range value is negative")}
+	}
+
+	rangeStart, _ := net.ParseMAC(conf.KubeMacPool.RangeStart)
+	if err := checkCast(rangeStart); err != nil {
+		return []error{errors.Errorf("faild to set mac address range. The first octet of RangeStart is not 02, 06, 0A or 0E")}
+	}
+	rangeEnd, _ := net.ParseMAC(conf.KubeMacPool.RangeEnd)
+	if err := checkCast(rangeEnd); err != nil {
+		return []error{errors.Errorf("faild to set mac address range. The first octet of RangeEnd is not 02, 06, 0A or 0E")}
```
phoracek: rangeEnd

pkg/network/kubemacpool.go
phoracek: rangeStart

pkg/network/kubemacpool.go
```diff
@@ -95,3 +109,22 @@ func generateRandomMacPrefix() ([]byte, error) {
...
 }
+
+func checkRange(startMac, endMac net.HardwareAddr) error {
+	for idx := 0; idx <= 5; idx++ {
+		if startMac[idx] < endMac[idx] {
+			return nil
+		}
+	}
+
+	return fmt.Errorf("Invalid range start: %s end: %s", startMac.String(), endMac.String())
```
phoracek: error messages should start with lower case

pkg/network/kubemacpool.go
```diff
@@ -95,3 +109,22 @@ func generateRandomMacPrefix() ([]byte, error) {
...
+}
+
+func checkCast(mac net.HardwareAddr) error {
+	s := string(mac[0])
+	if s == "\x02" || s == "\x06" || s == "\x0A" || s == "\x0E" {
+		return nil
+	}
+
+	return fmt.Errorf("Invalid mac address. The fisrt octet should be: 02, 06, 0A or 0E. Instead got: %s", s)
```
phoracek: start with lower case

first

pkg/network/kubemacpool_test.go
```diff
@@ -0,0 +1,102 @@
...
+
+import (
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	opv1alpha1 "github.com/kubevirt/cluster-network-addons-operator/pkg/apis/networkaddonsoperator/v1alpha1"
+)
+
+var _ = Describe("kubemacpool_test", func() {
+	It("should NOT return an error because kube mac pool can be nil", func() {
```
phoracek: Someone please correct me if I'm wrong, but I think that in Ginkgo, this should be something like:

```go
    Context("When kubeMacPool is set to nil", func() {
		It("validation should not fail", func() {
			clusterConfig := &opv1alpha1.NetworkAddonsConfigSpec{KubeMacPool: &opv1alpha1.KubeMacPool{}}
			errorList := validateKubeMacPool(clusterConfig)
			Expect(errorList).To(BeEmpty())
		})
	})
```

The same for the others, don't hesitate to ping me if you stuck somewhere.

phoracek:
@alonSadan don't forget to run `dep ensure`, to pull ginkgo and gomega dependencies under `vendor/` directory.

pkg/network/kubemacpool.go
```diff
@@ -95,3 +109,26 @@ func generateRandomMacPrefix() ([]byte, error) {
...
+		if startMac[idx] < endMac[idx] {
+			return nil
+		}
+	}
+
+	return fmt.Errorf("invalid range start: %s end: %s", startMac.String(), endMac.String())
+}
+
+func checkCast(mac net.HardwareAddr) error {
+	bit := 1 & mac[0]
```
phoracek: please use binary:

```go
const multicastBit = 0b00000001
...
castBit := 0b00000001 & mac[0]
if castBit == multicastBit {
...
```

pkg/network/kubemacpool.go
```diff
@@ -95,3 +109,26 @@ func generateRandomMacPrefix() ([]byte, error) {
...
+	if bit == 1 {
+		return fmt.Errorf("invalid mac address. Multicast addressing is not supported. The first octet is %#0X", mac[0])
+	}
+	s := mac[0]
```
phoracek: what does the `s` stand for?

pkg/network/kubemacpool.go
alonSadan: I think GO does not support binary literals.
@phoracek

phoracek: you right!

pkg/network/kubemacpool.go
```diff
@@ -95,3 +109,21 @@ func generateRandomMacPrefix() ([]byte, error) {
...
+	}
+	return fmt.Errorf("invalid range. Range end is lesser than its start. start: %v end: %v", startMac, endMac)
+}
+
+func checkCast(mac net.HardwareAddr) error {
+	castBit := 1 & mac[0]
+	if castBit == 1 {
+		return fmt.Errorf("invalid mac address. Multicast addressing is not supported. Should be unicast addressing. The first octet is %#0X", mac[0])
+	}
+
```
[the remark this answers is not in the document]
SchSeba: Yes but after a discussion we can't find a reason to block the user if he want to allocate a vendor reserve range. we just going to block if the range is a multicast one.

when we generate the range it will always be a local one.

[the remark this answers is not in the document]
phoracek: Fair enough.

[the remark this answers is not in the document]
alonSadan: We discussed about this at the CNV network sync , and we concluded that we should allow both local and universal .

pkg/network/kubemacpool.go
phoracek: Anyway, please add inline comment describing what you do there. Something about the right-most bit from the first octet.

pkg/network/kubemacpool_test.go
```diff
@@ -0,0 +1,127 @@
...
+		It("should return an error", func() {
+			clusterConfig := &opv1alpha1.NetworkAddonsConfigSpec{
+				KubeMacPool: &opv1alpha1.KubeMacPool{RangeStart: "02:00:00:ff:00:00", RangeEnd: "02:00:00:00:00:00"}}
+			errorList := validateKubeMacPool(clusterConfig)
+			Expect(len(errorList)).To(Equal(1), "validation failed due to an unexpected error: %s", errorList[0].Error())
+			Expect(errorList[0].Error()).To(Equal("failed to set mac address range: invalid range. Range end is lesser than or equal to its start. start: 02:00:00:ff:00:00 end: 02:00:00:00:00:00"))
+		})
+	})
+
+	Context("When range end is the same as it's start", func() {
```
phoracek: "its"

pkg/network/kubemacpool_test.go
```diff
@@ -0,0 +1,127 @@
...
+		It("should return an error", func() {
+			clusterConfig := &opv1alpha1.NetworkAddonsConfigSpec{
+				KubeMacPool: &opv1alpha1.KubeMacPool{RangeStart: "02:00:00:ff:00:00", RangeEnd: "02:00:00:ff:00:00"}}
+			errorList := validateKubeMacPool(clusterConfig)
+			Expect(len(errorList)).To(Equal(1), "validation failed due to an unexpected error: %s", errorList[0].Error())
+			Expect(errorList[0].Error()).To(Equal("failed to set mac address range: invalid range. Range end is lesser than or equal to its start. start: 02:00:00:ff:00:00 end: 02:00:00:ff:00:00"))
+		})
+	})
+
+	Context("when range end is greater than it's start only by 2", func() {
```
phoracek: "its"

pkg/network/kubemacpool.go
phoracek: For the record: We should not. Marking as resolved

pkg/network/kubemacpool.go
```diff
@@ -32,14 +33,27 @@ func validateKubeMacPool(conf *opv1alpha1.NetworkAddonsConfigSpec) []error {
...
 		return []error{errors.Errorf("failed to parse rangeEnd because the mac address is invalid")}
 	}
 
+	if err := validateRange(rangeStart, rangeEnd); err != nil {
+		return []error{errors.Errorf("failed to set mac address range: %v", err)}
+	}
+
+	if err := validateUnicast(rangeStart); err != nil {
+		return []error{errors.Errorf("failed to set RangeStart: %v", err)}
+	}
```
SchSeba: new line here please

pkg/network/kubemacpool.go
```diff
@@ -110,3 +124,23 @@ func generateRandomMacPrefix() ([]byte, error) {
...
+		if startMac[idx] < endMac[idx] {
+			return nil
+		}
+	}
+	return fmt.Errorf("invalid range. Range end is lesser than or equal to its start. start: %v end: %v", startMac, endMac)
+}
+
+func validateUnicast(mac net.HardwareAddr) error {
+	multicastBit := 1 & mac[0] //a bitwise AND between 00000001 and the mac address first octet.
+	//in case where the LSB of the first octet (the multicast bit) is on, it will return 1, and 0 otherwise.
```
SchSeba: put this on top the if statement

pkg/network/kubemacpool_test.go
```diff
@@ -0,0 +1,184 @@
...
+
+import (
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	opv1alpha1 "github.com/kubevirt/cluster-network-addons-operator/pkg/apis/networkaddonsoperator/v1alpha1"
+)
+
+var _ = Describe("Testing kubeMacPool", func() {
+	var _ = Describe("Testing kubeMacPool validation", func() {
```
SchSeba: remove the `var _ = ` here no need and please check the describe to `Validation function`

pkg/network/kubemacpool_test.go
```diff
@@ -0,0 +1,184 @@
...
+				clusterConfig := &opv1alpha1.NetworkAddonsConfigSpec{
+					KubeMacPool: &opv1alpha1.KubeMacPool{RangeStart: "02:00:00:00:00:00", RangeEnd: "0A:FF:FF:FF:FF:FF"}}
+				errorList := validateKubeMacPool(clusterConfig)
+				Expect(errorList).To(BeEmpty())
+			})
+		})
+
+	})
+
+	var _ = Describe("Testing fillDefaultsKubeMacPool", func() {
```
SchSeba: you can remove the `var _ = ` here

pkg/network/kubemacpool_test.go
```diff
@@ -0,0 +1,184 @@
...
+		Context("When kubeMacPool is nil", func() {
+			It("should NOT return an error and do nothing", func() {
+				currentClusterConfig := &opv1alpha1.NetworkAddonsConfigSpec{}
+				previousClusterConfig := &opv1alpha1.NetworkAddonsConfigSpec{
+					KubeMacPool: &opv1alpha1.KubeMacPool{RangeStart: "02:00:00:00:00:00", RangeEnd: "0A:FF:FF:FF:FF:FF"}}
+				errorList := fillDefaultsKubeMacPool(currentClusterConfig, previousClusterConfig)
+				Expect(errorList).To(BeEmpty())
```
SchSeba: please add a check here 
```golang
Expect(currentClusterConfig.KubeMacPool).To(BeNill())
```

pkg/network/kubemacpool_test.go
```diff
@@ -0,0 +1,184 @@
...
+
+		Context("When the user has explicitly requested a range", func() {
+			It("should leave the range as it is, and not return an error", func() {
+				previousClusterConfig := &opv1alpha1.NetworkAddonsConfigSpec{
+					KubeMacPool: &opv1alpha1.KubeMacPool{RangeStart: "02:00:00:00:00:00", RangeEnd: "0A:FF:FF:FF:FF:FF"}}
+				currentClusterConfig := &opv1alpha1.NetworkAddonsConfigSpec{
+					KubeMacPool: &opv1alpha1.KubeMacPool{RangeStart: "02:00:00:ff:00:00", RangeEnd: "02:00:00:ff:00:02"}}
+				errorList := fillDefaultsKubeMacPool(currentClusterConfig, previousClusterConfig)
+				Expect(errorList).To(BeEmpty())
+				Expect(currentClusterConfig.KubeMacPool.RangeStart).To(Equal(currentClusterConfig.KubeMacPool.RangeStart))
```
SchSeba: you send a pointer here so it will always be true.

please save the start and end range to environment variables and use it when you create the current config and also to compare here

pkg/network/kubemacpool.go
```diff
@@ -110,3 +125,24 @@ func generateRandomMacPrefix() ([]byte, error) {
 
 	return prefix, nil
 }
+
+func validateRange(startMac, endMac net.HardwareAddr) error {
+	for idx := 0; idx <= 5; idx++ {
```
phoracek: Why do we skip the last octet?

pkg/network/kubemacpool_test.go
```diff
@@ -0,0 +1,190 @@
...
+import (
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	opv1alpha1 "github.com/kubevirt/cluster-network-addons-operator/pkg/apis/networkaddonsoperator/v1alpha1"
+)
+
+var _ = Describe("Testing kubeMacPool", func() {
+	Describe("validation function", func() {
+		Context("When kubeMacPool is nil and do nothing", func() {
```
phoracek: I don't understand this sentence

pkg/network/kubemacpool_test.go
```diff
@@ -0,0 +1,190 @@
...
+			It("should NOT return an error", func() {
+				clusterConfig := &opv1alpha1.NetworkAddonsConfigSpec{}
+				errorList := validateKubeMacPool(clusterConfig)
+				Expect(errorList).To(BeEmpty())
+			})
+		})
+
+		Context("When both or none of the ranges are configured ", func() {
```
phoracek: this tests only one of the cases, doesn't it

pkg/network/kubemacpool_test.go
```diff
@@ -0,0 +1,190 @@
...
+				Expect(errorList[0].Error()).To(Equal("failed to parse rangeEnd because the mac address is invalid"))
+			})
+		})
+
+		Context("When range end is lesser than its start", func() {
+			It("should return an error", func() {
+				clusterConfig := &opv1alpha1.NetworkAddonsConfigSpec{
+					KubeMacPool: &opv1alpha1.KubeMacPool{RangeStart: "02:00:00:ff:00:00", RangeEnd: "02:00:00:00:00:00"}}
+				errorList := validateKubeMacPool(clusterConfig)
+				Expect(len(errorList)).To(Equal(1), "validation failed due to an unexpected error: %s", errorList[0].Error())
```
phoracek: `"validation returned unexpected errors: %v", errorList`

What if the first one is the one we expected, but following are indicating what is wrong

pkg/network/kubemacpool_test.go
```diff
@@ -0,0 +1,190 @@
...
+					errorList := fillDefaultsKubeMacPool(currentClusterConfig, previousClusterConfig)
+					Expect(errorList).To(BeEmpty())
+					Expect(currentClusterConfig.KubeMacPool.RangeStart).To(Equal(previousClusterConfig.KubeMacPool.RangeStart))
+					Expect(currentClusterConfig.KubeMacPool.RangeEnd).To(Equal(previousClusterConfig.KubeMacPool.RangeEnd))
+
+				})
+			})
+
+			Context("When a previous kubeMacpool doesn't exits", func() {
+				It("should generate a new range for the current kubeMacPoll, and not return an error", func() {
```
phoracek: kubeMacPool

pkg/network/kubemacpool_test.go
```diff
@@ -0,0 +1,190 @@
...
+					previousClusterConfig := &opv1alpha1.NetworkAddonsConfigSpec{}
+					currentClusterConfig := &opv1alpha1.NetworkAddonsConfigSpec{
+						KubeMacPool: &opv1alpha1.KubeMacPool{RangeStart: "", RangeEnd: ""}}
+					errorList := fillDefaultsKubeMacPool(currentClusterConfig, previousClusterConfig)
+					Expect(errorList).To(BeEmpty())
+					Expect(currentClusterConfig.KubeMacPool.RangeStart).To(Not(Equal("")), "RangeStart:")
```
phoracek: That message in case of an error "RangeStart:" is not saying anything. Could you explain what went unexpected instead?

pkg/network/kubemacpool.go
alonSadan: Thank you  for the notes Petr.
I think we don't skip the last octet since there are six octets, and we use "<=", right?

pkg/network/kubemacpool_test.go
alonSadan: I'm not sure I get you here. do you mind clarifying what you meant?
Thanks.

SchSeba:
ci test please

pkg/network/kubemacpool.go
phoracek: :man_facepalming: sorry, you are absolutely right

pkg/network/kubemacpool_test.go
phoracek: So you test that exactly **one** error has been returned, if not, you raise. The raised message however only includes the first error. But what if we got 0 errors? Then we panic. And if we get 2 or more errors, we don't show them to the user running the test suite.

alonSadan:
Thank you both for the instructive notes

SchSeba:
ci test please

pkg/network/kubemacpool.go
```diff
@@ -91,9 +106,9 @@ func renderKubeMacPool(conf *opv1alpha1.NetworkAddonsConfigSpec, manifestDir str
 	data.Data["RangeStart"] = conf.KubeMacPool.RangeStart
 	data.Data["RangeEnd"] = conf.KubeMacPool.RangeEnd
 
-	objs, err := render.RenderDir(filepath.Join(manifestDir, "kubemacpool"), &data)
```
phoracek: this change introduced a bug, it has to remain lowercase

SchSeba: Good catch!

SchSeba:
ci test please

SchSeba:
for reference https://kubevirt-master-stdci-production.apps.ovirt.org/job/kubevirt_cluster-network-addons-operator_standard-check-pr/117/artifact/ci_build_summary.html

SchSeba:
ci test please

SchSeba:
ci test please

PR #59, opened 2019-04-11
Create sriov network attachment definition

booxter:
To use SR-IOV with KubeVirt, one needs to have a network attachment
definition that links back to resourceName used in SR-IOV DP config
file. This is the last missing bit in end-to-end SR-IOV integration.

I am using the `default` namespace and not `sriov` because it's assumed
that regular users will attach their pods / VMIs to the network, so
keeping it under `default` seems to make more sense. Otherwise one would
need to refer to the new network with namespace prefix included, as in:
`sriov/sriov-network` which is a bit too wordy.

Note that this PR doesn't allow to configure IPAM parameters for the
new network. I'd like to leave this piece separate from this patch.

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,6 +88,18 @@ can deploy with default configuration file and then modify
 You may need to restart SR-IOV device plugin pods to catch up configuration
 file changes.
 
+The operator will also deploy a new network attachment definition for SR-IOV
+network. By default, its name is `sriov-network` but it can be changed using
+the `SRIOV_NETWORK_NAME` environment variable.
+
+One may also want to change the type of the newly created network attachment
+definition from the default `sriov` to something else. For example, this is
+needed for Red Hat's CNV product that is deployed on top of OpenShift that may
+be already shipped with `sriov` CNI plugin of different version incompatible
+with KubeVirt. In this case an admin may want to use different network types
+for KubeVirt and OpenShift SR-IOV CNI plugins. This can be achieved using the
+`SRIOV_NETWORK_TYPE` environment variable.
+
 Additionally, container images used to deliver these plugins can be set using
 `SRIOV_DP_IMAGE` and `SRIOV_CNI_IMAGE` environment variables in operator
 deployment manifest.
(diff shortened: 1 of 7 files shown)


booxter:
@SchSeba @phoracek

phoracek:
@booxter I'm not sure I get the logic with `default`. Users will probably be in their respective projects/namespaces and in order to access the `NetworkAttachmentDefinition`, they will need to set the network to `default/sriov-network`, wouldn't they?

booxter:
@phoracek yeah sorry, I misinterpreted what Multus does when looking for a network without namespace prefix. I thought it reads from the `default` in addition to current namespace if it can't find it. It looks like it doesn't.

/hold

booxter:
@phoracek pushed an update that changes the single namespace line and cleans up commit message.

phoracek:
@booxter I think that in Multus, if namespace name is ommited, it looks up namespace of the Pod that is requesting the network.

README.md
```diff
@@ -88,6 +88,18 @@ can deploy with default configuration file and then modify
...
+network. By default, its name is `sriov-network` but it can be changed using
+the `SRIOV_NETWORK_NAME` environment variable.
+
+One may also want to change the type of the newly created network attachment
+definition from the default `sriov` to something else. For example, this is
+needed for Red Hat's CNV product that is deployed on top of OpenShift that may
+be already shipped with `sriov` CNI plugin of different version incompatible
+with KubeVirt. In this case an admin may want to use different network types
+for KubeVirt and OpenShift SR-IOV CNI plugins. This can be achieved using the
+`SRIOV_NETWORK_TYPE` environment variable.
```
phoracek: Would it make sense to use this variable to deploy the binary? So we could simply build operator manifest with different environment variable while building for CNV?

Now it is done by combining adjusted sriov image build and build of operator. If we keep it in here, it would be easier even on u/s. i.e. use the same image with different binary name.

booxter: This is a good idea.

booxter:
@phoracek rebased and now using the variable for final binary name.

README.md
booxter: Interestingly, I didn't need to change a word in README for this.

phoracek:
Neat!

PR #60, opened 2019-04-12
make addons images configurable during `make manifests`

### LINKED ISSUE #56: make components' image urls configurable in make manifests

phoracek:
With this patch it is possible to build alternative operator manifests
grabbing network addons from alternative registries.

Resolves #56


phoracek:
@booxter @SchSeba

PR #69, opened 2019-04-15
upgrade base image packages while building images

phoracek:
Images we use have known issues in their packages, make sure to upgrade
on build.

### DIFF
--- a/build/operator/Dockerfile
+++ b/build/operator/Dockerfile
@@ -9,6 +9,9 @@ ENV ENTRYPOINT=/entrypoint \
     OPERATOR=/cluster-network-addons-operator \
     USER_UID=1001 \
     USER_NAME=cluster-network-addons-operator
+RUN \
+    yum -y update \
+    yum clean all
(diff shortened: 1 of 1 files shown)


phoracek:
@booxter @SchSeba

build/registry/Dockerfile
```diff
@@ -2,6 +2,8 @@ FROM quay.io/openshift/origin-operator-registry
 
 COPY manifests /registry
 
+RUN yum -y update
```
SchSeba: Oh you can't run this command on this image (this user there doesn't have permissions) I think we can remove the `yum update` from this image

phoracek: quay reports that it found vulnerabilities there, lemme try to use sudo

phoracek: `sudo` is not installed in the image. dropped update of the registry image for now.

SchSeba: I think the pod doesn't have the sudo command. you can try to save the user then run `USER root` yum update and return the USER again.

I am not sure if this going to work but you can try

PR #71, opened 2019-04-16
allow namespace adjustment during manifests build

phoracek:
We need this change to integrate with HCO.

### DIFF
--- a/tools/manifest-templator/manifest-templator.go
+++ b/tools/manifest-templator/manifest-templator.go
@@ -136,6 +136,7 @@ func getCNA(data *templateData) {
 
 	// Get CNA Deployment
 	cnadeployment := components.GetDeployment(
+		data.Namespace,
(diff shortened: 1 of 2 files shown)


phoracek:
@SchSeba @booxter please take a look

tools/manifest-templator/manifest-templator.go
```diff
@@ -136,6 +136,7 @@ func getCNA(data *templateData) {
 
 	// Get CNA Deployment
 	cnadeployment := components.GetDeployment(
+		data.Namespace,
```
SchSeba: there is a reason to not send the data object?

phoracek: We would need to expose the object to components/components.go. In this state we pass basic explicit values (except addons images). `GetDeployment` is also used in HCO. I'd rather keep it basic (for now).

PR #76, opened 2019-04-18
Call syncFailing on SetNotFailing method

phoracek:
https://github.com/openshift/cluster-network-operator/pull/136:

Since nil is used as value for having a condition as
non failing, we need to call it outside of the if
otherwise the failing condition will always set to nil.


phoracek:
@SchSeba @booxter

PR #77, opened 2019-04-18
status: catch a few more Progressing cases

phoracek:
https://github.com/openshift/cluster-network-operator/pull/143:

The easiest way to tell that a daemonset is progressing is that
UpdatedNumberScheduled is less than Desired.

Also, catch the case where the daemonset update hasn't yet been observed
by the controller.


phoracek:
@SchSeba @booxter

PR #78, opened 2019-04-18
 run basic functional test on STD-CI

phoracek:
This test deploys operator, creates its config CR and waits until
all components are successfully deployed.

### DIFF
--- a/automation/test.sh
+++ b/automation/test.sh
@@ -1,0 +1,13 @@
+#!/bin/bash -ex
+
+kubectl() { cluster/kubectl.sh "$@"; }
+
+export CLUSTER_PROVIDER=$TARGET
+
+# Make sure that the VM is properly shut down on exit
+trap '{ make cluster-down; }' EXIT SIGINT SIGTERM SIGSTOP
+
+make cluster-down
+make cluster-up
+make cluster-sync
+make cluster-functest
--- a/cluster/functest.sh
+++ b/cluster/functest.sh
@@ -1,0 +1,56 @@
+#!/bin/bash -e
+
+function new_test() {
+    name=$1
+
+    printf "%0.s=" {1..80}
+    echo
+    echo ${name}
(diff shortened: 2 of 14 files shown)


phoracek:
@SchSeba @booxter

automation/test.sh
```diff
@@ -0,0 +1,18 @@
...
+export CLUSTER_PROVIDER=$TARGET
+
+# Make sure that the VM is properly shut down on exit
+trap '{ make cluster-down; }' EXIT SIGINT SIGTERM SIGSTOP
+
+make cluster-down
+make cluster-up
+make cluster-sync
+
+DEPLOY_DIR=_out/ VERSION=latest make generate-manifests
```
SchSeba: I think we don't need this the `make cluster-sync` already do it.

take a look on the logs here https://kubevirt-master-stdci-production.apps.ovirt.org/job/kubevirt_cluster-network-addons-operator_standard-check-pr/61//artifact/check-patch.k8s-1.11.0.el7.x86_64/mock_logs/script/stdout_stderr.log

you will the the manifest tool run two times.

phoracek: The problem is that manifests that are already generated use future version (so now, manifests are under 0.5.0, once we bump version it will change). I wanted to follow an explicit way so we know where to look for manifests. However, there is another problem, we overwrite the package file. That might be a surprise if we use it one day. Lemme change it a bit.

phoracek:
@SchSeba fixed. Please take another look.

SchSeba:
This PR is ready? can you drop the `hold` or not yet?

cluster/functest.sh
```diff
@@ -0,0 +1,56 @@
+#!/bin/bash -e
```
SchSeba: nice!

phoracek:
@SchSeba ready

PR #81, opened 2019-04-19
add helper script to make releasing easier

phoracek:
Make releases easier.

### DIFF
--- a/hack/release.sh
+++ b/hack/release.sh
@@ -1,0 +1,38 @@
+#!/usr/bin/env bash
+
+set -e
+
+released_version=$1
+future_version=$2
+origin_remote=$3
+fork_remote=$4
+
+echo 'Fetching origin'
+git fetch ${origin_remote}
+
+echo 'Checkout origin/master into release branch'
+git checkout ${origin_remote}/master -b release_${released_version}
+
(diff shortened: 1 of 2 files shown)


phoracek:
@SchSeba @booxter

hack/release.sh
```diff
@@ -0,0 +1,38 @@
...
+echo 'Commit Makefile with bumped versions'
+git add Makefile
+
+echo 'Commit updated Makefile'
+git commit -s -m "release ${released_version} - bump versions in Makefile"
+
+echo 'Push changes to forked repo'
+git push ${fork_remote}
+
+echo 'The rest is on you, open a new PR. Once the PR is merged, do not forget to tag the first of these commits, include change log and upload released manifests'
```
SchSeba: maybe change this to tag the commit started with `Merge pull request <pr-number>` ?

phoracek: I know it is a little weird, but release should be on the first one, where we generate manifests for the new version. The second commit is starting next version.

Practically there shouldn't be a difference. If you think it makes more sense to keep it on top of the PR, I will change it.

phoracek:
/hold

I need to set container tag too.

PR #82, opened 2019-04-19
release 0.5.0

phoracek:


phoracek:
@SchSeba @booxter

manifests/cluster-network-addons/0.5.0/cluster-network-addons-operator.0.5.0.clusterserviceversion.yaml
```diff
@@ -0,0 +1,190 @@
...
+          "kind":"NetworkAddonsConfig",
+          "metadata": {
+            "name":"cluster"
+          },
+          "spec": {
+            "multus":{},
+            "linuxBridge":{},
+            "sriov":{},
+            "kubeMacPool": {
+              "rangeStart": "02:00:00:00:00:00",
```
SchSeba: maybe now we support the range to be auto generated we can drop this from the examples?

manifests/cluster-network-addons/0.5.0/cluster-network-addons-operator.0.5.0.clusterserviceversion.yaml
```diff
@@ -0,0 +1,190 @@
...
+                    value: quay.io/booxter/sriov-device-plugin:latest
+                  - name: SRIOV_CNI_IMAGE
+                    value: docker.io/nfvpe/sriov-cni:latest
+                  - name: SRIOV_ROOT_DEVICES
+                  - name: SRIOV_NETWORK_NAME
+                    value: sriov-network
+                  - name: SRIOV_NETWORK_TYPE
+                    value: sriov
+                  - name: KUBEMACPOOL_IMAGE
+                    value: quay.io/kubevirt/kubemacpool:latest
```
SchSeba: we don't want to point the v0.2.0 on this release?

manifests/cluster-network-addons/0.5.0/cluster-network-addons-operator.0.5.0.clusterserviceversion.yaml
```diff
@@ -0,0 +1,190 @@
...
+                  - name: OPERATOR_NAMESPACE
+                    valueFrom:
+                      fieldRef:
+                        fieldPath: metadata.namespace
+                  - name: POD_NAME
+                    valueFrom:
+                      fieldRef:
+                        fieldPath: metadata.name
+                  - name: WATCH_NAMESPACE
+                  image: quay.io/kubevirt/cluster-network-addons-operator:latest
```
SchSeba: I think the image tag here needs to be the same as the version no?

manifests/cluster-network-addons/0.5.0/cluster-network-addons-operator.0.5.0.clusterserviceversion.yaml
phoracek: Oh you are totally right! I have to fix the `make release` script

manifests/cluster-network-addons/0.5.0/cluster-network-addons-operator.0.5.0.clusterserviceversion.yaml
phoracek: Good point. Definitely can be dropped, but there is no reason to do this before this release.

manifests/cluster-network-addons/0.5.0/cluster-network-addons-operator.0.5.0.clusterserviceversion.yaml
phoracek: Sorry, missed that it was for kubemacpool. We need to set it in https://github.com/kubevirt/cluster-network-addons-operator/blob/master/pkg/components/components.go#L16

PR #88, opened 2019-04-23
Fix operator-sdk generate command

SchSeba:
I also add build/_output to the .gitignore file

---
This change is [Reviewable](https://reviewable.io/reviews/kubevirt/cluster-network-addons-operator/88)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@ _out
 # Temporary build files
 build/operator/_output
 build/operator/_test
+build/_output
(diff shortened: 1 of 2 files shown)


sjpotter:
works, but removes the apache license at the top of the generated code.

phoracek:
@sjpotter we are guilty of not keeping the license header in most (or all?) of our modules. Gotta fix that in a separate PR.

PR #92, opened 2019-05-01
Pin components version

SchSeba:
Pin network components for the operator.

After this PR I will open another one to make a release of version 0.6.0 for the operator.

---
This change is [Reviewable](https://reviewable.io/reviews/kubevirt/cluster-network-addons-operator/92)

### DIFF
--- a/pkg/components/components.go
+++ b/pkg/components/components.go
@@ -14,11 +14,11 @@ import (
 const Name = "cluster-network-addons-operator"
 
 const (
-	MultusImageDefault            = "docker.io/nfvpe/multus:latest"
+	MultusImageDefault            = "quay.io/kubevirt/cluster-network-addon-multus:v3.2.0-1.gitbf61002"
 	LinuxBridgeCniImageDefault    = "quay.io/kubevirt/cni-default-plugins:v0.1.0"
 	LinuxBridgeMarkerImageDefault = "quay.io/kubevirt/bridge-marker:0.1.0"
 	SriovDpImageDefault           = "quay.io/booxter/sriov-device-plugin:latest"
-	SriovCniImageDefault          = "docker.io/nfvpe/sriov-cni:latest"
+	SriovCniImageDefault          = "quay.io/kubevirt/cluster-network-addon-sriov-cni:v1.1.0-1.git9e4c973"
 	KubeMacPoolImageDefault       = "quay.io/kubevirt/kubemacpool:v0.2.0"
 )
 


booxter:
@SchSeba how did you come up with versions for images? (Also, how do I get access to quay to be able to push there in the future?)

AFAIU upstream DP has a 'release-v1' branch that includes an older version of the plugin. It probably makes sense to have our version be higher than v1 (not v0.1.0 as proposed) to reflect the fact that master is newer than the v1 older version.

SchSeba:
@booxter Let's do that I will tag the dp as v1.0.0 and the git commit.

But iirc the image we use was build from your fork no? 

What should be the git commit?

booxter:
@SchSeba we should build from upstream now, no reason to keep it forked. I would do it but, as I indicated in several places, I don't have permission to push images into newly created repos.

SchSeba:
@booxter Can we take the master from there dockerhub?
 
https://hub.docker.com/r/nfvpe/sriov-device-plugin/tags

booxter:
@SchSeba yes. Need to check if CNI has autobuilds set up. If they have them for master too, we can also use that.

SchSeba:
@booxter can you take a look ? I remove the sriov device plugin.

lets open a separate PR after you find the right version to tag.

pkg/components/components.go
```diff
@@ -14,11 +14,11 @@ import (
 const Name = "cluster-network-addons-operator"
 
 const (
-	MultusImageDefault            = "docker.io/nfvpe/multus:latest"
+	MultusImageDefault            = "quay.io/kubevirt/cluster-network-addon-multus:v3.2-1.gitbf61002"
 	LinuxBridgeCniImageDefault    = "quay.io/kubevirt/cni-default-plugins:v0.1.0"
 	LinuxBridgeMarkerImageDefault = "quay.io/kubevirt/bridge-marker:0.1.0"
 	SriovDpImageDefault           = "quay.io/booxter/sriov-device-plugin:latest"
-	SriovCniImageDefault          = "docker.io/nfvpe/sriov-cni:latest"
+	SriovCniImageDefault          = "quay.io/kubevirt/cluster-network-addon-sriov-cni:v1.1.0.git9e4c973"
```
phoracek: this one is missing release number, why?

SchSeba: sorry I don't understand this is not the release `v1.1.0.git9e4c973` ?

phoracek: I think it should be either `registry/image-name:vX.Y.Z` for released images, or `registry/image-name:vX.Y.Z-1.gitXXXXXXX` for unreleased. Version is the last released version, release number (1) will be bumped once we push a newer unreleased image and the git ID is there for a simple way to identify used sources.

SchSeba: ok no problem can you take another look now?

phoracek: Looks great, thank you

PR #93, opened 2019-05-02
Pin sriovdp to specific version

booxter:
This pin also bumps version to the latest in upstream master. It should
work fine because we already create sriovdp configuration file and the
latest image is not broken anymore. (It was at some point broken by
alpine change, which was the original reason to pin it in
kubevirt-ansible, from where the forked image migrated to this
operator).


pkg/components/components.go
```diff
@@ -17,7 +17,7 @@ const (
 	MultusImageDefault            = "docker.io/nfvpe/multus:latest"
 	LinuxBridgeCniImageDefault    = "quay.io/kubevirt/cni-default-plugins:v0.1.0"
 	LinuxBridgeMarkerImageDefault = "quay.io/kubevirt/bridge-marker:0.1.0"
-	SriovDpImageDefault           = "quay.io/booxter/sriov-device-plugin:latest"
+	SriovDpImageDefault           = "quay.io/kubevirt/cluster-network-addon-sriov-device-plugin:v2.0.0-1.git9a20829",
```
SchSeba: Can you remove the "," at the end of the line?

PR #99, opened 2019-05-14
Fill kubemacpool range in the fillDefaults function

SchSeba:
This PR moves the range generation from the render to the fillDefaults functions.

---
This change is [Reviewable](https://reviewable.io/reviews/kubevirt/cluster-network-addons-operator/99)

### DIFF
--- a/pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
+++ b/pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
@@ -241,7 +241,11 @@ func (r *ReconcileNetworkAddonsConfig) renderObjects(networkAddonsConfig *opv1al
 	}
 
 	// Fill all defaults explicitly
-	network.FillDefaults(&networkAddonsConfig.Spec, prev)
+	if err := network.FillDefaults(&networkAddonsConfig.Spec, prev); err != nil {
+		log.Printf("failed to fill defaults: %v", err)
+		errors.Wrapf(err, "failed to fill defaults: %v", err)
+		return objs, err
+	}
 
 	// Compare against previous applied configuration to see if this change
 	// is safe.
--- a/pkg/network/image-pull-policy.go
+++ b/pkg/network/image-pull-policy.go
@@ -21,7 +21,7 @@ func validateImagePullPolicy(conf *opv1alpha1.NetworkAddonsConfigSpec) []error {
 	return []error{}
 }
 
-func fillDefaultsImagePullPolicy(conf, previous *opv1alpha1.NetworkAddonsConfigSpec) {
+func fillDefaultsImagePullPolicy(conf, previous *opv1alpha1.NetworkAddonsConfigSpec) []error {
 	if conf.ImagePullPolicy == "" {
 		if previous != nil && previous.ImagePullPolicy != "" {
 			conf.ImagePullPolicy = previous.ImagePullPolicy
@@ -29,13 +29,15 @@ func fillDefaultsImagePullPolicy(conf, previous *opv1alpha1.NetworkAddonsConfigS
 			conf.ImagePullPolicy = defaultImagePullPolicy
 		}
 	}
+
+	return []error{}
 }
 
 func changeSafeImagePullPolicy(prev, next *opv1alpha1.NetworkAddonsConfigSpec) []error {
 	if prev.ImagePullPolicy != "" && prev.ImagePullPolicy != next.ImagePullPolicy {
 		return []error{errors.Errorf("cannot modify ImagePullPolicy configuration once components were deployed")}
 	}
-	return nil
+	return []error{}
 }
 
 // Verify if the value is a valid PullPolicy
--- a/pkg/network/kubemacpool.go
+++ b/pkg/network/kubemacpool.go
@@ -43,23 +43,22 @@ func validateKubeMacPool(conf *opv1alpha1.NetworkAddonsConfigSpec) []error {
 	return []error{}
 }
 
-func changeSafeKubeMacPool(prev, next *opv1alpha1.NetworkAddonsConfigSpec) []error {
-	if prev.KubeMacPool != nil && !reflect.DeepEqual(prev.KubeMacPool, next.KubeMacPool) {
-		return []error{errors.Errorf("cannot modify KubeMacPool configuration once it is deployed")}
-	}
-	return nil
-}
-
-// renderLinuxBridge generates the manifests of Linux Bridge
-func renderKubeMacPool(conf *opv1alpha1.NetworkAddonsConfigSpec, manifestDir string) ([]*unstructured.Unstructured, error) {
+func fillDefaultsKubeMacPool(conf, previous *opv1alpha1.NetworkAddonsConfigSpec) []error {
 	if conf.KubeMacPool == nil {
-		return nil, nil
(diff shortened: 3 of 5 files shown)


pkg/network/image-pull-policy.go
```diff
@@ -21,14 +21,16 @@ func validateImagePullPolicy(conf *opv1alpha1.NetworkAddonsConfigSpec) []error {
 	return []error{}
 }
 
-func fillDefaultsImagePullPolicy(conf, previous *opv1alpha1.NetworkAddonsConfigSpec) {
+func fillDefaultsImagePullPolicy(conf, previous *opv1alpha1.NetworkAddonsConfigSpec) error {
```
phoracek: We should use []error instead like we do in validate functions. So in case multiple components have invalid config, we report all at once.

pkg/network/kubemacpool.go
```diff
@@ -84,6 +71,28 @@ func renderKubeMacPool(conf *opv1alpha1.NetworkAddonsConfigSpec, manifestDir str
 	return objs, nil
 }
 
+func fillDefaultsKubeMacPool(conf, previous *opv1alpha1.NetworkAddonsConfigSpec) error {
```
phoracek: please move this below validateKubeMacPool, so it follows the execution order

pkg/network/network.go
```diff
@@ -37,8 +37,17 @@ func Validate(conf *opv1alpha1.NetworkAddonsConfigSpec, openshiftNetworkConfig *
...
 // Defaults are carried forward from previous if it is provided. This is so we
 // can change defaults as we move forward, but won't disrupt existing clusters.
-func FillDefaults(conf, previous *opv1alpha1.NetworkAddonsConfigSpec) {
-	fillDefaultsImagePullPolicy(conf, previous)
+func FillDefaults(conf, previous *opv1alpha1.NetworkAddonsConfigSpec) error {
+	err := fillDefaultsImagePullPolicy(conf, previous)
+	if err != nil {
+		return err
+	}
+	err = fillDefaultsKubeMacPool(conf, previous)
```
phoracek: nit: empty line before this row

pkg/network/network.go
```diff
@@ -37,8 +37,17 @@ func Validate(conf *opv1alpha1.NetworkAddonsConfigSpec, openshiftNetworkConfig *
 //
 // Defaults are carried forward from previous if it is provided. This is so we
 // can change defaults as we move forward, but won't disrupt existing clusters.
-func FillDefaults(conf, previous *opv1alpha1.NetworkAddonsConfigSpec) {
-	fillDefaultsImagePullPolicy(conf, previous)
+func FillDefaults(conf, previous *opv1alpha1.NetworkAddonsConfigSpec) error {
+	err := fillDefaultsImagePullPolicy(conf, previous)
```
phoracek: like in Validate, please collect all errors and return only in the end

pkg/network/kubemacpool.go
```diff
@@ -84,6 +71,28 @@ func renderKubeMacPool(conf *opv1alpha1.NetworkAddonsConfigSpec, manifestDir str
...
+	if conf.KubeMacPool.RangeStart == "" || conf.KubeMacPool.RangeEnd == "" {
+		if previous != nil && previous.KubeMacPool != nil {
```
phoracek: What if `kubeMacPool: {}`. Is it considered `nil` or empty structure?

pkg/network/kubemacpool.go
```diff
@@ -84,6 +71,28 @@ func renderKubeMacPool(conf *opv1alpha1.NetworkAddonsConfigSpec, manifestDir str
 	return objs, nil
 }
 
+func fillDefaultsKubeMacPool(conf, previous *opv1alpha1.NetworkAddonsConfigSpec) error {
+	if conf.KubeMacPool.RangeStart == "" || conf.KubeMacPool.RangeEnd == "" {
```
phoracek: please add an inline comment saying that if user hasn't explicitly requested a range, we try to reuse previously applied

pkg/network/kubemacpool.go
```diff
@@ -84,6 +71,28 @@ func renderKubeMacPool(conf *opv1alpha1.NetworkAddonsConfigSpec, manifestDir str
...
+		if previous != nil && previous.KubeMacPool != nil {
+			conf.KubeMacPool = previous.KubeMacPool
+			return nil
+		}
+
+		prefix, err := generateRandomMacPrefix()
```
phoracek: please add an inline comment that if no range was specified, we generated a random prefix

SchSeba:
@phoracek can you please make another review

pkg/network/image-pull-policy.go
```diff
@@ -21,14 +21,16 @@ func validateImagePullPolicy(conf *opv1alpha1.NetworkAddonsConfigSpec) []error {
...
+func fillDefaultsImagePullPolicy(conf, previous *opv1alpha1.NetworkAddonsConfigSpec) []error {
 	if conf.ImagePullPolicy == "" {
 		if previous != nil && previous.ImagePullPolicy != "" {
 			conf.ImagePullPolicy = previous.ImagePullPolicy
 		} else {
 			conf.ImagePullPolicy = defaultImagePullPolicy
 		}
 	}
+
+	return nil
```
phoracek: this needs to be []error{}, this would fail (right?)

pkg/network/kubemacpool.go
```diff
@@ -69,6 +68,22 @@ func renderKubeMacPool(conf *opv1alpha1.NetworkAddonsConfigSpec, manifestDir str
...
 	}
 
+	return []error{}
+}
+
+func changeSafeKubeMacPool(prev, next *opv1alpha1.NetworkAddonsConfigSpec) []error {
+	if prev.KubeMacPool != nil && !reflect.DeepEqual(prev.KubeMacPool, next.KubeMacPool) {
+		return []error{errors.Errorf("cannot modify KubeMacPool configuration once it is deployed")}
+	}
+	return nil
```
phoracek: Oh I see it was used before. Wonder how that works with append

phoracek: Oh yes, we should, otherwise we endup with []error{<nil>}. If you don't mind

SchSeba: Done :) can you please take a look?

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
```diff
@@ -241,7 +241,11 @@ func (r *ReconcileNetworkAddonsConfig) renderObjects(networkAddonsConfig *opv1al
 	}
 
 	// Fill all defaults explicitly
-	network.FillDefaults(&networkAddonsConfig.Spec, prev)
+	if err := network.FillDefaults(&networkAddonsConfig.Spec, prev); err != nil {
+		log.Printf("failed to fill defaults: %v", err)
+		errors.Wrapf(err, "failed to fill defaults: %v", err)
```
phoracek: looks like a bug, we don't use the return value. i don't think that err is updated in place

SchSeba: Yes you are right, Can I create a new PR to fix this?

PR #104, opened 2019-05-16
Fix nmstate crd

phoracek:
Fix missing pieces from nmstate component.


phoracek:
@SchSeba @qinqon could you please take a look? We still miss NetworkManager on the local cluster thought.

data/nmstate/003-nmstated.yaml
```diff
@@ -1,5 +1,5 @@
 ---
-apiVersion: extensions/v1beta1
+apiVersion: apps/v1
```
SchSeba:
```
message: 'could not apply (apps/v1, Kind=DaemonSet) nmstate/nm-state-controller:
could not create (apps/v1, Kind=DaemonSet) nmstate/nm-state-controller: DaemonSet.apps
"nm-state-controller" is invalid: spec.template.metadata.labels: Invalid value:
map[string]string{"nmstate.io":"state-controller"}: `selector` does not match
template `labels`'
```

With the new version you should add the same label in the spec section

phoracek: Oh yeah, you are right. Thanks

phoracek:
@qinqon we can introduce some templating, but not in this patch since it would require way more changes. Also, I'm not so sure about it since we still have to pull manifests from third-party repository.

The rename is so we create CRD first and the daemon only at the end. Not sure it matters, but in all the other components we start DS last. For the sake of consistency.

data/nmstate/003-nmstated.yaml
phoracek: Fixed, could you please take another look?

phoracek:
ci test please

SchSeba:
ci test please

phoracek:
@SchSeba could you check again?

PR #106, opened 2019-05-16
fix nmstate version to v0.1.0

phoracek:
We need to use fixed images version, otherwise, a change in :latest
component's image might be devastating in combination with manifests
shipped by us.

### DIFF
--- a/pkg/components/components.go
+++ b/pkg/components/components.go
@@ -20,7 +20,7 @@ const (
 	SriovDpImageDefault             = "quay.io/kubevirt/cluster-network-addon-sriov-device-plugin:v2.0.0-1.git9a20829"
 	SriovCniImageDefault            = "quay.io/kubevirt/cluster-network-addon-sriov-cni:v1.1.0-1.git9e4c973"
 	KubeMacPoolImageDefault         = "quay.io/kubevirt/kubemacpool:v0.3.0"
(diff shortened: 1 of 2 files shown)


phoracek:
@SchSeba @qinqon could you please take a look?

pkg/components/components.go
```diff
@@ -20,7 +20,7 @@ const (
 	SriovDpImageDefault             = "quay.io/kubevirt/cluster-network-addon-sriov-device-plugin:v2.0.0-1.git9a20829"
 	SriovCniImageDefault            = "quay.io/kubevirt/cluster-network-addon-sriov-cni:v1.1.0-1.git9e4c973"
 	KubeMacPoolImageDefault         = "quay.io/kubevirt/kubemacpool:v0.3.0"
-	NMStateStateHandlerImageDefault = "quay.io/nmstate/kubernetes-nmstate-state-handler:latest"
+	NMStateStateHandlerImageDefault = "quay.io/nmstate/kubernetes-nmstate-state-handler:0.1"
```
qinqon: Maybe more semver format friendly v0.1.0 ?

phoracek: oh I forgot about Z! Personally, I prefer versions without the leading v. It is not even a part of semver AFAIK. What do you think?

PR #108, opened 2019-05-16
update manifests templates with nmstate changes

phoracek:
We forgot to update our templates.


phoracek:

phoracek:
Skipping CI since the change does not affect it at all.

PR #109, opened 2019-05-16
increase timeout for functional tests

phoracek:
Now when we added nmstate component to the operator, deployment
takes even more time and we need to increase test timeout in
order to make it pass on STD-CI.


phoracek:

PR #112, opened 2019-05-27
test render.go module

phoracek:


phoracek:
@SchSeba @qinqon

pkg/render/render_suite_test.go
```diff
@@ -0,0 +1,13 @@
+package render
```
qinqon: Let's use different package for tests to explicitly include stuff we want to test "render_test"

pkg/render/render_test.go
```diff
@@ -0,0 +1,114 @@
...
+		]
+	}
+}
+`
+			Expect(renderedFromYAML[0].MarshalJSON()).To(MatchJSON(expected))
+		})
+	})
+
+	Context("when a identical YAML and JSONs manifest are given as a parameters", func() {
+		It("should return identical objects", func() {
```
qinqon: Isn't this identical to previous test  more or less ?

pkg/render/render_test.go
```diff
@@ -0,0 +1,114 @@
...
+var _ = Describe("RenderTemplate", func() {
+	Context("when a valid YAML manifest is given as a parameter", func() {
+		It("should successfully render", func() {
+			renderData := MakeRenderData()
+
+			renderedFromYAML, err := RenderTemplate("testdata/simple.yaml", &renderData)
+			Expect(err).NotTo(HaveOccurred())
+
+			Expect(renderedFromYAML).To(HaveLen(1))
+			expected := `
```
qinqon: As I understand this is testing RenderTemplate that converts yaml manifest to the structs from the api, so it's better to write the "expected"·data as structs too.

pkg/render/render_test.go
```diff
@@ -0,0 +1,114 @@
...
+	Context("when YAML manifest with multiple objects given as a parameter", func() {
+		It("should successfully render list of objects", func() {
+			renderData := MakeRenderData()
+
+			rendered, err := RenderTemplate("testdata/multiple.yaml", &renderData)
+			Expect(err).NotTo(HaveOccurred())
+
+			Expect(rendered).To(HaveLen(3))
+
+			Expect(rendered[0].GetObjectKind().GroupVersionKind().String()).To(Equal("/v1, Kind=Pod"))
```
qinqon: Order here is not relevant also I suppose we need to compare with all the fields not just GroupVersionKind

pkg/render/render_test.go
```diff
@@ -0,0 +1,114 @@
...
+			Expect(rendered[1].GetObjectKind().GroupVersionKind().String()).To(Equal("rbac.authorization.k8s.io/v1, Kind=ClusterRoleBinding"))
+			Expect(rendered[2].GetObjectKind().GroupVersionKind().String()).To(Equal("/v1, Kind=ConfigMap"))
+		})
+	})
+
+	Context("when YAML manifest including templating given as a parameter", func() {
+		Context("when requested functions are missing", func() {
+			It("should fail with expected error", func() {
+				renderData := MakeRenderData()
```
qinqon: Shouldn't we add ' renderData.Data["Namespace"] = "myns" ' here to ensure tha the failure is just function not defined ?

Or maybe having template.yaml with fun and other with entry.

pkg/render/render_test.go
```diff
@@ -0,0 +1,114 @@
...
+
+		Context("when all functions variables are provided", func() {
+			It("should successfully render", func() {
+				renderData := MakeRenderData()
+				renderData.Data["Namespace"] = "myns"
+				renderData.Funcs["fname"] = func(s string) string { return "test-" + s }
+				rendered, err := RenderTemplate("testdata/template.yaml", &renderData)
+				Expect(err).NotTo(HaveOccurred())
+
+				Expect(rendered[0].GetName()).To(Equal("test-podname"))
```
qinqon: Let create the full struct to compare

pkg/render/render_test.go
```diff
@@ -0,0 +1,114 @@
...
+var _ = Describe("RenderDir", func() {
+	Context("rendering all files in a directory", func() {
+		It("should successfully render all JSON/YAML manifest, including templates", func() {
+			renderData := MakeRenderData()
+			renderData.Funcs["fname"] = func(s string) string { return s }
+			renderData.Data["Namespace"] = "myns"
+
+			rendered, err := RenderDir("testdata", &renderData)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(rendered).To(HaveLen(6))
```
qinqon: We can create 6 len slice expectation and use all of it or part of in in the test suites to compare.

pkg/render/render_suite_test.go
phoracek: I'd love to test it as a proper library. But for e.g. tests under pkg/network/ we cannot really do it, they have private functions like validate and more. Actually, let me change it here and then rework pkg/network to be public and properly organized (and tested) later.

phoracek: Done

pkg/render/render_test.go
phoracek: This one is making sure JSON works fine. Would you write it differently?

phoracek: Actually, I should just read JSON and compare it to expected struct

phoracek: Done

pkg/render/render_test.go
phoracek: Done

pkg/render/render_test.go
phoracek: The rest of the fields is not important, we test that in previous (single pod) tests. This one is just to make sure all objects are there.

phoracek: You are right that the order is not important, but I'd say it is not worth making the test know it. Given test files is static and part of the test. If you think there is a benefit in making it more robust, tell me.

pkg/render/render_test.go
phoracek: Done

pkg/render/render_test.go
phoracek: Done

pkg/render/render_test.go
phoracek: If you don't mind, I'd rather keep it minimal for now. Changes we will change render.go are quite small and I'd rather not spend too much time on testing it. Also, once we have that slice, you would argue it might be worth it using it to populate testdata/ and that would be too much. Tell me if I'm just being lazy.

PR #117, opened 2019-05-31
test network package

phoracek:
Finish test coverage for pkg/network. This patch also fixes a bug in multus
validation where we tried to compared bool pointers.

### DIFF
--- a/pkg/network/image-pull-policy_test.go
+++ b/pkg/network/image-pull-policy_test.go
@@ -1,0 +1,113 @@
+package network
+
+import (
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	v1 "k8s.io/api/core/v1"
+
+	opv1alpha1 "github.com/kubevirt/cluster-network-addons-operator/pkg/apis/networkaddonsoperator/v1alpha1"
+)
+
+var _ = Describe("Testing image-pull-policy", func() {
+	Describe("validateImagePullPolicy", func() {
+		Context("when configuration uses invalid policy type", func() {
+			spec := &opv1alpha1.NetworkAddonsConfigSpec{ImagePullPolicy: v1.PullPolicy("BAD")}
+
+			It("should fail", func() {
+				errorList := validateImagePullPolicy(spec)
+				Expect(errorList).To(HaveLen(1), "validation failed due to an unexpected error: %v", errorList)
+				Expect(errorList[0]).To(MatchError("requested imagePullPolicy 'BAD' is not valid"))
+			})
+		})
+
+		Context("when configuration uses a valid policy type", func() {
+			spec := &opv1alpha1.NetworkAddonsConfigSpec{ImagePullPolicy: v1.PullAlways}
+
+			It("should pass", func() {
+				errorList := validateImagePullPolicy(spec)
+				Expect(errorList).To(BeEmpty())
+			})
+		})
+	})
+
+	Describe("fillDefaultsImagePullPolicy", func() {
+		Context("when no policy specified", func() {
+			Context("and there was a policy specified in the previous config", func() {
+				new := &opv1alpha1.NetworkAddonsConfigSpec{}
+				prev := &opv1alpha1.NetworkAddonsConfigSpec{ImagePullPolicy: v1.PullAlways}
+
+				It("should successfully pass", func() {
+					errorList := fillDefaultsImagePullPolicy(new, prev)
+					Expect(errorList).To(BeEmpty())
+				})
+
+				It("and fill in the previously defined policy", func() {
+					Expect(new.ImagePullPolicy).To(Equal(v1.PullAlways))
+				})
+			})
+
+			Context("and there was no policy specified in the last config", func() {
+				new := &opv1alpha1.NetworkAddonsConfigSpec{}
+				prev := &opv1alpha1.NetworkAddonsConfigSpec{}
+
+				It("should successfully pass", func() {
+					errorList := fillDefaultsImagePullPolicy(new, prev)
+					Expect(errorList).To(BeEmpty())
+				})
+
+				It("should fill in the default policy", func() {
+					Expect(new.ImagePullPolicy).To(Equal(defaultImagePullPolicy))
+					// Following is in an ideal case a duplicate of previous line.
+					// The reason we have this check is, that some tests in this module
+					// are expecting this to be the default and they would need some
+					// changes in case the default changes.
+					Expect(new.ImagePullPolicy).To(Equal(v1.PullIfNotPresent))
+				})
+			})
+		})
+	})
+
+	Describe("changeSafeImagePullPolicy", func() {
+		Context("when it is kept disabled", func() {
+			prev := &opv1alpha1.NetworkAddonsConfigSpec{}
+			new := &opv1alpha1.NetworkAddonsConfigSpec{}
+
+			It("should pass", func() {
+				errorList := changeSafeImagePullPolicy(prev, new)
+				Expect(errorList).To(BeEmpty())
+			})
+		})
+
+		Context("when there is no previous value", func() {
+			prev := &opv1alpha1.NetworkAddonsConfigSpec{}
+			new := &opv1alpha1.NetworkAddonsConfigSpec{LinuxBridge: &opv1alpha1.LinuxBridge{}}
+
+			It("should accept any configuration", func() {
+				errorList := changeSafeImagePullPolicy(prev, new)
+				Expect(errorList).To(BeEmpty())
+			})
+		})
+
+		Context("when the previous and new configuration match", func() {
+			prev := &opv1alpha1.NetworkAddonsConfigSpec{LinuxBridge: &opv1alpha1.LinuxBridge{}}
+			new := &opv1alpha1.NetworkAddonsConfigSpec{LinuxBridge: &opv1alpha1.LinuxBridge{}}
+
+			It("should accept the configuration", func() {
+				errorList := changeSafeImagePullPolicy(prev, new)
+				Expect(errorList).To(BeEmpty())
+			})
+		})
+
+		Context("when there is previous value, but the new one is empty (removing component)", func() {
+			prev := &opv1alpha1.NetworkAddonsConfigSpec{LinuxBridge: &opv1alpha1.LinuxBridge{}}
+			new := &opv1alpha1.NetworkAddonsConfigSpec{}
+
+			// If ImagePullPolicy is omitted, default or previously applied will be used
+			It("should pass", func() {
+				errorList := changeSafeImagePullPolicy(prev, new)
+				Expect(errorList).To(BeEmpty())
+			})
+		})
+	})
+})
--- a/pkg/network/linux-bridge_test.go
+++ b/pkg/network/linux-bridge_test.go
@@ -1,0 +1,49 @@
+package network
+
+import (
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	opv1alpha1 "github.com/kubevirt/cluster-network-addons-operator/pkg/apis/networkaddonsoperator/v1alpha1"
+)
+
+var _ = Describe("Testing linux-bridge", func() {
+	Describe("changeSafeLinuxBridge", func() {
+		Context("when it is kept disabled", func() {
+			prev := &opv1alpha1.NetworkAddonsConfigSpec{}
+			new := &opv1alpha1.NetworkAddonsConfigSpec{}
+			It("should pass", func() {
+				errorList := changeSafeLinuxBridge(prev, new)
+				Expect(errorList).To(BeEmpty())
+			})
+		})
+
+		Context("when there is no previous value", func() {
+			prev := &opv1alpha1.NetworkAddonsConfigSpec{}
+			new := &opv1alpha1.NetworkAddonsConfigSpec{LinuxBridge: &opv1alpha1.LinuxBridge{}}
+			It("should accept any configuration", func() {
+				errorList := changeSafeLinuxBridge(prev, new)
+				Expect(errorList).To(BeEmpty())
+			})
+		})
+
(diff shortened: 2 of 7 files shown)


phoracek:
@booxter @SchSeba @qinqon could you please take a look?

pkg/network/sriov_test.go
```diff
@@ -0,0 +1,156 @@
...
+			sriovRequested := false
+			openshiftNetworkOperatorRunning := false
+			openshiftSriovEnabled := false
+			It("should pass", func() {
+				errorList := _validateSriov(sriovRequested, openshiftNetworkOperatorRunning, openshiftSriovEnabled)
+				Expect(errorList).To(BeEmpty())
+			})
+		})
+
+		Context("when openshfit network operator is running", func() {
```
booxter: openshift

pkg/network/sriov_test.go
```diff
@@ -0,0 +1,156 @@
...
+						errorList := _validateSriov(sriovRequested, openshiftNetworkOperatorRunning, openshiftSriovEnabled)
+						Expect(errorList).To(BeEmpty())
+					})
+				})
+				Context("and configuration requests sriov", func() {
+					sriovRequested := true
+					It("should fail with an error message", func() {
+						errorList := _validateSriov(sriovRequested, openshiftNetworkOperatorRunning, openshiftSriovEnabled)
+						Expect(len(errorList)).To(Equal(1), "validation failed due to an unexpected error: %v", errorList)
+						Expect(errorList[0].Error()).To(Equal("SR-IOV has been requested, but it's not compatible with OpenShift Cluster Network Operator SR-IOV support"))
```
booxter: (no action required) personally not a fan of matching against exact error messages but I guess it's subjective

pkg/network/sriov_test.go
```diff
@@ -0,0 +1,156 @@
...
+					sriovRequested := true
+					It("should pass", func() {
+						errorList := _validateSriov(sriovRequested, openshiftNetworkOperatorRunning, openshiftSriovEnabled)
+						Expect(errorList).To(BeEmpty())
+					})
+				})
+			})
+		})
+
+		Context("when openshfit network operator is not running", func() {
```
booxter: openshift

pkg/network/sriov_test.go
```diff
@@ -0,0 +1,156 @@
...
+			openshiftNetworkOperatorRunning := false
+			openshiftNetworkOperatorDisableMultiNetwork := false
```
booxter: the name of this variable differs from what is used above, and it doesn't seem like matching the meaning of the helper function parameter (?): the function parameter means "whether openshift sriov network is enabled" while the variable name suggests the meaning is "whether multi-network support (?) is disabled"?

pkg/network/nmstate_test.go
```diff
@@ -0,0 +1,49 @@
...
+		Context("when there is no previous value", func() {
+			prev := &opv1alpha1.NetworkAddonsConfigSpec{}
+			new := &opv1alpha1.NetworkAddonsConfigSpec{NMState: &opv1alpha1.NMState{}}
+			It("should accept any configuration", func() {
+				errorList := changeSafeNMState(prev, new)
+				Expect(errorList).To(BeEmpty())
+			})
+		})
+
+		Context("when the previous and new configuration matches", func() {
```
booxter: matches -> match

pkg/network/network_test.go
```diff
@@ -0,0 +1,128 @@
...
+				err := Validate(conf, openshiftNetworkConf)
+				Expect(err).To(HaveOccurred())
+				Expect(err.Error()).To(ContainSubstring("invalid configuration"))
+				Expect(err.Error()).To(ContainSubstring("both or none of the KubeMacPool ranges needs to be configured"))
+				Expect(err.Error()).To(ContainSubstring("requested imagePullPolicy 'bar' is not valid"))
+			})
+		})
+	})
+
+	Describe("FillDefaults", func() {
```
booxter: I probably see why you don't have a negative case here (because the only way to receive an error from the function is if rand.Read() fails, which is hard to simulate); but in case you care to cover all branches (which seems like you effectively do with this PR), you could introduce a test mock / helper function for rand.Read that you could replace for test scenarios to trigger an error.

pkg/network/network_test.go
```diff
@@ -0,0 +1,128 @@
...
+	})
+
+	Describe("Render", func() {
+		Context("when given valid arguments", func() {
+			conf := &opv1alpha1.NetworkAddonsConfigSpec{ImagePullPolicy: v1.PullAlways, Multus: &opv1alpha1.Multus{}, LinuxBridge: &opv1alpha1.LinuxBridge{}}
+			manifestDir := "../../data"
+			openshiftNetworkConf := &osv1.Network{}
+			clusterInfo := &ClusterInfo{SCCAvailable: true, OpenShift4: false}
+
+			It("should successfully render a set of objects", func() {
```
booxter: perhaps you could add a negative case here, this seems like an important case to cover.

pkg/network/multus.go
```diff
@@ -21,7 +21,7 @@ func validateMultus(conf *opv1alpha1.NetworkAddonsConfigSpec, openshiftNetworkCo
 	}
 
 	if openshiftNetworkConfig != nil {
-		if openshiftNetworkConfig.Spec.DisableMultiNetwork == newTrue() {
```
booxter: perhaps you should kill newTrue now then? it's not used anywhere else.

pkg/network/multus_test.go
```diff
@@ -0,0 +1,147 @@
...
+			conf := &opv1alpha1.NetworkAddonsConfigSpec{}
+			if multusRequested {
+				conf.Multus = &opv1alpha1.Multus{}
+			}
+
+			var openshiftNetworkConfig *osv1.Network
+			if openshiftNetworkOperatorRunning {
+				openshiftNetworkConfig = &osv1.Network{}
+				if openshiftNetworkOperatorDisableMultiNetwork {
+					newTrue := func() *bool {
```
booxter: I probably just don't know Golang well but what's this construction for? Can't we just have it as a simple struct field?

pkg/network/multus_test.go
```diff
@@ -0,0 +1,147 @@
...
+			multusRequested := false
+			openshiftNetworkOperatorRunning := false
+			openshiftNetworkOperatorDisableMultiNetwork := false
+			It("should pass", func() {
+				errorList := _validateMultus(multusRequested, openshiftNetworkOperatorRunning, openshiftNetworkOperatorDisableMultiNetwork)
+				Expect(errorList).To(BeEmpty())
+			})
+		})
+
+		Context("when openshfit network operator is running", func() {
```
booxter: openshift

pkg/network/multus_test.go
```diff
@@ -0,0 +1,147 @@
...
+					multusRequested := true
+					It("should pass", func() {
+						errorList := _validateMultus(multusRequested, openshiftNetworkOperatorRunning, openshiftNetworkOperatorDisableMultiNetwork)
+						Expect(errorList).To(BeEmpty())
+					})
+				})
+			})
+		})
+
+		Context("when openshfit network operator is not running", func() {
```
booxter: openshift

pkg/network/multus_test.go
```diff
@@ -0,0 +1,147 @@
...
+		Context("when there is no previous value", func() {
+			prev := &opv1alpha1.NetworkAddonsConfigSpec{}
+			new := &opv1alpha1.NetworkAddonsConfigSpec{Multus: &opv1alpha1.Multus{}}
+			It("should accept any configuration", func() {
+				errorList := changeSafeMultus(prev, new)
+				Expect(errorList).To(BeEmpty())
+			})
+		})
+
+		Context("when the previous and new configuration matches", func() {
```
booxter: matches -> match

pkg/network/linux-bridge_test.go
```diff
@@ -0,0 +1,49 @@
...
+		Context("when there is no previous value", func() {
+			prev := &opv1alpha1.NetworkAddonsConfigSpec{}
+			new := &opv1alpha1.NetworkAddonsConfigSpec{LinuxBridge: &opv1alpha1.LinuxBridge{}}
+			It("should accept any configuration", func() {
+				errorList := changeSafeLinuxBridge(prev, new)
+				Expect(errorList).To(BeEmpty())
+			})
+		})
+
+		Context("when the previous and new configuration matches", func() {
```
booxter: matches -> match

pkg/network/image-pull-policy_test.go
```diff
@@ -0,0 +1,108 @@
...
+			spec := &opv1alpha1.NetworkAddonsConfigSpec{ImagePullPolicy: v1.PullPolicy("BAD")}
+
+			It("should fail", func() {
+				errorList := validateImagePullPolicy(spec)
+				Expect(len(errorList)).To(Equal(1), "validation failed due to an unexpected error: %v", errorList)
+				Expect(errorList[0].Error()).To(Equal("requested imagePullPolicy 'BAD' is not valid"))
+			})
+		})
+
+		Context("when configration uses a valid policy type", func() {
```
booxter: configuration

pkg/network/image-pull-policy_test.go
```diff
@@ -0,0 +1,108 @@
...
+				It("and fill in the previously defined policy", func() {
+					Expect(new.ImagePullPolicy).To(Equal(v1.PullAlways))
+				})
+			})
+
+			Context("and there was no policy specified in the last config", func() {
+				new := &opv1alpha1.NetworkAddonsConfigSpec{}
+				prev := &opv1alpha1.NetworkAddonsConfigSpec{}
+
+				It("should successfullt pass", func() {
```
booxter: successfully

pkg/network/image-pull-policy_test.go
```diff
@@ -0,0 +1,108 @@
...
+				new := &opv1alpha1.NetworkAddonsConfigSpec{}
+				prev := &opv1alpha1.NetworkAddonsConfigSpec{ImagePullPolicy: v1.PullAlways}
+
+				It("should successfully pass", func() {
+					errorList := fillDefaultsImagePullPolicy(new, prev)
+					Expect(errorList).To(BeEmpty())
+				})
+
+				It("and fill in the previously defined policy", func() {
+					Expect(new.ImagePullPolicy).To(Equal(v1.PullAlways))
```
booxter: while we know that default policy is not Always, it's not strictly speaking a guarantee. And if for some reason it would be made into Always then this test case would not be distinguishable from behavior of e.g. filling it the result with default (now Always) value.

pkg/network/image-pull-policy_test.go
```diff
@@ -0,0 +1,108 @@
...
+			prev := &opv1alpha1.NetworkAddonsConfigSpec{}
+			new := &opv1alpha1.NetworkAddonsConfigSpec{LinuxBridge: &opv1alpha1.LinuxBridge{}}
+
+			It("should accept any configuration", func() {
+				errorList := changeSafeImagePullPolicy(prev, new)
+				Expect(errorList).To(BeEmpty())
+			})
+		})
+
+		Context("when the previous and new configuration matches", func() {
```
booxter: matches -> match

pkg/network/sriov_test.go
phoracek: Is there something like "error type" in Go? Like checking Exception class in Python?

pkg/network/sriov_test.go
phoracek: I see, I instantiated it just so there is a value, it doesn't matter whether it's true or false, since operator is disabled. Adding a comment.

openshift's Network type has attribute DisableMultiNetwork. I don't like negative parameters, so I'm trying to hide it from upper level (eyes of a developer running these tests).

pkg/network/network_test.go
phoracek: I will just add a TODO, not that I am lazy, but I want to leave a low-hanging fruit for others (or future me)

pkg/network/network_test.go
phoracek: Done for invalid manifestsDir

pkg/network/multus_test.go
phoracek: You cannot assign &true. You have to do this extra step to get pointer to bool. Don't know why, don't know about any other nicer way.

pkg/network/image-pull-policy_test.go
phoracek: Probably my bad specifying first the new config and then old. We set PullAlways in `prev`, we don't specify any policy in `new`. There is no defaulting. Or maybe I don't understand your comment.

phoracek:
@booxter @qinqon addressed your comments. Resolved most.

@qinqon regarding tables - I don't like that it doesn't allow you to include `It` to it (right? or is it Context). Anyway, we can do refactoring of that in a different PR, I don't really want to touch it now if you don't mind.

Guys, sorry about being lazy ignoring some of your comments. Although I appreciate them, I'd like to get over this PR. But if you feel like fighting for those, please do!

pkg/network/sriov_test.go
booxter: I don't think it can be directly mimicked. I don't know if there is a good solution. Is the exact message important enough to check in the first place?

pkg/network/sriov_test.go
booxter: it seems to me that the variable name doesn't belong here and was copy pasted from another test module (for multus).

pkg/network/sriov_test.go
phoracek: I'd say it does no harm. I'd rather change the message at two places than have ambiguous test.

pkg/network/sriov_test.go
phoracek: Should be better now

qinqon:
LGTM awesome stuff

PR #118, opened 2019-05-31
Functional tests

phoracek:
This initial set of functions tests covers:
- Check that invalid config triggers Failing condition
- That applying valid config over invalid recovers Failing
  to Running
- Deployment of each individual component
- Deployment of all the components, one by one
- Deployment of all the components at once

Note that #138 #139 #140 #124 #123 #120 and #119 has to be addressed first for all tests to pass.


phoracek:
@booxter @qinqon @SchSeba could you please take a look?

phoracek:
I will do some refactoring on deployment_test.go, plus a fix of Sriov test.

phoracek:
Half of the time is spent on removal of deployed components, probably due to graceful shutdown. Will try to force shorted period to make tests faster.

test/log/log.go
```diff
@@ -0,0 +1,27 @@
+package log
```
qinqon: All this is already implemente by think go 

For example Step would be By("openning connection...  blahblah")

test/check/check.go
```diff
@@ -0,0 +1,270 @@
...
+		if err != nil {
+			panic(err)
+		}
+		return fmt.Errorf("DaemonSet %s/%s is not ready, current state:\n%v", namespace, name, string(manifest))
+	}
+
+	return nil
+}
+
+func waitForNamespaceRemoval(name string, timeout time.Duration) error {
```
qinqon: We have [waitForDeletion](https://github.com/operator-framework/operator-sdk/blob/master/pkg/test/e2eutil/wait_util.go#L74) at operator-sdk testing, maybe we can use that to remove scaffolding here.

test/check/check.go
```diff
@@ -0,0 +1,270 @@
...
+	return wait.PollImmediate(time.Second, timeout, func() (bool, error) {
+		err := framework.Global.Client.Get(context.TODO(), types.NamespacedName{Name: name}, &corev1.Namespace{})
+		if apierrors.IsNotFound(err) {
+			return true, nil
+		}
+		return false, err
+	})
+}
+
+func waitForClusterRoleRemoval(name string, timeout time.Duration) error {
```
qinqon: ditto

test/check/check.go
```diff
@@ -0,0 +1,270 @@
...
+	return wait.PollImmediate(time.Second, timeout, func() (bool, error) {
+		err := framework.Global.Client.Get(context.TODO(), types.NamespacedName{Name: name}, &rbacv1.ClusterRole{})
+		if apierrors.IsNotFound(err) {
+			return true, nil
+		}
+		return false, err
+	})
+}
+
+func waitForClusterRoleBindingRemoval(name string, timeout time.Duration) error {
```
qinqon: ditto

test/check/check.go
```diff
@@ -0,0 +1,270 @@
...
+	return wait.PollImmediate(time.Second, timeout, func() (bool, error) {
+		err := framework.Global.Client.Get(context.TODO(), types.NamespacedName{Name: name}, &rbacv1.ClusterRoleBinding{})
+		if apierrors.IsNotFound(err) {
+			return true, nil
+		}
+		return false, err
+	})
+}
+
+func waitForSecurityContextConstraintsRemoval(name string, timeout time.Duration) error {
```
qinqon: ditto

test/check/check.go
```diff
@@ -0,0 +1,270 @@
...
+
+func checkForSecurityContextConstraints(name string) error {
+	err := framework.Global.Client.Get(context.TODO(), types.NamespacedName{Name: name}, &securityapi.SecurityContextConstraints{})
+	if strings.Contains(err.Error(), "no kind is registered for the type") {
+		return nil
+	}
+	return err
+}
+
+func checkForDeployment(name string, namespace string) error {
```
qinqon: Can we use operator-sdk [waitForDeployment](https://github.com/operator-framework/operator-sdk/blob/master/pkg/test/e2eutil/wait_util.go#L46) ?

test/check/check.go
```diff
@@ -0,0 +1,270 @@
...
+	}
+
+	if component.SecurityContextConstraints != "" {
+		errsAppend(waitForSecurityContextConstraintsRemoval(component.SecurityContextConstraints, 10*time.Second))
+	}
+
+	return errsToErr(errs)
+}
+
+func errsToErr(errs []error) error {
```
qinqon: We have all this with gomega

test/check/check.go
```diff
@@ -0,0 +1,270 @@
...
+	}
+
+	for _, deployment := range component.Deployments {
+		errsAppend(checkForDeployment(deployment, component.Namespace))
+	}
+
+	return errsToErr(errs)
+}
+
+func waitForComponentRemoval(component *Component) error {
```
qinqon: ditto

test/check/check.go
```diff
@@ -0,0 +1,270 @@
...
+func CheckConfigCondition(conditionType ConditionType, conditionStatus ConditionStatus, timeout time.Duration) error {
+	if err := waitForCondition(opv1alpha1.NetworkAddonsConditionType(conditionType), corev1.ConditionStatus(conditionStatus), timeout); err != nil {
+		log.Fail("Condition %q has not reached desired state %q within given timeout", conditionType, conditionStatus)
+		return err
+	}
+	log.Ok("Condition %q has reached desired state %q", conditionType, conditionStatus)
+	return nil
+}
+
+func checkForComponent(component *Component) error {
```
qinqon: It's possible to pass a function here and the function declaration to be runtime.Object so we remove the if statement.

Another approach would be the typical Runnable interface and pass it to this function.

test/check/check.go
```diff
@@ -0,0 +1,270 @@
...
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
+
+	opv1alpha1 "github.com/kubevirt/cluster-network-addons-operator/pkg/apis/networkaddonsoperator/v1alpha1"
+	"github.com/kubevirt/cluster-network-addons-operator/pkg/names"
+	"github.com/kubevirt/cluster-network-addons-operator/test/log"
+)
+
+func Assert(t *testing.T, err error) {
```
qinqon: Let's just use gomega

cluster/sync.sh
```diff
@@ -23,3 +23,4 @@ done
 ./cluster/kubectl.sh create -f _out/cluster-network-addons/${VERSION}/namespace.yaml
 ./cluster/kubectl.sh create -f _out/cluster-network-addons/${VERSION}/network-addons-config.crd.yaml
 ./cluster/kubectl.sh create -f _out/cluster-network-addons/${VERSION}/operator.yaml
+./cluster/kubectl.sh -n cluster-network-addons-operator wait deployment cluster-network-addons-operator --for condition=Available --timeout=600s
```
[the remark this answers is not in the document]
phoracek: We don't follow operators manifest structure nor list RBAC requirements in the source code. Currently, this is the only way to test it.

test/check/check.go
phoracek: I don't want to poll. Check calls is to be called after Ready condition turns True. At that moment, all components should be available. If they aren't, it is a bug in our condition handling.

test/check/check.go
phoracek: I don't want to use operator-sdk wait methods since they use Poll, not PollImmediate. i.e. they wait before the first check, that slows down tests quite a lot.

phoracek:
Thanks @qinqon, will address that

test/check/check.go
```diff
@@ -0,0 +1,270 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+}
+
+func CheckComponentsDeployment(components []Component) error {
+	for _, component := range components {
+		if err := checkForComponent(&component); err != nil {
+			log.Fail("Component %q has not created", component.ComponentName)
```
booxter: hasn't been created

test/check/check.go
```diff
@@ -0,0 +1,270 @@
...
+func waitForComponentRemoval(component *Component) error {
+	errs := []error{}
+	errsAppend := func(err error) {
+		if err != nil {
+			errs = append(errs, err)
+		}
+	}
+
+	if component.Namespace != "" {
+		errsAppend(waitForNamespaceRemoval(component.Namespace, 4*time.Minute))
```
booxter: perhaps it would help to explain why you use different timeouts for different resources. is it the namespace cleanup bug that makes you bump the timeout to 4 full minutes here?

test/check/check.go
```diff
@@ -0,0 +1,270 @@
...
+	return framework.Global.Client.Get(context.TODO(), types.NamespacedName{Name: name}, &rbacv1.ClusterRole{})
+}
+
+func checkForClusterRoleBinding(name string) error {
+	return framework.Global.Client.Get(context.TODO(), types.NamespacedName{Name: name}, &rbacv1.ClusterRoleBinding{})
+}
+
+func checkForSecurityContextConstraints(name string) error {
+	err := framework.Global.Client.Get(context.TODO(), types.NamespacedName{Name: name}, &securityapi.SecurityContextConstraints{})
+	if strings.Contains(err.Error(), "no kind is registered for the type") {
```
booxter: I am confused a bit here: why do we ignore this error and consider it a success? in this check, shouldn't we wait for SCCs to be created?

test/e2e/deployment_test.go
```diff
@@ -0,0 +1,148 @@
...
+	// Deploy given Config Spec
+	Assert(t, CreateConfig(configSpec))
+
+	// If there are any components to deploy wait until Progressing condition is reported
+	if len(components) > 0 {
+		Assert(t, CheckConfigCondition(ConditionProgressing, ConditionTrue, time.Minute))
+	}
+
+	// Wait until Available condition is reported
+	Assert(t, CheckConfigCondition(ConditionAvailable, ConditionTrue, 20*time.Minute))
```
booxter: wow, 20 minutes? can we maybe narrow it down a bit? do we really need all 20 full minutes?

phoracek: It happens so we first run tests pulling individual components, so we load all images on hosts in advance. But in case someone runs only the DeployAllComponents test, it can take a while. However, not 20 minutes, I will set 10, just to be on the safe side while running on my parents' WiFi.

test/check/check.go
phoracek: If SCC is not a type, it means we are running on Kubernetes and there is no point in waiting for it, since it was never deployed. We can change it so we check for SCC availability during tests startup and then check for isSCCAvailable here (like we do in the operator itself). Your call

test/check/check.go
phoracek: Commented on it

cluster/functest.sh
```diff
@@ -1,57 +1,11 @@
...
-  imagePullPolicy: Always
-  kubeMacPool: {}
-  linuxBridge: {}
-  multus: {}
-  sriov: {}
-  nmstate: {}
-EOF
-assert_condition Progressing 60s
-assert_condition Ready 600s
+operator-sdk test \
```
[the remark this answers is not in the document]
phoracek: I don't want to mess with the makefile now. Afraid of rebasing for #131 . This PR might take a while, so if we get #131 merged, I can put this into the Makefile.

automation/check-patch.sh
```diff
@@ -21,6 +21,10 @@ main() {
     curl -sL https://raw.githubusercontent.com/travis-ci/gimme/master/gimme | HOME=/gimme bash >> /etc/profile.d/gimme.sh
     source /etc/profile.d/gimme.sh
 
+    echo "Install operator-sdk"
+    curl -JL https://github.com/operator-framework/operator-sdk/releases/download/v0.8.0/operator-sdk-v0.8.0-x86_64-linux-gnu -o /usr/bin/operator-sdk
```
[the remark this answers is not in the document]
phoracek: Will use vendored operator-sdk once we rebase on #131

test/check/check.go
phoracek: Actually, for removal it should not be that bad.

test/check/check.go
phoracek: I don't understand your comment. Could you give some examples?

test/check/check.go
booxter: As long as you explain it a comment (and maybe consolidate the substring check in a helper function since you do it twice in this patch), it's ok.

test/e2e/deployment_test.go
booxter: maybe worth to mention in a comment here then your thinking

phoracek: Done

test/check/check.go
phoracek: Done

test/check/check.go
```diff
@@ -0,0 +1,270 @@
+package check
```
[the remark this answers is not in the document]
phoracek: Done, please review

test/check/check.go
phoracek: Done

phoracek:
ci test please

phoracek:
@booxter @qinqon @SchSeba I can't figure it out. After Config condition Ready turns True, component check is not able to find all expected components in the system. I tried to add some debugging to the status_manager.go code, and it appears that Condition handling is correct, it turns only when all components are ready. So first I thought, maybe framework Kubernetes client uses cache, I replaced it with plain client-go, the problem was still there. Any idea what could be wrong? To me it looks that test code sees older version of the cluster, is it even possible?

phoracek:
Found 3 bugs in status reporting... Will fix in a separate PR.

phoracek:
@qinqon @SchSeba @booxter ok, removed WIP, please take a look

phoracek:
/hold

There are 2 PRs we need to merge first

Makefile
```diff
@@ -87,8 +87,16 @@ cluster-down:
 cluster-sync:
 	VERSION=$(VERSION) ./cluster/sync.sh
 
+# To run profiling as well:
+#--go-test-flags '-v -timeout 2h -run TestDeployMultus -bench=. -benchmem -blockprofile block.out -cpuprofile profile.out'
 cluster-functest:
-	./cluster/functest.sh
+	go run ./vendor/github.com/operator-framework/operator-sdk/cmd/operator-sdk test \
```
qinqon: parameterize this with:
```make
OPERATOR_SDK := go run ./vendor/github.com/operator-framework/operator-sdk/cmd/operator-sdk
```

cluster/sync.sh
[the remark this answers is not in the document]
qinqon: Don't we want to ?

Also about versioning why do we have all versions here instead of branching ? if we do so we can follow operator-sdk way for resources.

test/check/check.go
qinqon: Na forget about this, was a total brain fart from my part.

cluster/sync.sh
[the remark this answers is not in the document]
phoracek: What you mean by branching?

Not sure if we want to, we can POC getting back to operator-sdk compatibility in future, but there are some issues. e.g. HCO needs our manifests to be vendorable as Go code (other projects may have solved it thought, it's a topic for discussion).

PR #122, opened 2019-06-01
update status even with empty configuration

phoracek:
There is a bug, where after deployment of an empty NetworkAddonsConfig
Spec, we never set condition to running (that is usually done by Pod
reconciliation process after we reconcile the first monitored pod).
When however there are no pods being monitored, we never update the
status.

This patch triggers initial status update right after NodeNetworkConfig
reconciliation is done.


phoracek:
@SchSeba @booxter @qinqon please take a look

PR #123, opened 2019-06-01
reset list of tracked pods after CR removal

phoracek:
There is a bug where after NetworkAddonsConfig removal, we never
remove unprovisioned component pods from our tracked list. Due to
that, after recreation of NodeAddonsConfig, we fail because of missing
pods.

With this patch we make sure that during NetworkAddonsConfig removal,
we reset the list.


phoracek:
@qinqon @booxter @SchSeba please take a look

phoracek:
ci test please

PR #124, opened 2019-06-01
retry Update of condition if needed

phoracek:
With the current code, it can happen that condition Update fails
due to a collision with someone else editing NetworkAddonsConfig
object.

With this patch, we retry the Update several times until it's
successfully updated or out of tries.


phoracek:
@booxter @qinqon @SchSeba please take a look

PR #144, opened 2019-06-11
Sync uses KUBEVIRT_PROVIDER not CLUSTER_PROVIDER

qinqon:
Fix upstream ci to test stdci at opensthift

### DIFF
--- a/automation/test.sh
+++ b/automation/test.sh
@@ -2,7 +2,7 @@
 
 kubectl() { cluster/kubectl.sh "$@"; }
 
-export CLUSTER_PROVIDER=$TARGET
+export KUBEVIRT_PROVIDER=$TARGET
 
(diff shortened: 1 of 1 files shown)


qinqon:
@phoracek and this ?

phoracek:
@qinqon we should fix the rest to use CLUSTER_PROVIDER as well, this has nothing to do with KubeVirt (if something, then kubevirtci, but I'm still in favor of cluster)

phoracek:
But what is broken exactly?

qinqon:
@phoracek I saw something wrong at k8s-nmstate so I translate it to addons, but we can discard this.

phoracek:
We use CLUSTER_PROVIDER everywhere in this project. I don't think this is a bug, but feel free to reopen.

PR #147, opened 2019-06-13
install OLM on local cluster

phoracek:
OLM will be used in later patches to test our CSVs and upgrades.


Makefile
```diff
@@ -86,8 +86,11 @@ cluster-up:
 cluster-down:
 	./cluster/down.sh
 
-cluster-sync:
```
SchSeba: Can we please leave the `cluster-sync` command ? and let him use the `cluster-install-operator` ?

cluster/install-olm.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash -e
+
+echo 'Install OLM on cluster'
+./cluster/kubectl.sh apply -f https://github.com/operator-framework/operator-lifecycle-manager/releases/download/0.10.0/crds.yaml
```
SchSeba: can we generate this files with number 0.0.0 to always use the one in develop?

phoracek:

Makefile
phoracek: why? i will add cluster-install-registry, that will not be compatible with cluster-install-operator, so i'd rather keep them symmetrical. any specific reason to keep it? if there are multiple ways to install the operator, who says which one should have the privilege to be under cluster-sync?

cluster/install-olm.sh
phoracek: this is not generated from our project, those are released OLM manifests. i don't think we should test their development version. maybe once we coverage for one stable, we can extend the matrix

SchSeba: sorry I miss that is the operator and not our CR

Makefile
phoracek: I'm fine with cluster-sync-*

cluster/install-olm.sh
```diff
@@ -0,0 +1,27 @@
...
+./cluster/kubectl.sh apply -f https://github.com/operator-framework/operator-lifecycle-manager/releases/download/0.10.0/olm.yaml
```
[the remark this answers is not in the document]
phoracek: Ack

Makefile
SchSeba: I think we should have the `cluster-install-operator` running when we call the `cluster-sync` to be able to continue fast development.

And for the why to leave the `cluster-sync` I think most of our project have the same convention of `make cluster-up && make cluster-sync && make functest && make cluster-down`
So I will like to save this.

It's easy when you work on multiple projects under the kubevirt org to use the same commands to create and run the code.

phoracek:
@qinqon @SchSeba please take another look

automation/test.sh
```diff
@@ -9,5 +9,5 @@ trap '{ make cluster-down; }' EXIT SIGINT SIGTERM SIGSTOP
 
 make cluster-down
 make cluster-up
-make cluster-sync
+make cluster-install-operator
```
SchSeba: you need to add this to the make file

related to:
```
+ make cluster-install-operator
make: *** No rule to make target `cluster-install-operator'.  Stop.
```

phoracek:
No need for this now. I first want to run lifecycle test suite without OLM.

PR #154, opened 2019-06-18
drop additional manifests from bundle

phoracek:
All manifests included in OLM bundle are automatically installed.
Therefore, we have to remove all extra manifests (e.g. namespace,
RBAC, deployment) from our bundle while building it.


phoracek:
With this change, 0.10.0 (future) bundle works OK.

PR #156, opened 2019-06-19
report version in Status

phoracek:
Let's report operator version and version of deployed components set
in Status portion on NetworkAddonsConfig.

### DIFF
--- a/pkg/controller/statusmanager/status_manager.go
+++ b/pkg/controller/statusmanager/status_manager.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"log"
+	"os"
 	"reflect"
 	"strings"
 	"time"
@@ -22,6 +23,12 @@ const (
 	conditionsUpdateRetries  = 10
 	conditionsUpdateCoolDown = 50 * time.Millisecond
 )
+
+var operatorVersion string
+
+func init() {
+	operatorVersion = os.Getenv("OPERATOR_VERSION")
+}
 
 // StatusLevel is used to sort priority of reported failure conditions. When operator is failing
 // on two levels (e.g. handling configuration and deploying pods), only the higher level will be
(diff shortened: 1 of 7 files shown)


test/check/check.go
```diff
@@ -70,6 +70,49 @@ func CheckConfigCondition(conditionType ConditionType, conditionStatus Condition
 	}
 }
 
+func CheckConfigVersions(operatorVersion, observedVersion, targetVersion string, timeout, duration time.Duration) {
+	By(fmt.Sprintf("Checking that status contains expected versions Operator: %q, Observed: %q, Target: %q", operatorVersion, observedVersion, targetVersion))
+	config := &opv1alpha1.NetworkAddonsConfig{}
+
+	getAndCheckVersions := func() error {
+		err := framework.Global.Client.Get(context.Background(), types.NamespacedName{Name: names.OPERATOR_CONFIG}, config)
+		if err != nil {
```
qinqon: Gomega assert ?

pkg/controller/statusmanager/status_manager.go
```diff
@@ -85,6 +92,13 @@ func (status *StatusManager) set(conditions ...opv1alpha1.NetworkAddonsCondition
 		updateCondition(&config.Status, condition)
 	}
 
+	config.Status.OperatorVersion = operatorVersion
+	config.Status.TargetVersion = operatorVersion
```
qinqon: I don't really get this, they are both the same, also ObserverVersion is going to go from "" to operatorVersion so maybe you just need to show a flag saying "version reached" or the like instead of this tuple of three values that are the same.

test/check/check.go
phoracek: It is used as callback for Eventually and Consistently, therefore we gotta return something to assert later

pkg/controller/statusmanager/status_manager.go
phoracek: I got you. Operator and Target versions are always the same, but it is only implementation detail. In KubeVirt for instance you can have operator in version 2 handling operand in version 1. We don't have this functionality, right after someone updates operator, operand upgrade is triggered. However, this is implicit and could change. Does it make sense?

Now for Target and Observed, those will differ during upgrade. Upgrade tests will be introduced in following PRs.

phoracek: (Also, this API follows KubeVirt)

qinqon: Ahh upgrades it is, now I understand.

PR #157, opened 2019-06-19
reference operator on deployed objects

phoracek:
With this patch, we extend all deployed objects with an annotation
referencing version of the operator that deployed them. This should
make debugging easier. For instance, it should allow us to see whether
there are any leftovers from older operator versions.


pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
```diff
@@ -287,6 +293,16 @@ func (r *ReconcileNetworkAddonsConfig) renderObjects(networkAddonsConfig *opv1al
 	}
 	objs = append([]*unstructured.Unstructured{applied}, objs...)
 
+	// Label objects with version of the operator they were created by
+	for _, obj := range objs {
+		labels := obj.GetLabels()
+		if labels == nil {
+			labels = map[string]string{}
+		}
+		labels["networkaddonsoperator.network.kubevirt.io/version"] = operatorVersion
```
qinqon: You can take the prefix of the string from here https://github.com/kubevirt/cluster-network-addons-operator/blob/bdc8d36b05e55c084fb19fd0ed96e8359b960ab3/pkg/apis/networkaddonsoperator/v1alpha1/register.go#L13

phoracek: Done

booxter:
Can we test it too?

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
```diff
@@ -287,6 +293,16 @@ func (r *ReconcileNetworkAddonsConfig) renderObjects(networkAddonsConfig *opv1al
 	}
 	objs = append([]*unstructured.Unstructured{applied}, objs...)
 
+	// Label objects with version of the operator they were created by
+	for _, obj := range objs {
+		labels := obj.GetLabels()
+		if labels == nil {
+			labels = map[string]string{}
+		}
+		labels[opv1alpha1.SchemeGroupVersion.Group] = operatorVersion
```
qinqon: opv1alpha1.SchemeGroupVersion.Group + "/version"

phoracek:
@qinqon fixed, please take a look

PR #160, opened 2019-06-19
lifecycle test suite

phoracek:
This PR introduces lifecycle test suite. Unlike workflow test suite, which covers tests of operands, lifecycle will focus on operator itself. In this PR, we will test installation and reinstallation of the operator. Following PRs will add test coverage of upgrades.

In order to make testing easier, `Release` object has been introduced. This object keeps version of the release, manifests for operator installation, list of container images used by that release (so we can validate that running containers have expected versions) and supported spec. Supported spec keeps subset of components. If we put something into supported spec, we are committed to support its upgrade (that is not the case for SR-IOV and nmstate now).

When releasing a new version, `release.sh` will commit master release version as released and use it to create a new master. Master component can be edited by developers while developing given version.

Lifecycle test suite will be ran in a separate CI lane.

`cluster-sync` has been split into `cluster-operator-install` and `cluster-operator-push`, since lifecycle requires only the latter.


phoracek:
/cc @booxter @qinqon @fedepaol @SchSeba 

There are two TODOs, but most stuff is already in place. The PR is little bigger, so I tried to explain it in PR description. Will be happy to get any feedback.

Makefile
```diff
@@ -92,6 +92,14 @@ cluster-down:
 cluster-sync:
 	VERSION=$(VERSION) ./cluster/sync.sh
 
+cluster-sync: cluster-operator-push cluster-operator-install
+
+cluster-operator-push:
+	VERSION=$(VERSION) ./cluster/operator-push.sh
```
qinqon: You can replace VERSION=$(VERSION) for 
export VERSION := $(VERSION) in the top of the Makefile

Makefile
```diff
@@ -101,6 +109,15 @@ test/e2e/workflow:
 		--kubeconfig ./cluster/.kubeconfig \
 		--go-test-flags "$(E2E_TEST_ARGS)"
 
+test/e2e/lifecycle:
+	$(OPERATOR_SDK) test \
+		local \
+		./test/e2e/lifecycle \
```
qinqon: $@ even to same in th other, we can even merge the too into a more general test/e2e/*.

automation/test.sh
```diff
@@ -2,12 +2,18 @@
...
 trap '{ make cluster-down; }' EXIT SIGINT SIGTERM SIGSTOP
 
 make cluster-down
 make cluster-up
-make cluster-sync
-make E2E_TEST_EXTRA_ARGS="-ginkgo.noColor" test/e2e/workflow
+if [ "${TEST_SUITE}" == "workflow" ]; then
+    make cluster-operator-push
+    make cluster-operator-install
+    make E2E_TEST_EXTRA_ARGS="-ginkgo.noColor" test/e2e/workflow
```
qinqon: put E2E_TEST_EXTRA_ARGS in a common env var and use it.

hack/release.sh
```diff
@@ -27,6 +27,13 @@ sed -i "s/\(.*startingCSV.*\)${previous_version}\(.*\)/\1${released_version}\2/g
 echo 'Add updated readme to the source tree'
 git add README.md
 
+echo 'Generating new release for workflow e2e tests'
+cp tests/releases/${released_version}.go tests/releases/${future_version}.go
+sed -i "s/Version: \"${released_version}\",/Version: \"${future_version}\",/" tests/releases/${future_version}.go
+
+echo 'Committing new master release module'
+git add tests/releases/${future_version}.go
```
qinqon: This is really necessary ? playing around with git is dangerous.

test/check/check.go
```diff
@@ -113,6 +114,17 @@ func CheckConfigVersions(operatorVersion, observedVersion, targetVersion string,
...
 }
 
+func CheckOperatorIsReady(timeout time.Duration) {
+	By("Checking that the operator is up and running")
+	if timeout != CheckImmediately {
+		Eventually(func() error {
+			return checkForDeployment(components.Name, components.Namespace)
+		}, timeout, time.Second).ShouldNot(HaveOccurred(), fmt.Sprintf("Timed out waiting for the operator to become ready"))
+	} else {
+		Expect(checkForDeployment(components.Name, components.Namespace)).ShouldNot(HaveOccurred(), "Operator is not ready")
```
qinqon: we can do a timeout=1 and interval=1 so it tries only once and wait for 1 so the only thing we do is play with timeout and interval but reuse the assert.

test/e2e/lifecycle/deployment_test.go
```diff
@@ -0,0 +1,65 @@
...
+				CheckConfigCondition(ConditionAvailable, ConditionTrue, 10*time.Minute, CheckDoNotRepeat)
+				CheckReleaseUsesExpectedContainerImages(masterRelease)
+			})
+
+			It("stays in Available condition while the operator is being removed and redeployed", func() {
+				configIsAvailable := func() {
+					CheckConfigCondition(ConditionAvailable, ConditionTrue, CheckImmediately, CheckDoNotRepeat)
+				}
+
+				reinstallingOperator := func() {
```
qinqon: Isn't this different test ?

hack/release.sh
phoracek: I feel doing all these steps manually is more dangerous and most of all annoying. It will end up as a PR, so there is still review process involved and it worked until now. Let's hope it won't delete master branch.

test/e2e/lifecycle/deployment_test.go
phoracek: What do you mean? This is a test reinstalling operator and making sure it doesn't affect operand.

test/check/check.go
phoracek: It is consistent with the other checks. Opened an issue to address this in another PR https://github.com/kubevirt/cluster-network-addons-operator/issues/167

Makefile
phoracek: It is consistent with other targets, opened an issue to address it in another PR (no need to make this one even bigger) https://github.com/kubevirt/cluster-network-addons-operator/issues/168

Makefile
phoracek: Done

automation/test.sh
phoracek: Done

test/e2e/lifecycle/deployment_test.go
qinqon: Smell like context needed "when operator is re-installed" but maybe this is too ginkgo idiomatic stuff.

Makefile
```diff
@@ -92,10 +95,18 @@ cluster-down:
 cluster-sync:
 	VERSION=$(VERSION) ./cluster/sync.sh
 
-test/e2e/workflow:
+cluster-sync: cluster-operator-push cluster-operator-install
+
```
fedepaol: Duplicate `cluster-sync:` entry? Is this something wanted & if so does the double entry improves readability?

phoracek: It was requested earlier in another PR to keep `cluster-sync` since it is well known in kubevirt projects. This target does what would a new user who saw other projects expect from sync IMO

fedepaol: Yes, but there are two entries now:
```
cluster-sync:
	VERSION=$(VERSION) ./cluster/sync.sh

cluster-sync: cluster-operator-push cluster-operator-install
```

Is this something we want?

phoracek: ahhh, no, i'm just blind. thanks

Issue #141, opened 2019-06-08
KubeMacPool does not stay ready


phoracek:
When enabling KubeMacPool component. Status reports Progressing condition, then it successfully turns Ready. However, after some time (sometimes sooner than in a minute), it triggers operator to turn into Progressing/Failing again.

`tests/e2e/deployment_test.go` from https://github.com/kubevirt/cluster-network-addons-operator/pull/118 should include skipped tests that are reproducing the issue.

This issue might be a cause of https://bugzilla.redhat.com/show_bug.cgi?id=1712851, since it caused other components that were deployed at the same time fail to create their resources.

phoracek:
@SchSeba do we still have the initial restart issue with kubemacpool deployment? It is restaring only once after the start, right?

phoracek:
This has been resolved.

PR #162, opened 2019-06-20
include KubeMacPool in functional tests

phoracek:
KubeMacPool is known to restart shortly after first started, skip
this initial restart.

https://github.com/kubevirt/cluster-network-addons-operator/issues/141


phoracek:
@SchSeba please take another look. It's not perfect. But I wan't to have a simple solution (and 100% coverage) and fix it on KubeMacPool side as soon as I find time to do that.

PR #164, opened 2019-06-20
add keep checking while test helper

phoracek:
Nobody wants to read goroutines while going through tests.

### DIFF
--- a/test/check/check.go
+++ b/test/check/check.go
@@ -113,6 +113,30 @@ func CheckConfigVersions(operatorVersion, observedVersion, targetVersion string,
 	}
 }
 
+func KeepCheckingWhile(check func(), while func()) {
+	ticker := time.NewTicker(time.Second)
+	defer ticker.Stop()
+	done := make(chan bool)
+
+	go func() {
+		// Perform some long running operation
+		while()
+
+		// Finally close the validator
+		close(done)
+	}()
+
+	// Keep checking while the goroutine is running
+	for {
+		select {
+		case <-done:
+			return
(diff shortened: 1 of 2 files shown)


tests/e2e/workflow/version_test.go
```diff
@@ -41,42 +41,27 @@ var _ = Describe("NetworkAddonsConfig", func() {
...
-				Multus:      &opv1alpha1.Multus{},
-				LinuxBridge: &opv1alpha1.LinuxBridge{},
+			versionRemainsTheSame := func() {
+				CheckConfigVersions(operatorVersion, operatorVersion, operatorVersion, CheckImmediately, CheckDoNotRepeat)
 			}
 
-			// Run goroutine that will run update of the config on background,
-			// this is needed so we can continously verify reported versions throughout
-			// the setup.
-			go func() {
```
qinqon: We can remove this goroutine all together we just need a goroutine with the validator and two messages send in the two timers.

tests/check/check.go
```diff
@@ -113,6 +113,29 @@ func CheckConfigVersions(operatorVersion, observedVersion, targetVersion string,
...
+
+	// Keep checking while the goroutine is running
+	for {
+		select {
+		case <-done:
+			return
+		default:
+			check()
+		}
+		time.Sleep(time.Second)
```
fedepaol: Instead of this you could take a channel out of a time.Ticker (see https://golang.org/pkg/time/#NewTicker) and put it in the select (instead of sleeping), feels a bit more idiomatic

PR #169, opened 2019-06-24
fix generating of release modules in release.sh

phoracek:


phoracek:
Change does not affect CI.

PR #172, opened 2019-06-24
fix Release sort

phoracek:
Let's not call gomega functions outside tests scope.


phoracek:
Passed lifecycle tests, workflow are not touched by this change.

PR #176, opened 2019-06-25
document bridge configuration

phoracek:


README.md
```diff
@@ -65,6 +65,18 @@ The bridge marker image used to deliver a bridge marker detecting the availabili
...
 variable in operator deployment manifest.
 
+### Configure bridge on node
+
+Following snippets can be used to configure linux bridge on your node.
+
+```shell
+# create the bridge using NetworkManager
+nmcli con add type bridge ifname br10
+
```
qinqon: ports ? just add a place holder like "<port-name>" or the like.

phoracek: this doc should show bare minimum, i don't care how users will connect their bridges, L2 or L3

PR #182, opened 2019-07-13
Add the option to change sriov type

irosenzw:
Since sriov type is changed to 'cnv-sriov' in d/s,
adding an option to change it when gererating a CNA deployment manifest.

This importent for the the automatic build process
of the HCO

Signed-off-by: Ido Rosenzwig <irosenzw@redhat.com>


irosenzw:
@SchSeba please review

phoracek:
@irosenzw @SchSeba SR-IOV will be dropped from network addons operator. It is advised to use https://github.com/openshift/sriov-network-operator instead. Therefore I don't think this PR is needed. Would you agree?

PR #202, opened 2019-08-06
Bump nmstate to 0.6.0

qinqon:
It includes upgrade of nmstate to 0.0.8 and activation of
vlan_filtering for linux-bridges

Signed-off-by: Quique Llorente <ellorent@redhat.com>

### DIFF
--- a/data/nmstate/001-rbac.yaml
+++ b/data/nmstate/001-rbac.yaml
@@ -106,6 +106,7 @@ metadata:
   name: nmstate
 allowPrivilegedContainer: true
 allowHostDirVolumePlugin: true
+allowHostNetwork: true
 runAsUser:
   type: RunAsAny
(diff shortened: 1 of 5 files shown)


data/nmstate/001-rbac.yaml
```diff
@@ -106,10 +106,11 @@ metadata:
...
 allowHostDirVolumePlugin: true
+allowHostNetwork: true
 runAsUser:
   type: RunAsAny
 seLinuxContext:
   type: RunAsAny
 users:
 - system:serviceaccount:nmstate:nmstate-handler
-{{ end }}
+  {{ end }}
```
phoracek: should not be indented

phoracek:
It fails on CI, but I think it is because of there is a bug in the new conditions code...

PR #206, opened 2019-08-07
Release 0.12.0

qinqon:


test/releases/0.13.0.go
```diff
@@ -0,0 +1,50 @@
+package releases
```
SchSeba: I think you need to change this file I will like to see here and also in the 0.12.0 file the ovs and nmstate installation

phoracek: You are right, since we want to use this release in HCO with ovs and nmstate enabled by default, we should test it for future upgrades.

qinqon: Ok don't, can you guys verify ?

phoracek:
https://github.com/kubevirt/cluster-network-addons-operator/pull/206

PR #207, opened 2019-08-07
Bump ovs version to v0.6.0

SchSeba:
This PR bump the ovs version.

The new ovs version was builed with a static GCC and the go ssl lib.

---
This change is [Reviewable](https://reviewable.io/reviews/kubevirt/cluster-network-addons-operator/207)


phoracek:
Error on CI is not related to this PR. Our condition reporting is sometimes flaky. We should address it soon, but I don't see it as a blocker for initial release.

PR #215, opened 2019-08-09
Fix registry Dockerfile to work with origin-registry-operator 4.2.0

qinqon:
The 'latest' or '4.2.0' fails generating bundle.db at:
time="2019-08-09T07:01:21Z" level=fatal msg="could not decode contents of file
/registry/bundles.db into package:
error converting YAML to JSON: yaml: control characters are not allowed"

The error is already present at code base but was not fatal.

Issue was that initilizer is re-feeding itself with bundles.db that is not a JSON/YAML
this review move manifests to proper directory out of initializer

Signed-off-by: Quique Llorente <ellorent@redhat.com>

### DIFF
--- a/build/registry/Dockerfile
+++ b/build/registry/Dockerfile
@@ -1,13 +1,13 @@
-FROM quay.io/openshift/origin-operator-registry:4.1.0
+FROM quay.io/openshift/origin-operator-registry:4.2.0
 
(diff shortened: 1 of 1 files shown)


phoracek:
Passed the only affected tests - Travis

PR #216, opened 2019-08-12
don't install ovs-vsctl binary

phoracek:
This mechanism only causes issues. Incompatibility with system it
is installed on, not being on PATH or read-only filesystems. Let's
drop it and replace by proper ovs.db socket based solution in
a following patch.


phoracek:
The failure is known and unrelated.

PR #217, opened 2019-08-12
Release 0.12.0

phoracek:
* ovs component added
* added upgrade support for ovs and nmstate
* updated bridge-marker to 0.2
* updated bridge-cni to v0.8.1
* removed sr-iov (obsoleted by openshift/sriov-network-operator)
* updated nmstate to v0.8.0
* expose unified HCO conditions

### DIFF
--- a/manifests/cluster-network-addons/0.12.0/cluster-network-addons-operator.0.12.0.clusterserviceversion.yaml
+++ b/manifests/cluster-network-addons/0.12.0/cluster-network-addons-operator.0.12.0.clusterserviceversion.yaml
@@ -1,0 +1,201 @@
+apiVersion: operators.coreos.com/v1alpha1
+kind: ClusterServiceVersion
+metadata:
+  name: cluster-network-addons-operator.0.12.0
+  namespace: placeholder
+  annotations:
+    capabilities: "Full Lifecycle"
+    categories: "Networking"
+    alm-examples: |
+      [
+        {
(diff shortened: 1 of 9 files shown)


phoracek:
Lifecycle failed only 0.12.0 -> master upgrades which are expected to fail (chicken egg problem).

Worflow failed with something weird, maybe caused by a timeout. Will try to rerun on my local environment.

manifests/cluster-network-addons/0.12.0/cluster-network-addons-operator.0.12.0.clusterserviceversion.yaml
```diff
@@ -0,0 +1,201 @@
+apiVersion: operators.coreos.com/v1alpha1
+kind: ClusterServiceVersion
+metadata:
+  name: cluster-network-addons-operator.0.12.0
+  namespace: placeholder
```
mmirecki: What is the "placeholder" namespace?
I don't see it being replaced in any script. Where do we replace it?
It's been here since always.

phoracek: OLM replaces it by the namespace provided by administrator installing given component.

phoracek:
Passed workflow tests on both Kubernetes and OpenShift on my env. I think we need to tweak timeouts for CI.

PR #228, opened 2019-08-30
Release 0.14.0

phoracek:
* add operator version to CSV generator
* reduce kubemacpool resource requirements
* drop Failing condition from pre-0.12.0 versions


phoracek:
Only tests with upgrade from 0.14.0 failed, due to chicken-egg problem

PR #230, opened 2019-08-30
make condition checking interval shorter

phoracek:
Sometimes Progressing turns False so quick, we miss it in our
1 second long interval. With this patch we make it shorter.


test/check/check.go
```diff
@@ -66,13 +66,13 @@ func CheckConfigCondition(conditionType ConditionType, conditionStatus Condition
 	}
 
 	if timeout != CheckImmediately {
-		Eventually(getAndCheckCondition, timeout, time.Second).ShouldNot(HaveOccurred(), fmt.Sprintf("Timed out waiting for the condition, current config:\n%v", configToYaml(config)))
+		Eventually(getAndCheckCondition, timeout, 10*time.Millisecond).ShouldNot(HaveOccurred(), fmt.Sprintf("Timed out waiting for the condition, current config:\n%v", configToYaml(config)))
```
fedepaol: can this be a constant?

test/check/check.go
```diff
@@ -66,13 +66,13 @@ func CheckConfigCondition(conditionType ConditionType, conditionStatus Condition
...
 	} else {
 		Expect(getAndCheckCondition()).NotTo(HaveOccurred(), fmt.Sprintf("Condition is not in the expected state, current config:\n%v", configToYaml(config)))
 	}
 
 	if duration != CheckDoNotRepeat {
-		Consistently(getAndCheckCondition, duration, time.Second).ShouldNot(HaveOccurred(), fmt.Sprintf("Condition prematurely changed its value, current config:\n%v", configToYaml(config)))
+		Consistently(getAndCheckCondition, duration, 10*time.Millisecond).ShouldNot(HaveOccurred(), fmt.Sprintf("Condition prematurely changed its value, current config:\n%v", configToYaml(config)))
```
fedepaol: same as above

phoracek:
/hold

Not sure if the problem was solved by newer version of Kubernetes or this PR. Let's check other pending PRs first.

test/check/check.go
phoracek: It is used only twice. I personally prefer having these kind of values inline, directly on the place where they are needed. Otherwise we would pile up all time constants at the top of the file, I find that more confusing and harder to read, with no notable benefit.

phoracek:
ci test please

Let's make sure it was not only a coincidence

PR #231, opened 2019-09-01
Fix apply function

SchSeba:
This PR fix an issue with the PDB creation.

---
This change is [Reviewable](https://reviewable.io/reviews/kubevirt/cluster-network-addons-operator/231)


SchSeba:
@alonSadan Can you please cherry pick this change on your PR to bump the kubemacpool?

SchSeba:
the PDB is namespace object so we don't need this PR my bad.

PR #233, opened 2019-09-05
keep all operands in the same namespace

phoracek:
Keeping operands in different namespaces is not supportable on
OpenShift with OLM. Let's keep all we deploy in a single namespace.
By default, the namespace of the operator itself will be used.


pkg/components/components.go
```diff
@@ -146,6 +146,14 @@ func GetDeployment(version string, operatorVersion string, namespace string, rep
 										},
 									},
 								},
+								{
+									Name: "OPERAND_NAMESPACE",
```
SchSeba: why we need this?

can't we just use the `OPERATOR_NAMESPACE`

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
```diff
@@ -310,11 +312,13 @@ func (r *ReconcileNetworkAddonsConfig) renderObjects(networkAddonsConfig *opv1al
 // are removed when NetworkAddonsConfig config is
 func (r *ReconcileNetworkAddonsConfig) applyObjects(networkAddonsConfig *opv1alpha1.NetworkAddonsConfig, objs []*unstructured.Unstructured) error {
 	for _, obj := range objs {
-		// Mark the object to be GC'd if the owner is deleted
-		if err := controllerutil.SetControllerReference(networkAddonsConfig, obj, r.scheme); err != nil {
-			log.Printf("could not set reference for (%s) %s/%s: %v", obj.GroupVersionKind(), obj.GetNamespace(), obj.GetName(), err)
-			err = errors.Wrapf(err, "could not set reference for (%s) %s/%s", obj.GroupVersionKind(), obj.GetNamespace(), obj.GetName())
-			return err
+		// Mark the object to be GC'd if the owner is deleted. Don't set owner reference on namespaces if they are used by the operator itself
```
SchSeba: nice comment!

pkg/components/components.go
phoracek: We could, but what if we would need to use two different namespaces then? If we keep it split like this, all our components will work with the assumption those namespaces are not always identical. Maybe I'm too paranoid though.

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
phoracek: :cat:

PR #236, opened 2019-09-10
Upgrade kubernetes-nmstate-handler container to 0.8.0

qinqon:
- Support untagged vlan filtering linux bridge traffic
- Increase nmstate set timeout to 60 seconds


phoracek:
Looks like we still have a race on k8s workflow test. Unrelated though.

PR #257, opened 2019-11-08
use kubevirtci's cluster-up script

phoracek:
Replace the custom local cluster solution with kubevirtci. This should
be easier to read and maintain.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,8 +7,7 @@ build/operator/_test
 build/_output
 
 # Temporary local cluster files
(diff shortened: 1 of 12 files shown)


phoracek:
@fedepaol @qinqon could you please take a look? It should enable us to run tests against OKD providers which was not possible with the original code. It is also consist with the OVS CNi cluster code.

phoracek:
OpenShift lane is unstable and will be replaced by OKD in follow-up PR.

PR #258, opened 2019-11-08
run tests against OKD and simplify e2e automation scripts

phoracek:
Should go in after #257 

Get rid of smart PROVIDER parsing and simply keep single shell script
per target.

Since we switched to kubevirtci's cluster-up, switch from OpenShift 3
to OKD 4 as part of this patch.


phoracek:
@qinqon @oshoval  could you please take a look?

phoracek:
I have to disable Multus deployment checks on OKD (it is there by default)

PR #270, opened 2019-12-05
Handling special cases of outdated/incompatible objects during reconcile

RamLavi:
When a change is introduced to an object's parameters during upgrade, it is sometimes not possible to perform the regular reconcile process to the newer version (such as changing a mutable name).
For this reason we need to add a Special delete functionality which should be added specifically per case when such as change occurs.
Testing this functionality is included as part of the lifecycle tests.

### DIFF
--- a/data/multus/002-multus.yaml
+++ b/data/multus/002-multus.yaml
@@ -24,7 +24,7 @@ spec:
 apiVersion: apps/v1
 kind: DaemonSet
 metadata:
-  name: kube-multus-ds-amd64
+  name: multus
   namespace: {{ .Namespace }}
   labels:
     tier: node
--- a/pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
+++ b/pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
@@ -286,6 +286,13 @@ func (r *ReconcileNetworkAddonsConfig) renderObjects(networkAddonsConfig *opv1al
 		return objs, err
 	}
 
+	// Perform any special object changes that are impossible to do with regular Apply. e.g. Remove outdated objects
+	// and objects that cannot be modified by Apply method due to incompatible changes.
+	if err := network.SpecialCleanUp(&networkAddonsConfig.Spec, r.client); err != nil {
+		log.Printf("failed to Clean Up outdated objects: %v", err)
+		return objs, err
+	}
+
 	// The first object we create should be the record of our applied configuration
 	applied, err := appliedConfiguration(networkAddonsConfig, r.namespace)
 	if err != nil {
--- a/pkg/network/multus.go
+++ b/pkg/network/multus.go
@@ -1,13 +1,20 @@
 package network
 
 import (
+	"context"
+	"fmt"
+	"log"
 	"os"
 	"path/filepath"
 	"reflect"
 
 	osv1 "github.com/openshift/api/operator/v1"
 	"github.com/pkg/errors"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
+	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
 
 	opv1alpha1 "github.com/kubevirt/cluster-network-addons-operator/pkg/apis/networkaddonsoperator/v1alpha1"
 	"github.com/kubevirt/cluster-network-addons-operator/pkg/network/cni"
@@ -34,6 +41,50 @@ func changeSafeMultus(prev, next *opv1alpha1.NetworkAddonsConfigSpec) []error {
 		return []error{errors.Errorf("cannot modify Multus configuration once it is deployed")}
 	}
 	return nil
+}
+
+// cleanUpMultus checks specific multus outdated objects or ones that are no longer compatible and deletes them.
+func cleanUpMultus(conf *opv1alpha1.NetworkAddonsConfigSpec, ctx context.Context, client k8sclient.Client) []error {
+	if conf.Multus == nil {
+		return nil
+	}
+
+	errs := []error{}
+	errs = append(errs, cleanUpMultusOldName(ctx, client)...)
+
+	return errs
+
+}
+
+// cleanUpMultusOldName deletes multus ds object with old name after a new name was introduces in version 0.25.0.
+// REQUIRED_FOR upgrade from multus <= 0.25.0.
+func cleanUpMultusOldName(ctx context.Context, client k8sclient.Client) []error {
+	// Get existing
+	existing := &unstructured.Unstructured{}
+	gvk := schema.GroupVersionKind{Group: "apps", Version: "v1", Kind: "DaemonSet"}
+	existing.SetGroupVersionKind(gvk)
+	namespace := "cluster-network-addons"
+	name := "kube-multus-ds-amd64"
+
+	err := client.Get(ctx, types.NamespacedName{Name: name, Namespace: namespace}, existing)
+	// if we found the object
+	if err == nil {
+		objDesc := fmt.Sprintf("(%s) %s/%s", gvk.String(), namespace, name)
+		log.Printf("Cleaning up %s Object", objDesc)
+
+		// Delete the object
+		err = client.Delete(ctx, existing)
+		if err != nil {
+			log.Printf("Failed Cleaning up %s Object", objDesc)
+			return []error{err}
+		}
+
+	} else if apierrors.IsNotFound(err) {
+		// object not found, no need for action.
+		return nil
+	}
+
+	return []error{err}
 }
 
 // RenderMultus generates the manifests of Multus
--- a/pkg/network/network.go
+++ b/pkg/network/network.go
@@ -1,6 +1,7 @@
 package network
 
 import (
+	"context"
 	"log"
 	"reflect"
 	"strings"
@@ -8,6 +9,7 @@ import (
 	osv1 "github.com/openshift/api/operator/v1"
 	"github.com/pkg/errors"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
 
 	opv1alpha1 "github.com/kubevirt/cluster-network-addons-operator/pkg/apis/networkaddonsoperator/v1alpha1"
 )
@@ -45,6 +47,20 @@ func FillDefaults(conf, previous *opv1alpha1.NetworkAddonsConfigSpec) error {
 
 	if len(errs) > 0 {
 		return errors.Errorf("invalid configuration:\n%s", errorListToMultiLineString(errs))
+	}
+
+	return nil
+}
+
+// specialCleanUp checks if there are any specific outdated objects or ones that are no longer compatible and deletes them.
+func SpecialCleanUp(conf *opv1alpha1.NetworkAddonsConfigSpec, client k8sclient.Client) error {
+	errs := []error{}
+	ctx := context.TODO()
+
+	errs = append(errs, cleanUpMultus(conf, ctx, client)...)
+
+	if len(errs) > 0 {
+		return errors.Errorf("invalid configuration:\n%v", errorListToMultiLineString(errs))
 	}
 
 	return nil
--- a/test/check/components.go
+++ b/test/check/components.go
@@ -31,7 +31,7 @@ var (
 		ClusterRole:                "multus",
 		ClusterRoleBinding:         "multus",
 		SecurityContextConstraints: "multus",
-		DaemonSets:                 []string{"kube-multus-ds-amd64"},
+		DaemonSets:                 []string{"multus"},
 	}
 	NMStateComponent = Component{
 		ComponentName:              "NMState",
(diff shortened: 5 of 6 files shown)


pkg/network/image-pull-policy.go
```diff
@@ -40,6 +44,11 @@ func changeSafeImagePullPolicy(prev, next *opv1alpha1.NetworkAddonsConfigSpec) [
 	return []error{}
 }
 
+// Currently not implemented, since there are no obsolete objects under this module
+func CleanUpImagePullPolicy(ctx context.Context, client k8sclient.Client, objs []*unstructured.Unstructured) []error {
```
phoracek: No need to add this to every module. Add it only to those implementing it and then collect them in network.go

RamLavi: what do you mean by this?

phoracek: Look at the other functions (such as Validate), only modules implementing it have it. Others are simply not executed from network.go

pkg/network/network.go
```diff
@@ -50,6 +52,24 @@ func FillDefaults(conf, previous *opv1alpha1.NetworkAddonsConfigSpec) error {
...
+func SpecialCleanUp(networkAddonsConfig *opv1alpha1.NetworkAddonsConfig, client k8sclient.Client, objs []*unstructured.Unstructured) error {
+
+	errs := []error{}
+	ctx := context.TODO()
+
+	errs = append(errs, CleanUpMultus(ctx, client, objs)...)
+	errs = append(errs, CleanUpLinuxBridge(ctx, client, objs)...)
+
+	if len(errs) > 0 {
+		return errors.Errorf("invalid configuration:\n%v", errs)
```
phoracek: Please use `errorListToMultiLineString` as in other functions.

hackqOB
```diff
@@ -0,0 +1,4453 @@
+[33mcommit 3fdf3f73e7808a1401e5db66d08ba59c6d5b785c[m[33m ([m[1;36mHEAD -> [m[1;32mcnv-3393[m[33m)[m
```
phoracek: Please drop this fine. The same for the .swp one. I recommend you creating a global gitignore with all you project files.

https://help.github.com/en/github/using-git/ignoring-files#create-a-global-gitignore

pkg/network/multus.go
```diff
@@ -36,6 +48,35 @@ func changeSafeMultus(prev, next *opv1alpha1.NetworkAddonsConfigSpec) []error {
 	return nil
 }
 
+// Currently not implemented, since there are no obsolete objects under this module
+func CleanUpMultus(ctx context.Context, client k8sclient.Client, objs []*unstructured.Unstructured) []error {
+	// Get existing
```
phoracek: Please add subfunction per each case of additional cleanup. This one should be for the renaming. This function should have proper TODO to drop it once we don't support the old version anymore.

pkg/network/linux-bridge.go
```diff
@@ -20,6 +29,32 @@ func changeSafeLinuxBridge(prev, next *opv1alpha1.NetworkAddonsConfigSpec) []err
 	return nil
 }
 
+// In older versions of the operator, we used daemon sets of type 'extensions/v1beta1', later we
+// changed that to 'apps/v1'. Because of this change, we are not able to seamlessly upgrade using
+// only Update methods. Following this we find and delete the old daemonSet if configured
+func CleanUpLinuxBridge(ctx context.Context, client k8sclient.Client, objs []*unstructured.Unstructured) []error {
```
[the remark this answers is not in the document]
RamLavi: It is needed but I think I named it wrong. It should be CleanUpBridgeMarker().
Please confirm.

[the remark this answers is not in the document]
phoracek: That workaround was there to support upgrade from very old versions we do not support now. Therefore I believe we can drop it. If we move it here, I would require a test coverage for that (and that may be really difficult since we changed the namespaces since that version and you would need to change the whole lifecycle tests).

Let's drop the workaround we have there, if something bad happens, we can fix it. But again, that version is not supported for a long time (upgrades were not supported at all at the time).

dankenigsberg:
This PR misses a proper title and description. Can they be added?

Can you include a description on where the cleanup code is tested?

pkg/network/multus.go
```diff
@@ -36,6 +44,47 @@ func changeSafeMultus(prev, next *opv1alpha1.NetworkAddonsConfigSpec) []error {
...
+	existing := &unstructured.Unstructured{}
+	gvk := schema.GroupVersionKind{Group: "apps", Version: "v1", Kind: "DaemonSet"}
+	existing.SetGroupVersionKind(gvk)
+	namespace := "cluster-network-addons"
+	name := "kube-multus-ds-amd64"
+
+	err := client.Get(ctx, types.NamespacedName{Name: name, Namespace: namespace}, existing)
+	if err == nil || !(apierrors.IsNotFound(err) ||
+		strings.Contains(err.Error(), "no matches for kind")) {
```
phoracek: We want to proceed only if `err == nil`. If we failed to `Get` with an unexpected error, we want to return an error and fail the reconciliation. If not found, we can just return, all is good.

pkg/network/multus.go
```diff
@@ -36,6 +44,47 @@ func changeSafeMultus(prev, next *opv1alpha1.NetworkAddonsConfigSpec) []error {
...
+
+		objDesc := fmt.Sprintf("(%s) %s/%s", gvk.String(), namespace, name)
+		log.Printf("Cleaning up %s Object", objDesc)
+
+		//Delete the object
```
phoracek: ditto

pkg/network/multus.go
```diff
@@ -36,6 +44,47 @@ func changeSafeMultus(prev, next *opv1alpha1.NetworkAddonsConfigSpec) []error {
...
+		err = client.Delete(ctx, existing)
+		if err != nil {
+			log.Printf("failed Cleaning up %s Object", objDesc)
```
phoracek: Capital F please

pkg/network/network.go
```diff
@@ -50,6 +52,23 @@ func FillDefaults(conf, previous *opv1alpha1.NetworkAddonsConfigSpec) error {
 	return nil
 }
 
+// specialCleanUp checks if there are any outdated objects and deletes them.
+// This means that the object wil be installed as new and not by the upgrade kubectl algorithm
+// The cleanUp is performed on specific obsoleted objects, so some of the cleanUp functions may be not implemented.
+func SpecialCleanUp(client k8sclient.Client) error {
+
```
phoracek: please drop the line

data/multus/002-multus.yaml
```diff
@@ -24,7 +24,7 @@ spec:
 apiVersion: apps/v1
 kind: DaemonSet
 metadata:
-  name: kube-multus-ds-amd64
+  name: kube-multus-ds
```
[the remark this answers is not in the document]
RamLavi: also changed in 0.24.0.go

pkg/network/network.go
```diff
@@ -50,6 +52,23 @@ func FillDefaults(conf, previous *opv1alpha1.NetworkAddonsConfigSpec) error {
 	return nil
 }
 
+// specialCleanUp checks if there are any outdated objects and deletes them.
+// This means that the object wil be installed as new and not by the upgrade kubectl algorithm
+// The cleanUp is performed on specific obsoleted objects, so some of the cleanUp functions may be not implemented.
+func SpecialCleanUp(client k8sclient.Client) error {
```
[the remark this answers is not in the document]
phoracek: We are not deleting outdated objects though (in all cases at least). We use this function to overcome incompatible changes AND delete outdated stuff. e.g. In this case we perform renaming by removing and recreating.

You are right we should reflect that in the comment.

pkg/network/multus.go
```diff
@@ -36,6 +44,47 @@ func changeSafeMultus(prev, next *opv1alpha1.NetworkAddonsConfigSpec) []error {
...
+
+	//deletes the existing multus object by specific name
+	errs = append(errs, cleanUpMultusOldName(ctx, client)...)
+
+	return errs
+
+}
+
+// cleanUpMultusOldName deletes the Multus ds according to a specific old name and namespace.
+// TODO This function should be dropped once we don't support the version with this specific name & namespace.
```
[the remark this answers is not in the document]
phoracek: TODO is even more valuable for far future. When no one of us is around anymore, future maintainers will `git grep TODO` and find a message in a bottle asking them to drop this stuff they don't understand anymore.

pkg/network/multus.go
```diff
@@ -36,6 +44,47 @@ func changeSafeMultus(prev, next *opv1alpha1.NetworkAddonsConfigSpec) []error {
 	return nil
 }
 
+// cleanUpMultus handles special cleanup of obsoleted multus object
+func cleanUpMultus(ctx context.Context, client k8sclient.Client) []error {
+	errs := []error{}
+
+	//deletes the existing multus object by specific name
+	errs = append(errs, cleanUpMultusOldName(ctx, client)...)
```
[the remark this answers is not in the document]
phoracek: CNAO is something close to a framework. When there is a new manifest or a new functionality, network team members should not need to change existing functions really, they just need to attach new functionality to the framework. This place is similar to https://github.com/kubevirt/cluster-network-addons-operator/blob/master/pkg/network/network.go#L25 where we collect/register plugins that want to perform validation.

The goal here is to make the operator extendable.

pkg/network/network.go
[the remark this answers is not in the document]
dankenigsberg: I hate comments, they always rot.

pkg/network/multus.go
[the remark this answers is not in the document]
dankenigsberg: Why would we need to ever drop this function? I *would* say something like

> This function is required as long as we support upgrade from version <= 0.25.0.

In Vdsm we had `REQUIRED_FOR` string, to ease usage by future `git greppers`. My point that this is not a "TODO" which we are too lazy to implement now.

pkg/network/multus.go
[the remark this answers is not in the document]
dankenigsberg: Maybe the wrapper can be named/commented to reflect this.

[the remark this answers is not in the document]
dankenigsberg: BTW, sorry for having posted my comments only today. They have been Pending for a couple of days.

pkg/network/network.go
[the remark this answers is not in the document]
phoracek: It is not comment of how is this function implemented but on what should go in it. It is a guidance for users on how to use this "hook point". If there will be a contribution breaking the comment/purpose, it would lead to -1.

pkg/network/multus.go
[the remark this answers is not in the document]
phoracek: Gotcha. I like your suggestion.

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
```diff
@@ -278,6 +278,12 @@ func (r *ReconcileNetworkAddonsConfig) renderObjects(networkAddonsConfig *opv1al
 		}
 	}
 
+	// Clean Up any outdated obsoleted objects
```
phoracek: Please change the comment:

```
// Perform any special object changes that are impossible to do with regular Apply. e.g. Remove outdated objects and objects that cannot be modified by Apply method due to incompatible changes. 
```

Move this after the `Render` execution (in case Render fails, we should not do any of this).

pkg/network/multus.go
```diff
@@ -36,6 +43,51 @@ func changeSafeMultus(prev, next *opv1alpha1.NetworkAddonsConfigSpec) []error {
...
 }
 
+// cleanUpMultus checks specific multus outdated objects or ones that are no longer compatible and deletes them.
+func cleanUpMultus(conf *opv1alpha1.NetworkAddonsConfigSpec, ctx context.Context, client k8sclient.Client) []error {
+	if conf.Multus == nil {
+		return nil
+	}
+
+	errs := []error{}
+	// deletes the existing multus object by specific name
```
phoracek: You can drop this comment. The one on the function is enough

pkg/network/multus.go
```diff
@@ -36,6 +43,51 @@ func changeSafeMultus(prev, next *opv1alpha1.NetworkAddonsConfigSpec) []error {
...
+	errs = append(errs, cleanUpMultusOldName(ctx, client)...)
+
+	return errs
+
+}
+
+// cleanUpMultusOldName deletes multus ds object with old name after a new name was introduces in version 0.25.0.
+// TODO This function is required as long as we support upgrade from version <= 0.25.0.
```
phoracek: Please drop the TODO and keep only the rest based on Dan's comments

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
RamLavi: I like this comment rephrase.

pkg/network/multus.go
RamLavi: changing according to dan's comment to REQUIRED-FOR ( example in https://github.com/oVirt/vdsm/blob/f6a1ba0a0e232a28db6c38a3cdecf1a5cc96105a/lib/vdsm/network/kernelconfig.py#L232)

phoracek:
/override pull-e2e-cluster-network-addons-operator-lifecycle-okd
/override pull-e2e-cluster-network-addons-operator-lifecycle-workflow-okd

PR #277, opened 2019-12-16
Keep version of used components and their version at a single yaml file

RamLavi:
In order to make U/S->D/S automation possible and clearer which versions are we using, we will keep them in a dedicated yaml. This file should be machine-readable, and must be updated every time a supported version of a component is changed.

### DIFF
--- a/components.yaml
+++ b/components.yaml
@@ -1,0 +1,19 @@
+componenets:
+  multus:
+    version: "bf61002b4c83fcc0ca29c271108ca6afa381998f"
+    upstreamUrl: "https://github.com/intel/multus-cni"
+  linux-bridge:
+    version: "v0.8.1"
+    upstreamUrl: "https://github.com/containernetworking/plugins"
+  bridge-marker:
+    version: "0.2.0"
+    upstreamUrl: "https://github.com/kubevirt/bridge-marker"
+  kube-macpool:
+    version: "v0.8.0"
+    upstreamUrl: "https://github.com/k8snetworkplumbingwg/kubemacpool"
+  nmstate:
+    version: "v0.12.0"
+    upstreamUrl: "https://github.com/nmstate/kubernetes-nmstate"
+  ovs-cni:
+    version: "v0.9.0"
+    upstreamUrl: "https://github.com/kubevirt/ovs-cni"


pkg/components/componenets_config.yaml
```diff
@@ -0,0 +1,25 @@
+componenets:
```
phoracek: Let's keep it in the root of the project as components.yaml

pkg/components/componenets_config.yaml
```diff
@@ -0,0 +1,25 @@
...
+  multus:
+    componentName: "multus"
+    Version: "0.2.0"
```
phoracek: Use lower case "v".

Also, some of the tags start with "v", not the number alone. Please check.

pkg/components/componenets_config.yaml
```diff
@@ -0,0 +1,25 @@
+componenets:
+  multus:
+    componentName: "multus"
```
phoracek: If we keep components under key with their name, let's drop redundant "componentName".

components.yaml
```diff
@@ -0,0 +1,19 @@
+componenets:
+  multus:
+    version: "v3.4"
```
phoracek: In this case, we don't use 3.4 directly, we use an untagged version that came after it.

You can see it in the image tag here: https://github.com/kubevirt/cluster-network-addons-operator/blob/master/pkg/components/components.go#L22

In this case, the version should be therefore set to `bf61002b4c83fcc0ca29c271108ca6afa381998f`

RamLavi: is 
```version: "bf61002b4c83fcc0ca29c271108ca6afa381998f"```
 be good then? i.e., do you think future scripts will be able to interpret it in the same manner they do a specific version?

phoracek: We may need to add "branch" or "revision" later, but I think this is good enough for now. Let's adjust it once we have a real user and real issue.

I'm little nervous that we introduce this file and has some redundancy, but let's use it for reference and docs now, then we start basing automation (both D/S and U/S around it). We can change it whenever, it's no API, just our helper.

phoracek:
/approve
/lgtm
/release-note-none
/override pull-e2e-cluster-network-addons-operator-lifecycle-workflow-okd
/override pull-e2e-cluster-network-addons-operator-lifecycle-workflow-k8s
/override pull-e2e-cluster-network-addons-operator-lifecycle-okd

PR #278, opened 2020-01-10
status_manager: only require single ready pod for deployment

phoracek:
Currently we wait until all the pods in deployment become ready.
That is not necessary, we can turn Ready once at least one pod
in the deployment is able to serve requests.

This is needed, so we don't block when deployment uses leader
election in combination with readiness probe and only one pod is
marked as ready at the time.

### DIFF
--- a/pkg/controller/statusmanager/status_manager.go
+++ b/pkg/controller/statusmanager/status_manager.go
@@ -300,9 +300,7 @@ func (status *StatusManager) SetFromPods() {
 
 		// Finally check whether Pods belonging to this Deployments are being started or they
 		// are being scheduled.
-		if dep.Status.UnavailableReplicas > 0 {
-			progressing = append(progressing, fmt.Sprintf("Deployment %q is not available (awaiting %d nodes)", depName.String(), dep.Status.UnavailableReplicas))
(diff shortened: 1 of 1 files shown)


phoracek:
cc @AlonaKaplan @dankenigsberg

phoracek:
@RamLavi please take a look.

pkg/controller/statusmanager/status_manager.go
```diff
@@ -300,9 +300,7 @@ func (status *StatusManager) SetFromPods() {
 
 		// Finally check whether Pods belonging to this Deployments are being started or they
 		// are being scheduled.
-		if dep.Status.UnavailableReplicas > 0 {
-			progressing = append(progressing, fmt.Sprintf("Deployment %q is not available (awaiting %d nodes)", depName.String(), dep.Status.UnavailableReplicas))
-		} else if dep.Status.AvailableReplicas == 0 {
+		if dep.Status.AvailableReplicas == 0 {
```
qinqon: This depends on the nature of the app you are installing, some have leader elections other don't, maybe we have to make that configurable or check it.

PR #282, opened 2020-01-19
add script for kubemacpool bump

phoracek:
This patch adds a script that will be used to fetch new manifests and container images based on the components version listed in `components.yaml`. With this PR, we add such automation for CNAO.


phoracek:

Seems unrelated. Will tackle the upgrade issues in a follow-up PR.

PR #287, opened 2020-02-02
Support a component deprovision

oshoval:

Support a component deprovision
    
Allow to deprovision a component from the cluster,
by editing networkaddonsconfig.

```release-note
NONE
```


oshoval:
will remove DBGDBG when done debugging if possible please

pull-e2e-cluster-network-addons-operator-workflow-k8s isnt stable also on master right ?

pkg/apply/apply.go
```diff
@@ -74,10 +74,50 @@ func ApplyObject(ctx context.Context, client k8sclient.Client, obj *uns.Unstruct
 			}
 
 			return errors.Wrapf(err, "could not update object %s", objDesc)
-		} else {
-			log.Print("update was successful")
 		}
 	}
 
+	log.Print("update was successful")
```
phoracek: If the condition from 56 if false, there is no need to update and we don't need to log. Please keep it on the original place.

pkg/network/nmstate.go
```diff
@@ -14,14 +15,18 @@ import (
 
 func changeSafeNMState(prev, next *opv1alpha1.NetworkAddonsConfigSpec) []error {
 	if prev.NMState != nil && !reflect.DeepEqual(prev.NMState, next.NMState) {
-		return []error{errors.Errorf("cannot modify NMState state handler configuration once it is deployed")}
+		if !(prev.NMState != nil && next.NMState == nil) {
+			return []error{errors.Errorf("cannot modify NMState state handler configuration once it is deployed")}
```
phoracek: Now when we support removal of knmstate, we don't need this check anymore. You can drop the whole function.

pkg/apply/apply.go
```diff
@@ -74,9 +74,49 @@ func ApplyObject(ctx context.Context, client k8sclient.Client, obj *uns.Unstruct
 			}
 
 			return errors.Wrapf(err, "could not update object %s", objDesc)
-		} else {
-			log.Print("update was successful")
 		}
+		log.Print("update was successful")
```
oshoval: you right, hope its fine now logic, since the if block always return
the else can be removed,
if you prefer i will revert to original

pkg/network/network.go
```diff
@@ -93,47 +93,52 @@ func IsChangeSafe(prev, next *opv1alpha1.NetworkAddonsConfigSpec) error {
 	return nil
 }
 
-func Render(conf *opv1alpha1.NetworkAddonsConfigSpec, manifestDir string, openshiftNetworkConfig *osv1.Network, clusterInfo *ClusterInfo) ([]*unstructured.Unstructured, error) {
+func Render(prev, conf *opv1alpha1.NetworkAddonsConfigSpec, manifestDir string, openshiftNetworkConfig *osv1.Network, clusterInfo *ClusterInfo) ([]*unstructured.Unstructured, []*unstructured.Unstructured, error) {
```
phoracek: There is no reason why `Render` should handle anything related to removals. All it should do is to render templates from `data/` based on environment variables and data from `NetworkAddonsConfig`.

Please introduce a new function `ObejctsToRemove` which would calculate the removal.

pkg/network/nmstate.go
oshoval: hmm but arent there updates which arent legit ? (where deepequal might show a real error)
i understand, just making sure please

pkg/apply/apply.go
```diff
@@ -74,9 +74,49 @@ func ApplyObject(ctx context.Context, client k8sclient.Client, obj *uns.Unstruct
...
+		return errors.Wrapf(err, "object %s unsupported", objDesc)
+	}
+
+	// Get existing
+	existing := &uns.Unstructured{}
+	existing.SetGroupVersionKind(gvk)
+	err := client.Get(ctx, types.NamespacedName{Name: obj.GetName(), Namespace: obj.GetNamespace()}, existing)
+
+	log.Printf("DBGDBG at delete %v A1", name)
+	if name != "cluster-network-addons" {
```
oshoval: here there is a nit, 
on the last component in the ns , the ns should be deleted if we want to handle it
or we know that one will always exists

pkg/apply/apply.go
oshoval: you right, please re-review that snippet, wrote there a comment (this one is outdated)

pkg/apply/apply.go
oshoval: as Petr raised, the NS also has the cnao pod so it shouldnt be deleted even if no components left, so this bullet is fine
thanks

pkg/network/network.go
oshoval: done

oshoval:
will add support for rest of the components and update tests according

pkg/apply/apply.go
phoracek: This check should be done on a higher level. This method should be universal (not CNAO specific). It should merely Delete anything given to it.

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
```diff
@@ -195,7 +195,7 @@ func (r *ReconcileNetworkAddonsConfig) Reconcile(request reconcile.Request) (rec
 	}
 
 	// Canonicalize and validate NetworkAddonsConfig, finally render objects of requested components
-	objs, err := r.renderObjects(networkAddonsConfig)
+	objs, objsToRemove, err := r.renderObjects(networkAddonsConfig)
```
phoracek: Again, even here we should call to r.objectsToDelete, render is meant only to interface call to the Render.

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
```diff
@@ -210,6 +210,14 @@ func (r *ReconcileNetworkAddonsConfig) Reconcile(request reconcile.Request) (rec
 		return reconcile.Result{}, err
 	}
 
+	// Delete generated objsToRemove on Kubernetes API server
```
phoracek: Here you can drop the namespace from the list of objects (or in r.objectsToRemove, not sure what fits better). https://github.com/kubevirt/cluster-network-addons-operator/blob/39ca491bc9e99f98b996026adcd3f1fec8a2ffab/pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go#L46

phoracek: Sorry, to clarify, we should add a new var that reads from "OPERAND_NAMESPACE"

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
oshoval: yep this already did , but didnt push yet, sorry for the confusing

oshoval:
i need please green light that those two are flaky without my code
about the 2nd one it fails easilty on master as well
the first one i saw once fail on master i think,
and can present a legit order change of the render folders that cause it to appear (but based on my branch which act as master in aspect of failures before the mentioned change)

[Fail] NetworkAddonsConfig when there is no pre-existing Config [It] should be able to deploy all components one by one 
/home/oshoval/go/src/github.com/kubevirt/cluster-network-addons-operator/test/check/check.go:82

[Fail] NetworkAddonsConfig when all components are already deployed [It] should be able to remove the config and create it again 
/home/oshoval/go/src/github.com/kubevirt/cluster-network-addons-operator/test/check/check.go:82

oshoval:
not ready to re-review yet, some test would fail, and need first to fix current remarks
thx

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
oshoval: but renderObjects calculate much stuff that is needed for both network.Render and network.RenderObjsToRemove
its bad to do them twice i think ? i will push some commits and lets respin thinking ?

oshoval: but renderObjects calculate much stuff that is needed for both network.Render and network.RenderObjsToRemove
its bad to do them twice i think ? i will push some commits and lets respin thinking ?

pkg/apply/apply.go
oshoval: removing it as part of next commit totally, filtering the list of objToRemove so wont need it

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
oshoval: done (on next push)

pkg/apply/apply.go
oshoval: bump

pkg/network/nmstate.go
oshoval: bump

oshoval:
/cc @phoracek 
ready to re-review please, thanks for the intels

pkg/apply/apply.go
oshoval: bump

oshoval:
about the failing tests
lifecycle is known to be flaky, i hope there isnt regression

about workflow
tried @ validation_test.go line 42 to leave there the leftover unsupported removal 
imagePullPolicy, maybe the root problem there is something about the value it has
      imagepullpolicy: ""

because according image-pull-policy.go, this condition wont mark as bad,
do we want to leave this test there ? (i updated it to fit a component that cant be removed, but if we leave it it still adjustment)

```
	if prev.ImagePullPolicy != "" && prev.ImagePullPolicy != next.ImagePullPolicy {
		return []error{errors.Errorf("cannot modify ImagePullPolicy configuration once components were deployed")}
	}
```
• Failure [61.810 seconds]
NetworkAddonsConfig
/tmp/cluster-network-addons-operator/go/path/src/github.com/kubevirt/cluster-network-addons-operator/test/e2e/workflow/validation_test.go:15
  when a valid config is deployed
  /tmp/cluster-network-addons-operator/go/path/src/github.com/kubevirt/cluster-network-addons-operator/test/e2e/workflow/validation_test.go:39
    and a component which does not support removal is removed from the Spec 

the 2nd failing test is 
that fails on timeout, i believe this test is flaky, saying so before if i took a previous state of this PR, which had just one and the exact flaky test that clean master had, and just reordered there the renders it cause this test below to fail
(if needed i have a diff file that show what i did, but it cant be applied since the code was changed since)
 /tmp/cluster-network-addons-operator/go/path/src/github.com/kubevirt/cluster-network-addons-operator/test/e2e/workflow/deployment_test.go:16
  when there is no pre-existing Config
  /tmp/cluster-network-addons-operator/go/path/src/github.com/kubevirt/cluster-network-addons-operator/test/e2e/workflow/deployment_test.go:17
    should be able to deploy all components one by one [It]

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
phoracek: It does a lot of stuff that we absolutely don't need for the removal. Let's start by having a separate function for it and then we can see if we can make a parent function both of those to cover the shared parts.

It would make it cleared for the review to see what information you need to perform the removal. And when do you gather it.

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
```diff
@@ -339,6 +356,20 @@ func (r *ReconcileNetworkAddonsConfig) applyObjects(networkAddonsConfig *opv1alp
 	return nil
 }
 
+// Delete removed objects
+func (r *ReconcileNetworkAddonsConfig) deleteObjects(networkAddonsConfig *opv1alpha1.NetworkAddonsConfig, objs []*unstructured.Unstructured) error {
+	for _, obj := range objs {
+		// Apply all objects on apiserver
```
phoracek: Outdated doc, you don't apply, you delete. I think it is redundant in this tiny function, let's drop it.

pkg/apply/apply.go
```diff
@@ -74,9 +74,44 @@ func ApplyObject(ctx context.Context, client k8sclient.Client, obj *uns.Unstruct
...
+	}
+
+	return nil
+}
+
+// DeleteObject deletes an object in the apiserver
+func DeleteObject(ctx context.Context, client k8sclient.Client, obj *uns.Unstructured) error {
+	name := obj.GetName()
+	namespace := obj.GetNamespace()
+	if name == "" {
```
phoracek: When is this the case? Is it needed? All our objects have a name, don't they?

pkg/apply/apply.go
```diff
@@ -74,9 +74,44 @@ func ApplyObject(ctx context.Context, client k8sclient.Client, obj *uns.Unstruct
...
+		return errors.Errorf("object %s has no name", obj.GroupVersionKind().String())
+	}
+	gvk := obj.GroupVersionKind()
+	// used for logging and errors
+	objDesc := fmt.Sprintf("(%s) %s/%s", gvk.String(), namespace, name)
+	log.Printf("reconciling %s", objDesc)
+
+	if err := IsObjectSupported(obj); err != nil {
```
phoracek: This function is for `Apply`, we should support removal for everything with Delete.

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
```diff
@@ -339,6 +356,20 @@ func (r *ReconcileNetworkAddonsConfig) applyObjects(networkAddonsConfig *opv1alp
...
+		if err := apply.DeleteObject(context.TODO(), r.client, obj); err != nil {
+			log.Printf("could not apply (%s) %s/%s: %v", obj.GroupVersionKind(), obj.GetNamespace(), obj.GetName(), err)
+			err = errors.Wrapf(err, "could not apply (%s) %s/%s", obj.GroupVersionKind(), obj.GetNamespace(), obj.GetName())
```
phoracek: Again, this is not about apply but delete. Also, we should be graceful when deleting, if the object is not found, we don't need to fail. On the other hand, if we fail to remove an existing object, we should raise an error. Please handle it in the DeleteObject funnction.

pkg/network/network.go
```diff
@@ -93,49 +94,131 @@ func IsChangeSafe(prev, next *opv1alpha1.NetworkAddonsConfigSpec) error {
...
 	objs := []*unstructured.Unstructured{}
-
-	// render Multus
-	o, err := renderMultus(conf, manifestDir, openshiftNetworkConfig, clusterInfo)
-	if err != nil {
-		return nil, err
+	o := []*unstructured.Unstructured{}
+	var err error
+
+	if conf.KubeMacPool != nil {
```
phoracek: Why are you changing this function? Please rever the changes if not needed.

oshoval: we need it, it was in the function , but sometimes the function (rendernmstate) can allow nil so moved it from there
(on removal it allow conf nil if prev != nil && prev.nmstate != nil)

reorg is possible but then the render itself will need to get prev as arg as well

pkg/apply/apply.go
oshoval: just copy pasted from ApplyObject tbh, dont know by heart atm

pkg/network/network.go
phoracek: Are you talking about your new render2remove or the original one?

oshoval: the original one, but there the if was at renderNMState itself,
but i removed it from renderNMState , because now i need two different conditions on the callers of renderNMState
one conf != nil ,and the other (upon destroy) the long cond above

pkg/apply/apply.go
phoracek: Then please take the time and understand the code you introduce.

oshoval: i did understand to the extend i felt needed at that point but no problem

if we look on the original ApplyObject which has it
it gets rendered objects, that for example were created from yamls at data folder.
if one of them doesnt have a valid name , it will catch it

now lets say someone does delete, i think its a good sanity check that it will at least have a name, for example if file got corrupted

pkg/network/network.go
oshoval: i have an idea maybe, 
i will put the long condition
conf.module != nil || (prev != nil && prev.module != nil)
as an acceptance gate into each of the moduleRender functions

and then on the RenderDelete i will give conf and prev,
but on Render i will give just conf, so the condition would be as it was before,
depend just on conf

this way we can also toggle out new feature in order to test the failing tests
but if i am not wrong, indeed to reorder there can do troubles
we will get to it

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
oshoval: reread now all the stuff it does before render , i think i need all, but i will split to two func and make the common part as a func and then we can review that common func if i am not wrong that we need all of it

(the suffix is just needed for render itself)

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
oshoval: the logic you said is already there according what i checked now, see DeleteObject, in case i am not wrong
but /s/apply/delete/g/
thanks

pkg/network/network.go
oshoval: i am doing the suggested change above just to the kmp, so you can view and say if its ok

btw after that (or maybe not as a part of this PR), we can take it to a next level of creating an array of objects and iterate on it, each with its own fields, but first let fix all the test and such, this is more advance go practice i guess

oshoval: added a commit that show a suggested change only to kmp , so you see if you like the approach please

pkg/network/network.go
```diff
@@ -101,14 +101,12 @@ func Render(conf *opv1alpha1.NetworkAddonsConfigSpec, manifestDir string, opensh
...
 
-	if conf.KubeMacPool != nil {
-		// render kubeMacPool
-		o, err = renderKubeMacPool(conf, manifestDir)
-		if err != nil {
-			return nil, err
-		}
-		objs = append(objs, o...)
+	// render kubeMacPool
+	o, err = renderKubeMacPool(nil, conf, manifestDir)
```
phoracek: This is not terrible, but it breaks the logic of render. renderMultus should "render" the component, that's all. The decision whether render it or not is up to the higher layer - Render of the whole Spec.

I also don't see why is this needed for this deletion change. If you want to render objects to remove, you can reuse the current renderKubeMacPool method, except you would give it the `prev` config instead of the current one. You can use renderKubeMacPool at its current form. once with `prev` (for ObjsToRemove) once with `conf` (in this function).

oshoval:
https://github.com/kubevirt/cluster-network-addons-operator/pull/292
here i run empty commit to see master state

pkg/network/network.go
```diff
@@ -101,14 +101,12 @@ func Render(conf *opv1alpha1.NetworkAddonsConfigSpec, manifestDir string, opensh
...
 	var err error
 
-	if conf.KubeMacPool != nil {
-		// render kubeMacPool
-		o, err = renderKubeMacPool(conf, manifestDir)
-		if err != nil {
-			return nil, err
-		}
-		objs = append(objs, o...)
+	// render kubeMacPool
```
phoracek: The `if` should stay here and be dropped from inside of the renderKubeMacPool function.

oshoval: the if was there in the first place, and if its there, it saves a boolean expression on both render and renderDestroy
but up to you, if you want i will take it out, wanted to strive to minimal changes + little sanity checks in the leaf user

phoracek: I value the semantics more than 3 saved lines. Please keep it outside.

phoracek: Oh, you say it was there originally. Lemme see.

phoracek: You are right. Sorry, then keep it in the original place. The single commit diff confused me.

pkg/network/network.go
```diff
@@ -93,12 +94,15 @@ func IsChangeSafe(prev, next *opv1alpha1.NetworkAddonsConfigSpec) error {
...
 
+// Render creates component object list
 func Render(conf *opv1alpha1.NetworkAddonsConfigSpec, manifestDir string, openshiftNetworkConfig *osv1.Network, clusterInfo *ClusterInfo) ([]*unstructured.Unstructured, error) {
 	log.Print("starting render phase")
 	objs := []*unstructured.Unstructured{}
+	o := []*unstructured.Unstructured{}
+	var err error
 
 	// render Multus
-	o, err := renderMultus(conf, manifestDir, openshiftNetworkConfig, clusterInfo)
```
oshoval: will revert this

oshoval:
@phoracek 
is the code fine now, so we can move on to fix the tests please?
most of them pass as good as master, reordering the modules fixed a bunch of failing ones

RamLavi:
The PR misses a proper description. Can it be added?

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
```diff
@@ -30,6 +30,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/source"
 
 	opv1alpha1 "github.com/kubevirt/cluster-network-addons-operator/pkg/apis/networkaddonsoperator/v1alpha1"
+	opv1alphav1 "github.com/kubevirt/cluster-network-addons-operator/pkg/apis/networkaddonsoperator/v1alpha1"
```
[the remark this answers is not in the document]
oshoval: added those types as saw on other file, should we get rid of one of them and use just one of them ?

oshoval:
> The PR misses a proper description. Can it be added?

sure , but please open as discussion so we can resolve if possible

pkg/network/ovs_test.go
```diff
@@ -39,10 +39,9 @@ var _ = Describe("Testing Ovs", func() {
 		Context("when there is previous value, but the new one is empty (removing component)", func() {
 			prev := &opv1alpha1.NetworkAddonsConfigSpec{Ovs: &opv1alpha1.Ovs{}}
 			new := &opv1alpha1.NetworkAddonsConfigSpec{}
-			It("should fail", func() {
+			It("should success", func() {
```
phoracek: succeed

pkg/network/ovs_test.go
```diff
@@ -39,10 +39,9 @@ var _ = Describe("Testing Ovs", func() {
...
 				errorList := changeSafeOvs(prev, new)
-				Expect(len(errorList)).To(Equal(1), "validation of safe change failed due to an unexpected error: %v", errorList)
-				Expect(errorList[0].Error()).To(Equal("cannot modify Ovs configuration once it is deployed"))
+				Expect(len(errorList)).To(Equal(0), "validation of safe change failed due to an unexpected error: %v", errorList)
```
phoracek: No need to check for len, just check that the list is empty: https://github.com/kubevirt/cluster-network-addons-operator/pull/287/files#diff-2297106a77e4f1babcc0c6b321b7954dR35

pkg/network/ovs.go
```diff
@@ -15,8 +16,12 @@ import (
 
 func changeSafeOvs(prev, next *opv1alpha1.NetworkAddonsConfigSpec) []error {
```
phoracek: We can drop this function completely. All changes to the ovs value are not valid. There are only two options, it is there or it is not.

pkg/network/nmstate_test.go
```diff
@@ -36,13 +36,12 @@ var _ = Describe("Testing nmstate", func() {
 			})
 		})
 
-		Context("when there is previous value, but the new one is empty (removing component)", func() {
+		Context("when there is previous value, but the new one is empty (removing NMState component)", func() {
 			prev := &opv1alpha1.NetworkAddonsConfigSpec{NMState: &opv1alpha1.NMState{}}
 			new := &opv1alpha1.NetworkAddonsConfigSpec{}
-			It("should fail", func() {
+			It("should success", func() {
```
phoracek: succeed

pkg/network/network.go
```diff
@@ -136,6 +140,75 @@ func Render(conf *opv1alpha1.NetworkAddonsConfigSpec, manifestDir string, opensh
...
+	if conf.Ovs == nil && prev != nil {
+		// render Ovs for Removal
+		o, err = renderOvs(prev, manifestDir, clusterInfo)
+		if err != nil {
+			return nil, err
+		}
+		objsToRemove = append(objsToRemove, o...)
+	}
+
+	// Remove Namespace occurences, since its not allowed to remove it, as cnao pod is there
```
phoracek: Just filter out all Namespace type objects. Also, you need to read OPERAND_NAMESPACE if anything.

pkg/network/multus_test.go
```diff
@@ -137,10 +137,9 @@ var _ = Describe("Testing multus", func() {
 		Context("when there is previous value, but the new one is empty (removing component)", func() {
 			prev := &opv1alpha1.NetworkAddonsConfigSpec{Multus: &opv1alpha1.Multus{}}
 			new := &opv1alpha1.NetworkAddonsConfigSpec{}
-			It("should fail", func() {
+			It("should success", func() {
```
phoracek: succeed

pkg/network/multus.go
```diff
@@ -38,7 +38,10 @@ func validateMultus(conf *opv1alpha1.NetworkAddonsConfigSpec, openshiftNetworkCo
 
 func changeSafeMultus(prev, next *opv1alpha1.NetworkAddonsConfigSpec) []error {
```
phoracek: Not needed, drop it

pkg/network/kubemacpool.go
```diff
@@ -104,6 +108,7 @@ func renderKubeMacPool(conf *opv1alpha1.NetworkAddonsConfigSpec, manifestDir str
 	data.Data["Namespace"] = os.Getenv("OPERAND_NAMESPACE")
 	data.Data["KubeMacPoolImage"] = os.Getenv("KUBEMACPOOL_IMAGE")
 	data.Data["ImagePullPolicy"] = conf.ImagePullPolicy
+
```
phoracek: drop it

pkg/network/linux-bridge.go
```diff
@@ -15,7 +16,10 @@ import (
 
 func changeSafeLinuxBridge(prev, next *opv1alpha1.NetworkAddonsConfigSpec) []error {
```
phoracek: drop it

pkg/network/kubemacpool.go
```diff
@@ -88,7 +89,10 @@ func fillDefaultsKubeMacPool(conf, previous *opv1alpha1.NetworkAddonsConfigSpec)
 
 func changeSafeKubeMacPool(prev, next *opv1alpha1.NetworkAddonsConfigSpec) []error {
 	if prev.KubeMacPool != nil && !reflect.DeepEqual(prev.KubeMacPool, next.KubeMacPool) {
```
phoracek: `if prev.KubeMacPool != nil && next.KubeMacPool != nil &7 !equal`, no need for the second if

pkg/network/network.go
RamLavi: I am not sure why are we checking for cnao namespace. is removing the cnao instance a possible scenario that requires protecting? Please elaborate

test/e2e/workflow/validation_test.go
```diff
@@ -39,7 +39,7 @@ var _ = Describe("NetworkAddonsConfig", func() {
 	Context("when a valid config is deployed", func() {
 		BeforeEach(func() {
 			configSpec := opv1alpha1.NetworkAddonsConfigSpec{
-				LinuxBridge: &opv1alpha1.LinuxBridge{},
+				ImagePullPolicy: v1.PullAlways,
```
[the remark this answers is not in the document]
oshoval: then i will remove the test totally because i can revert to linuxbridge as it now supports removal, ok ?
(btw this is the failing test that doesnt fail on master, maybe just this one is the leftover)

test/e2e/workflow/validation_test.go
```diff
@@ -57,4 +57,29 @@ var _ = Describe("NetworkAddonsConfig", func() {
 			})
 		})
 	})
+
+	Context("when a valid config is deployed", func() {
```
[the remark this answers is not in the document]
oshoval: removing the one of line 42 (please see above comment) , so this will be the leftover one

pkg/network/ovs.go
oshoval: if  remove it , need to remove ovs_test.go totally, because this is all it test there
on it, unless you say otherwise

pkg/network/network.go
```diff
@@ -136,6 +140,75 @@ func Render(conf *opv1alpha1.NetworkAddonsConfigSpec, manifestDir string, opensh
...
+// RenderObjsToRemove creates removal component object list
+func RenderObjsToRemove(prev, conf *opv1alpha1.NetworkAddonsConfigSpec, manifestDir string, openshiftNetworkConfig *osv1.Network, clusterInfo *ClusterInfo) ([]*unstructured.Unstructured, error) {
+	log.Print("starting rendering objects to delete phase")
+	objsToRemove := []*unstructured.Unstructured{}
+	o := []*unstructured.Unstructured{}
+	var err error
+
+	operatorNamespace := os.Getenv("OPERATOR_NAMESPACE")
+
+	if conf.Multus == nil && prev != nil {
```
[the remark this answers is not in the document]
oshoval: conf.Multus == nil  must be checked, so a module wont find itself both in the create list and the remove list if we dont check it.

since renderMultus checks just for its conf (which is prev in our case) for 
conf.Module != nil and not conf != nil (s/conf/prev/), we are doing it here

pkg/network/network.go
oshoval: you mean to change the comment ? 
i do read OPERATOR_NAMESPACE at func begining
or need also OPERAND_NAMESPACE ?

pkg/network/kubemacpool.go
oshoval: what does &7 !equal means ?

you mean just to change to 
if prev.KubeMacPool != nil && next.KubeMacPool != nil 
?
if so i dont understand, we allow the following
prev.KubeMacPool != nil && next.KubeMacPool == nil

guess i miss something

pkg/network/network.go
oshoval: part of the module rendering has the namespace where the cnao pod itself resides
so its forbidden to remove it, else you will end without cnao operator

oshoval:
doing some sanity tests and will upload a commit with the requested changes (beside the one that i didnt understand + title that will add at the end with squash)
thanks

oshoval:
hope i didnt destroy lifecycle with this push

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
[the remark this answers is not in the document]
RamLavi: just one of them is fine.

pkg/network/network.go
RamLavi: I understand the motive, but not sure it is plausible - the objsToRemove list is comprised by very specific modules :ovs, nmstate, kmp, linux-bridge and multus. 
any object in objsToRemove, no matter its namespace, could not be the cnao pod. could it?

oshoval: the object will be a list of k8s entities such as namespace, deamonset , configmap, clusterrole, clusterRolebinding, 

all need to be deleted beside the namespace, the namespace was generated as part of the creation and since the component is now at the operator component , the ns cant be deleted, it was tested and groomed with Petr as well
on the demo i will show the parts that are removed

phoracek: The code is not as clear as it could be but the situation is following:

All components are created in OPERAND_NAMESPACE, operator is running in OPERATOR_NAMESPACE. By default, these variables are the same. We need to make sure we don't ever remove OPERATOR_NAMESPACE, OPERAND_NAMESPACE can be removed only if it is not used anymore. I'm fine with leaving the removal of it and just keep it hanging no matter what. We can handle it in a follow-up.

pkg/network/kubemacpool.go
phoracek: IF kubemacpool was set AND kubemacpool should be still set AND the previous configuration is not the same as the future one; THEN and only then raise and error.

pkg/network/nmstate.go
oshoval: think we removed it, so its old

pkg/network/network.go
oshoval: so any ai or not as i understand ?

oshoval: do we take this in the future on other PR ?
or what is the AI here please?

oshoval:
@phoracek 
@RamLavi 
ready to rereview please 
thx

oshoval:
what is the right method to write down, in case needed, future AI

1. handling the zombie configmap of kmp by adding owner reference.
2. improving ns removal logic to support OPERAND on its last used component getting deleted

pkg/network/network.go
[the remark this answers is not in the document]
phoracek: I see, thanks for the explanation

pkg/network/network.go
```diff
@@ -136,6 +134,70 @@ func Render(conf *opv1alpha1.NetworkAddonsConfigSpec, manifestDir string, opensh
 	return objs, nil
 }
 
+// RenderObjsToRemove creates list of components to be removed
+func RenderObjsToRemove(prev, conf *opv1alpha1.NetworkAddonsConfigSpec, manifestDir string, openshiftNetworkConfig *osv1.Network, clusterInfo *ClusterInfo) ([]*unstructured.Unstructured, error) {
+	log.Print("starting rendering objects to delete phase")
+	objsToRemove := []*unstructured.Unstructured{}
+	o := []*unstructured.Unstructured{}
+	var err error
+
```
phoracek: Could you add an early return here? We can skip the whole function if pref == nil.

pkg/network/network.go
phoracek: There is an action item. For now let's settle with "We will not remove any namespace". So iterate objsToRemove and drop all objects of Namespace type. Add a TODO comment to cleanup OPERAND_NAMESPACE once there are no components using it.

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
```diff
@@ -339,6 +376,19 @@ func (r *ReconcileNetworkAddonsConfig) applyObjects(networkAddonsConfig *opv1alp
 	return nil
 }
 
+// Delete removed objects
+func (r *ReconcileNetworkAddonsConfig) deleteObjects(networkAddonsConfig *opv1alpha1.NetworkAddonsConfig, objs []*unstructured.Unstructured) error {
```
phoracek: You don't use networkAddonsConfig in this function.

pkg/apply/apply.go
oshoval: aging 7 days auto close, xD

pkg/network/kubemacpool.go
```diff
@@ -87,9 +87,10 @@ func fillDefaultsKubeMacPool(conf, previous *opv1alpha1.NetworkAddonsConfigSpec)
 }
 
 func changeSafeKubeMacPool(prev, next *opv1alpha1.NetworkAddonsConfigSpec) []error {
-	if prev.KubeMacPool != nil && !reflect.DeepEqual(prev.KubeMacPool, next.KubeMacPool) {
+	if prev.KubeMacPool != nil && next.KubeMacPool != nil && !reflect.DeepEqual(prev.KubeMacPool, next.KubeMacPool) {
```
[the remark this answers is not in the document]
oshoval: is it fine to just add kmp to the test https://github.com/kubevirt/cluster-network-addons-operator/blob/master/test/e2e/workflow/validation_test.go

configSpec add kmp, and then it will remove it ?
(did same there for nmstate)

pkg/network/network.go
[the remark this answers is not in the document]
oshoval: btw, i would even move the conf != nil into the leaf functions, its a good practice that the function that use a parameter is the one that test it, up to you

pkg/network/network.go
oshoval: please see if i did fine, thanks

pkg/network/network.go
```diff
@@ -136,6 +134,72 @@ func Render(conf *opv1alpha1.NetworkAddonsConfigSpec, manifestDir string, opensh
...
+		o, err := renderOvs(prev, manifestDir, clusterInfo)
+		if err != nil {
+			return nil, err
+		}
+		objsToRemove = append(objsToRemove, o...)
+	}
+
+	// Remove operator namespace occurences, since its not allowed to remove it, because cnao pod is there.
+	// TODO cleanup OPERAND_NAMESPACE once there are no components using it.
+	operatorNamespace := os.Getenv("OPERATOR_NAMESPACE")
```
phoracek: If anything, this should be using OPERAND namespace, not OPERATOR.

pkg/network/network.go
```diff
@@ -136,6 +134,72 @@ func Render(conf *opv1alpha1.NetworkAddonsConfigSpec, manifestDir string, opensh
...
+	index := 0
+	for _, obj := range objsToRemove {
+		name := obj.GetName()
+		if name != operatorNamespace {
```
phoracek: You cannot just check for the Name, that could remove any object with the same name. Please add a check for the object type. If it is Namespace with the operand namespace name, don't remove it.

pkg/network/network.go
```diff
@@ -136,6 +134,72 @@ func Render(conf *opv1alpha1.NetworkAddonsConfigSpec, manifestDir string, opensh
...
+			objsToRemove[index] = obj
+			index++
```
phoracek: This is a weird approach to drop an item from a list. It is hard to read. Would you please just create a new list `objsToRemoveWithoutNamespace` and if the iterated object is not the namespace, add it there.

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
```diff
@@ -194,14 +194,28 @@ func (r *ReconcileNetworkAddonsConfig) Reconcile(request reconcile.Request) (rec
 		return reconcile.Result{}, err
 	}
 
+	openshiftNetworkConfig, prev, err := r.renderPrepare(networkAddonsConfig)
```
phoracek: This is a bad function signature, it should not return openshiftNetworkConfig. If you need it for renderObjects, move out the function that fetches the openshiftNetworkConfig on this level and pass it to renderPrepare. Also, why does renderPrepare returns only `prev`. Please don't create misc functions that do everything and return random stuff. Keep it logical

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
```diff
@@ -210,6 +242,14 @@ func (r *ReconcileNetworkAddonsConfig) Reconcile(request reconcile.Request) (rec
 		return reconcile.Result{}, err
 	}
 
+	// Delete generated objsToRemove on Kubernetes API server
+	err = r.deleteObjects(objsToRemove)
+	if err != nil {
+		// If failed, set NetworkAddonsConfig to failing and requeue
+		r.statusManager.SetFailing(statusmanager.OperatorConfig, "FailedToApply", err.Error())
```
oshoval: need to change on next commit to "deleteObjects" @ SetFailing

phoracek:

phoracek:
/hold

@oshoval cancel the hold when ready

phoracek:

The PR was only rebased and squashed.

PR #288, opened 2020-02-06
replace : in version name with _

phoracek:
When SHA is used for versioning, it contains character ':' which is
not allowed to be used as an annotation value. With this PR, we replace
it with '_'.


phoracek:
The CI failure is not related to  the PR.

phoracek:

PR #291, opened 2020-02-07
rename multus daemon set in variable namespace

phoracek:

### DIFF
--- a/pkg/network/multus.go
+++ b/pkg/network/multus.go
@@ -63,7 +63,7 @@ func cleanUpMultusOldName(ctx context.Context, client k8sclient.Client) []error 
 	existing := &unstructured.Unstructured{}
 	gvk := schema.GroupVersionKind{Group: "apps", Version: "v1", Kind: "DaemonSet"}
 	existing.SetGroupVersionKind(gvk)
-	namespace := "cluster-network-addons"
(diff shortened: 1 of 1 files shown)


phoracek:

phoracek:
@RamLavi please review

pkg/network/multus.go
```diff
@@ -63,7 +63,7 @@ func cleanUpMultusOldName(ctx context.Context, client k8sclient.Client) []error
 	existing := &unstructured.Unstructured{}
 	gvk := schema.GroupVersionKind{Group: "apps", Version: "v1", Kind: "DaemonSet"}
 	existing.SetGroupVersionKind(gvk)
-	namespace := "cluster-network-addons"
+	namespace := os.Getenv("OPERAND_NAMESPACE")
```
RamLavi: Since this is a special cleanup function, shouldn't the parameters be specific rather than generic?

phoracek: Yes, but "cluster-network-addons" is specific only to what happened to be the default on U/S at the time. On D/S we use a different namespace and any user can use any namespace.

The assumption of CNAO is that you run all operands in a single namespace and it is not supported to switch namespace between releases. Therefore we can be sure the same namespace was used in the past.

PR #294, opened 2020-02-14
pin kubemacpool to v0.8.1

phoracek:


phoracek:

Multus deployment is faulty, but seems unrelated.

PR #304, opened 2020-02-19
Fix ocp make cluster-sync

oshoval:

### DIFF
--- a/cluster/operator-push.sh
+++ b/cluster/operator-push.sh
@@ -17,7 +17,7 @@
 set -ex
 
 registry_port=$(./cluster/cli.sh ports registry | tr -d '\r')
-if [[ "${KUBEVIRT_PROVIDER}" =~ ^okd-.*$ ]]; then \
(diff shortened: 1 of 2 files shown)


oshoval:
wokflow failure here is the one that master clean has as well

phoracek:
It fails with `message: Expected DaemonSet "cluster-network-addons/bridge-marker" does not exist`. I have not seen that on master.

oshoval:
> It fails with `message: Expected DaemonSet "cluster-network-addons/bridge-marker" does not exist`. I have not seen that on master.

there is timeout there
    [91mTimed out after 900.219s.
the test is the same one that is failing right ?

i can remove the bump if needed (and use it just for my testing locally), but i dont think this is what caused it, but that the test is unstable, rest of the code in this PR is 2 lines about ocp /okd which arent affecting the tests

phoracek:
The timeout happened due to a different issue. On master it is due to Multus. I'm not aware of any other component failing.

Please drop the bump.

oshoval:
test done, looks like master i think ?

phoracek:

PR #320, opened 2020-02-28
don't fail the test if we fail to list all objects

phoracek:
With the current, kubectl describe sometimes fail to list objects that
are being removed. It is probably between listing of all objects and obtaining
of their descriptions. In any case, we can merely print both the description
and error if this helper function fails.


phoracek:

PR #321, opened 2020-02-28
self-heal by re-reconciling

phoracek:
Kubernetes sometimes fails to apply objects while we remove and recreate
components, despite reporting success. In order to self-heal after these
incidents, keep requeing.


RamLavi:
is there an open issue on this with k8s? if there is, can you attach so we could keep an eye on a possible future fix?

phoracek:
I'm not aware of a specific bug where this would be tracked. Was not able to find specific issue for "while terminating, apply returns success but does not apply all passed objects".

PR #323, opened 2020-02-28
disable knmstate webhook

phoracek:
We dropped the webhook configuration, but forgot to disable the webhook
server. This is causing knmstate pod to fail.


phoracek:
/hold

The webhook manifest is still there

phoracek:
/hold cancel

It is not...

PR #325, opened 2020-03-02
add prepare-release scripts

phoracek:
This should make it easier to release new versions of CNAO. In a follow-up PR
we should automate the next step - releasing. So a maintainer would use this
new script to prepare a release PR and automation performs the release once this
PR is merged.

Signed-off-by: Petr Horacek <phoracek@redhat.com>

PR #327, opened 2020-03-02
DNM Release 0.28.0

phoracek:
/hold

Just to test #325


phoracek:
Failed only on chicken-egg, good!

PR #328, opened 2020-03-03
clarify what is "public branch" in release readme

phoracek:

### DIFF
--- a/README.md
+++ b/README.md
@@ -255,7 +255,7 @@ make cluster-down
 
 # Releasing
 
-1. Checkout a public branch
+1. Checkout a public branch (e.g. `master` or `release-*`)
 2. Call `make prepare-patch|minor|major` and prepare release notes
 3. Open a new PR
 4. Once the PR is merged, create a new release in GitHub and attach new manifests


README.md
```diff
@@ -255,7 +255,7 @@ make cluster-down
 
 # Releasing
 
-1. Checkout a public branch
+1. Checkout a public branch (e.g. `master` or `release-*`)
 2. Call `make prepare-patch|minor|major` and prepare release notes
```
RamLavi: I would add after this line another bullet:
Push to remote repository in order to make it public `git push --set-upstream origin <new-branch-name>`

README.md
```diff
@@ -255,7 +255,7 @@ make cluster-down
...
 3. Open a new PR
```
RamLavi: Might I suggest change the 2 last bullets into one like so:
- Once the PR is merged, Open a new PR 
- Create a new release in GitHub and attach new manifests and push it to the PR.

phoracek: "Once the PR is merged, Open a new PR" what do you mean by that?

RamLavi: there are essentially 2 PRs here (maybe we should call them PR#1 and PR#2 to avoid confusion)
What I meant is, `Once PR#1 is merged, open PR#2` - meaning they shouldn't be opened in parallel*.
then `Create a new release in GitHub and attach new manifests and push it to PR#2.`

*  If this a wrong assumption then we can drop this of course. but at the very least, it needs to be clear what should come into each PR.

phoracek: There is a single PR. The release process looks as following:

- Checkout master
- Call prepare-minor, this creates a new commit with bump of the version and new set of manifests
- Open a PR with this new commit
- If all looks good, merge

Step 4 is not done via a PR. You just open GitHub UI, draft a new Release, update PRs. No review process is present in this step.

README.md
phoracek: `origin` as remote branch is opinionated. For me it is usually the original repo, for others it is their fork. I don't want to tell people how to open PRs, that depends on their git workflow. Anyway, I'll try to make the description more clear.

README.md
RamLavi: much clearer to me now. Thanks!

PR #338, opened 2020-03-10
limit versions tested in lifecycle lane

phoracek:
Currently we always test upgrade from all released versions to the master
branch.

That has two major issues:

The upgrade test of more than 20 versions back is ran on every PR. This takes a
lot of time and slows down the development cycle. With this PR we test upgrades
only from releases pinned to HCO on regular PRs. That should be sufficient since
we test the full upgrade on release PRs.

When releasing, we test upgrade from to-be-released version to master. That is
failing since the to-be-released version is not yet accessible in Quay. With
this PR we make sure to exclude this version from the test.

Lifecycle tests without this change take around 1h20m on a regular PR. With this change, the time was reduced to 30m. Time needed for the workflow lane is around 20m (parallel to the lifecycle one).

We can further optimalize lifecycle upgrade tests in a follow-up PRs.

### DIFF
--- a/test/releases/releases.go
+++ b/test/releases/releases.go
@@ -2,10 +2,12 @@ package releases
 
 import (
 	"fmt"
+	"os"
 	"sort"
 	"strings"
 
 	"github.com/blang/semver"
+	"github.com/gobwas/glob"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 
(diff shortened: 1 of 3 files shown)


phoracek:
cc @qinqon

test/releases/releases.go
```diff
@@ -28,14 +30,45 @@ type Release struct {
...
 
 // Returns list of releases sorted from oldest to newest
 func Releases() []Release {
-	if releasesSorted {
+	if releasesProcessed {
 		return releases
 	}
 
+	// Keep only releases matching the selector
+	if releasesSelectorRaw, found := os.LookupEnv("RELEASES_SELECTOR"); found {
```
qinqon: What happens if we have both selector and deselector?

phoracek: Selectors are applied first and then we use deselectors on that. In other words, they are merged. You cannot SELECT something that was DESELCTED.

PR #339, opened 2020-03-10
bump knmstate to 0.17.0

phoracek:


phoracek:
/hold

We should release 0.17 and use it instead. 0.16 has a race in webhook startup.

phoracek:

Race in tests.

PR #342, opened 2020-03-12
drop limitations of resources

phoracek:
There seems to be a bug in CRI-O causing pods to fail with
"lstat /proc/63538/ns/ipc: no such file or directory" when there is not enough
resources allocated to them.

With this PR we drop all the resource limitations. As long as the cluster is not
low on resources, we should not trigger the bug.

BZ: https://bugzilla.redhat.com/1776236


phoracek:

PR #346, opened 2020-03-12
Integrate macvtap cni

maiqueb:
Add macvtap CNI to the list of network addons installed by the CNAO operator.

It deploys and empty configuration, which currently doesn't expose any host interfaces.

Once https://github.com/kubevirt/macvtap-cni/pull/11 is merged, it would start to expose all the host's interfaces.

```release-note
Add macvtap-cni as a network addon.
```


tools/manifest-templator/manifest-templator.go
```diff
@@ -55,7 +55,7 @@ type templateData struct {
 	OperatorVersion string
 	Namespace       string
 	ContainerPrefix string
-	ImageName	string
+	ImageName	    string
 	ContainerTag    string
```
qinqon: Space

tools/manifest-templator/manifest-templator.go
```diff
@@ -242,7 +243,7 @@ func main() {
 		OperatorVersion: *operatorVersion,
 		Namespace:       *namespace,
 		ContainerPrefix: *containerPrefix,
-		ImageName:	 *imageName,
+		ImageName:	 	 *imageName,
```
qinqon: Ditto

data/macvtap/001-rbac.yaml
```diff
@@ -0,0 +1,50 @@
...
+  name: macvtap-cni-cr
+rules:
+  - apiGroups:
+      - ""
+    resources:
+      - nodes
+      - nodes/status
+    verbs:
+      - get
+      - update
```
phoracek: Is this needed? It was for OVS where we use extended resources (manually set the capacity), but with macvtap, the device plugin framework should do that.

data/macvtap/002-macvtap-daemonset.yaml
```diff
@@ -0,0 +1,63 @@
+---
+kind: ConfigMap
```
phoracek: Please try to edit this CM deployed by CNAO. I have a bad feeling the settings would get reverted by the operator. e.g. deploy macvtap, edit the CM, additionally deploy ovs, see if the CM remained in the original state.

If the CM changes, we have two options:
* Ask user to deploy it themselves
* Add explicit setting to CNAO not to touch it

I feel the former is better.

data/macvtap/002-macvtap-daemonset.yaml
```diff
@@ -0,0 +1,63 @@
...
+        imagePullPolicy: {{ .ImagePullPolicy }}
+        resources:
+          requests:
+            cpu: "60m"
+            memory: "30Mi"
+          limits:
+            cpu: "60m"
+            memory: "30Mi"
+        securityContext:
+          privileged: true
```
phoracek: Is this needed? I forgot the scope of the macvtap DP. Is it just passing the character device or is it creating it as well?

test/e2e/workflow/validation_test.go
```diff
@@ -41,6 +42,7 @@ var _ = Describe("NetworkAddonsConfig", func() {
 			configSpec := opv1alpha1.NetworkAddonsConfigSpec{
 				LinuxBridge: &opv1alpha1.LinuxBridge{},
 				NMState:     &opv1alpha1.NMState{},
+				MacvtapCni:  &opv1alpha1.MacvtapCni{},
```
phoracek: No need to deploy this additionally. Point of the test is to deploy some minimal valid config.

test/check/components.go
```diff
@@ -38,7 +38,7 @@ var (
 		ClusterRoleBinding:         "nmstate-handler",
 		ClusterRole:                "nmstate-handler",
 		SecurityContextConstraints: "nmstate",
-		DaemonSets:                 []string{
+		DaemonSets: []string{
```
phoracek: Needs go formatting

pkg/network/macvtap.go
```diff
@@ -0,0 +1,38 @@
...
+	data.Data["EnableSCC"] = clusterInfo.SCCAvailable
+	data.Data["MacvtapImage"] = os.Getenv("MACVTAP_CNI_IMAGE")
+	if clusterInfo.OpenShift4 {
+		data.Data["CniMountPath"] = cni.BinDirOpenShift4
+	} else {
+		data.Data["CniMountPath"] = cni.BinDir
+	}
+	objs, err := render.RenderDir(filepath.Join(manifestDir, "macvtap"), &data)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to render nmstate state handler manifests")
```
phoracek: Find the copy-paste

pkg/network/macvtap.go
```diff
@@ -0,0 +1,38 @@
+package network
+
+import (
+	"os"
+	"path/filepath"
+
+	"github.com/kubevirt/cluster-network-addons-operator/pkg/network/cni"
```
phoracek: internal imports should be in the last group

data/macvtap/002-macvtap-daemonset.yaml
maiqueb: Yup, after updating the value, the operator puts it back the way it was.

Remember that in https://github.com/kubevirt/macvtap-cni/pull/11 the configmap is still required, and that the macvtap DS won't come up until the config map is available.

That would require us (at least in the tests) to create the config map by some other means (hacky move).

How can we add the explicit setting to CNAO to stay away from the config map ? What are the disadvantages of this option ?

data/macvtap/002-macvtap-daemonset.yaml
maiqueb: It creates it. We need to be privileged.

pkg/network/macvtap.go
maiqueb: Done.

pkg/network/macvtap.go
maiqueb: Done.

test/check/components.go
maiqueb: It's correct, sorry. Look ugly, but this was go fmt that did it; it's consistent w/ https://github.com/kubevirt/cluster-network-addons-operator/pull/346/files#diff-a0979fa1282fb419073a584875cb4b30R50

test/e2e/workflow/validation_test.go
maiqueb: OK.

tools/manifest-templator/manifest-templator.go
maiqueb: go fmt is good w/ it ...

data/macvtap/001-rbac.yaml
maiqueb: I sincerely do not know, but the current macvtap-cni repo manifest does not use **any** of this.

I'm assuming I can get rid of it - leaving only the SCC section.

README.md
```diff
@@ -147,6 +148,31 @@ spec:
...
+metadata:
+  name: cluster
+spec:
+  macvtap: {}
+```
+
+Macvtap-cni must be explicitly configured by the administrator, indicating the
+interfaces on top of which logical networks can be created.
+
+A simple example on how to do so, the user must deploy a `ConfigMap`, such as in [this example](https://github.com/kubevirt/macvtap-cni/blob/master/examples/macvtap-deviceplugin-config.yaml).
```
maiqueb: Should we set this example in https://github.com/kubevirt/cluster-network-addons-operator/pull/346/files#diff-e1a65158257c165bb0c41bf08016e58bR2-R8 ?

phoracek: Currently we don't have any tests of functionality of operands. I really want to add at least basic sanity checks, but we need to setup test lane for it first. No need to add the test.

data/macvtap/002-macvtap-daemonset.yaml
```diff
@@ -0,0 +1,63 @@
...
+        envFrom:
+          - configMapRef:
+              name: macvtap-dp-config
+        image: {{ .MacvtapImage }}
+        imagePullPolicy: {{ .ImagePullPolicy }}
+        resources:
+          requests:
+            cpu: "60m"
+            memory: "30Mi"
+          limits:
```
phoracek: Please drop limits. It causes obscure bugs on OpenShift. `requests` is enough

data/macvtap/002-macvtap-daemonset.yaml
phoracek: Disadvantage is that we spoil the generic code with "if macvtap DP".

I was writing a guide on "how to hack it", but I think there is an elegant way too.

So this issue where we want to keep user changes on objects deployed is pretty common, although we were successful avoiding addressing it, until now.

So let's think how can we make it generic in the CNAO framework and keep everything specific to the component under data/ or pkg/network.

*Option A*: The simplest fit into the current API would be to introduce a Kubernetes object label `networkaddonsoperator.network.kubevirt.io/untouchable: "true"` (or any better name after `/`). That can be set by a user who wants to debug manual changes on any deployed component (e.g. they want to change resource requests, add debugging flags to running apps etc). It can be also set by default under data/ manifests. The only code we need to introduce then would live around https://github.com/kubevirt/cluster-network-addons-operator/blob/master/pkg/apply/apply.go#L37, where we check if an existing object we want to update has this flag, if it does, we ignore it. If you implement it, please do it in a separate PR (in case we need to backport it) and document it at the bottom of the README.

I am tempted to say "let's just keep it as it is and wait for the proper API", but that would make it harder to test under kubevirt and if somebody unaware uses it (despite it us "experimental", they may get hurt).

test/check/components.go
```diff
@@ -51,11 +51,21 @@ var (
 			"ovs-cni-amd64",
 		},
 	}
+	MacvtapComponent = Component{
+		ComponentName:              "Macvtap",
+		ClusterRole:                "",
```
phoracek: You can drop these two empty lines. `""` is the default

PR #351, opened 2020-03-14
tests: try to catch progressing state more reliably

phoracek:
There is a race on our CI where we sometimes don't catch Progressing state. That
can happen when the the image of a deployed component is already in registry and
therefore the start of it is swift. With this change, we run configuration
change with a little delay after we start listening for the progressing state.
This can still fail, but our chances are better. Better approach would be to
start a monitor instead of polling, that is however not yet supported in the
controller-runtime Client.

### DIFF
--- a/test/e2e/workflow/deployment_test.go
+++ b/test/e2e/workflow/deployment_test.go
@@ -181,16 +181,26 @@ var _ = Describe("NetworkAddonsConfig", func() {
 })
 
 func testConfigCreate(configSpec opv1alpha1.NetworkAddonsConfigSpec, components []Component) {
-	CreateConfig(configSpec)
-	checkConfigChange(components)
+	checkConfigChange(components, func() {
+		CreateConfig(configSpec)
+	})
 }
 
 func testConfigUpdate(configSpec opv1alpha1.NetworkAddonsConfigSpec, components []Component) {
-	UpdateConfig(configSpec)
-	checkConfigChange(components)
+	checkConfigChange(components, func() {
+		UpdateConfig(configSpec)
+	})
 }
 
-func checkConfigChange(components []Component) {
+func checkConfigChange(components []Component, while func()) {
+
+	// Start the function with a little delay to give the Progressing check a better chance
+	// of catching the event
+	go func() {
+		time.Sleep(time.Second)
+		while()
+	}()
+
 	// On OpenShift 4, Multus is already deployed by default
 	onlyMultusOnOKDCluster := (len(components) == 1 &&
 		IsOnOKDCluster() &&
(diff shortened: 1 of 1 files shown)


test/e2e/workflow/deployment_test.go
```diff
@@ -181,16 +181,26 @@ var _ = Describe("NetworkAddonsConfig", func() {
...
 func testConfigUpdate(configSpec opv1alpha1.NetworkAddonsConfigSpec, components []Component) {
-	UpdateConfig(configSpec)
-	checkConfigChange(components)
+	checkConfigChange(components, func() {
+		UpdateConfig(configSpec)
+	})
 }
 
-func checkConfigChange(components []Component) {
+func checkConfigChange(components []Component, while func()) {
```
RamLavi: I would add a TODO comment to change this once controler-runtime support monitor functionality

test/e2e/workflow/deployment_test.go
```diff
@@ -181,16 +181,26 @@ var _ = Describe("NetworkAddonsConfig", func() {
 })
 
 func testConfigCreate(configSpec opv1alpha1.NetworkAddonsConfigSpec, components []Component) {
-	CreateConfig(configSpec)
-	checkConfigChange(components)
+	checkConfigChange(components, func() {
+		CreateConfig(configSpec)
```
RamLavi: I'm not sure how [CreateConfig](https://github.com/kubevirt/cluster-network-addons-operator/blob/09a72b225be8c8c671beaed8eb16d9096846f962/test/operations/operations.go#L34) is considered a "while function" (as suggested from the input parameter of checkConfigChange). can you explain?

test/e2e/workflow/deployment_test.go
```diff
@@ -181,16 +181,26 @@ var _ = Describe("NetworkAddonsConfig", func() {
...
+	})
 }
 
 func testConfigUpdate(configSpec opv1alpha1.NetworkAddonsConfigSpec, components []Component) {
-	UpdateConfig(configSpec)
-	checkConfigChange(components)
+	checkConfigChange(components, func() {
+		UpdateConfig(configSpec)
```
RamLavi: ditto

test/e2e/workflow/deployment_test.go
phoracek: Good call, opened an RFE: https://github.com/operator-framework/operator-sdk/issues/2655

test/e2e/workflow/deployment_test.go
phoracek: It is `checking config change` while `creating config`. I will add a docstring describing the functionality, but you are right, it may need a better name, I can't think of any though. Any suggestion?

PR #359, opened 2020-04-07
Bump knmstate to v0.15.5

qinqon:
Bump kubernetes-nmstate to v0.15.5 so it handles the upgrade between CNAO installed kubernetes-nmstate and non CNAO installed kubernetes-nmstate using a lock.

**Release note**:

```release-note
knmstate: Lock handler if another handler is running at the same node, needed for upgrades between cnao installed knmstate and non-cnao installed knmstate.
```


qinqon:
@phoracek is the workflow job related ?

phoracek:
May be.

PR #363, opened 2020-04-08
add bump scripts for linux bridge and set it to 0.8.5

phoracek:
Let's automate releasing of new container images containing containernetworking
plugins linux-bridge and tuning.

### DIFF
--- a/hack/components.sh
+++ b/hack/components.sh
@@ -29,6 +29,8 @@ function get_component_tag() {
     )
 }
 
+PUSH_IMAGES=${PUSH_IMAGES:-}
+
 CNAO_VERSION=${VERSION} # Exported from Makefile
 
 echo 'Setup temporary Go path'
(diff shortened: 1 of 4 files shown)


hack/components.sh
```diff
@@ -17,7 +19,8 @@ KUBEMACPOOL_PATH=${GOPATH}/src/${KUBEMACPOOL_REPO}
 
 echo 'Fetch kubemacpool sources'
 (
-    go get ${KUBEMACPOOL_REPO} || true
+    mkdir -p ${KUBEMACPOOL_PATH}
+    git clone ${KUBEMACPOOL_URL} ${KUBEMACPOOL_PATH} || true
```
maiqueb, line 23: how can this fail ? (other than no internet connectivity ? I mean that possibly the `|| true` is not needed when using w/ git.

sorry for nit-picking, but I think this belongs on a different commit (no need to do it if in the end all commits get squashed into a single one).

hack/components.sh
```diff
@@ -116,3 +120,58 @@ MACVTAP_IMAGE_TAGGED=${MACVTAP_IMAGE}:${MACVTAP_TAG}
...
+
+echo 'Fetch linux-bridge sources'
+(
+    mkdir -p ${LINUX_BRIDGE_PATH}
+    git clone ${LINUX_BRIDGE_URL} ${LINUX_BRIDGE_PATH} || true
+    cd ${LINUX_BRIDGE_PATH}
+    git fetch origin
+    git reset --hard
+    git checkout ${LINUX_BRIDGE_COMMIT}
+)
```
maiqueb, line 138: We should eventually move this into a function - I also didn't do it, so I don't think it's fair to make you jump through that hoop.

hack/components.sh
```diff
@@ -116,3 +120,58 @@ MACVTAP_IMAGE_TAGGED=${MACVTAP_IMAGE}:${MACVTAP_TAG}
...
+
+echo 'Get linux-bridge tag'
+LINUX_BRIDGE_TAG=$(
+    cd ${LINUX_BRIDGE_PATH}
+    git describe --tags
+)
```
maiqueb, line 144: ditto

hack/components.sh
phoracek: You are so right. Thanks for fighting my lazy ass! 

https://github.com/kubevirt/cluster-network-addons-operator/pull/366

hack/components.sh
phoracek: Good call

https://github.com/kubevirt/cluster-network-addons-operator/pull/367

hack/components.sh
```diff
@@ -125,3 +127,48 @@ MACVTAP_IMAGE_TAGGED=${MACVTAP_IMAGE}:${MACVTAP_TAG}
...
+ENV GOPATH=/go
+COPY . /go/src/github.com/containernetworking/plugins/
+WORKDIR /go/src/github.com/containernetworking/plugins/
+RUN ./build_linux.sh
+
+FROM registry.access.redhat.com/ubi8/ubi-minimal
+RUN mkdir -p /usr/src/containernetworking/plugins/bin
+RUN microdnf install -y findutils
+COPY --from=builder /go/src/github.com/containernetworking/plugins/bin/bridge /usr/src/containernetworking/plugins/bin/bridge
+COPY --from=builder /go/src/github.com/containernetworking/plugins/bin/tuning /usr/src/containernetworking/plugins/bin/tuning
```
maiqueb, line 159: shouldn't this 'live' somewhere else ? I understand we're packaging cni's repo code in a docker img, which, imho would merit a file of itself - not sure if in ` cmd/...` .

hack/components.sh
```diff
@@ -125,3 +127,48 @@ MACVTAP_IMAGE_TAGGED=${MACVTAP_IMAGE}:${MACVTAP_TAG}
...
+EOF
+    docker build -t ${LINUX_BRIDGE_IMAGE_TAGGED} .
+)
+
+echo 'Push the image to KubeVirt repo'
+(
+    if [ ! -z ${PUSH_IMAGES} ]; then
+        docker push ${LINUX_BRIDGE_IMAGE_TAGGED}
```
maiqueb, line 167: same here.

phoracek:
/hold

The image releasing part indeed needs to be rethinked.

PR #366, opened 2020-04-09
clone components instead of go get

phoracek:
We cannot use `go get` anymore due to -mod=vendor. This patch replaces it with
mere `git clone` which should be more reliable and predictable.


qinqon:
/hold

Hey you can use go install instead.

phoracek:
@qinqon you assume all components are go projects and compatible with he go setup we have here (go version, flags, etc)

PR #368, opened 2020-04-13
Bump macvtap-cni to 0.2.0

maiqueb:
Bump macvtap-cni to 0.2.0

```release-note
Macvtap-cni now exposes all physical host links (also bonds) in each k8s node by default.
```


maiqueb:
```release-note
Macvtap-cni now exposes all physical host links (also bonds) in each k8s node by default.
```

PR #375, opened 2020-05-12
Knmstate v0.20.0

qinqon:
This PR bump kubernetes-nmstate so webhook are moved to a Deployment and we don't run webhook at a privileged pod anymore, apart from that the cert-manager has being upgraded now it works fine at openshift and self generated keys/certs.

```release-note
knmstate bumped to v0.20.0
```


pkg/network/nmstate.go
```diff
@@ -32,3 +39,36 @@ func renderNMState(conf *opv1alpha1.NetworkAddonsConfigSpec, manifestDir string,
...
+func cleanUpNMState(conf *opv1alpha1.NetworkAddonsConfigSpec, ctx context.Context, client k8sclient.Client) []error {
+	if conf.NMState == nil {
+		return nil
+	}
+
+	// Get existing
+	existing := &unstructured.Unstructured{}
+	gvk := schema.GroupVersionKind{Group: "apps", Version: "v1", Kind: "Deployment"}
+	existing.SetGroupVersionKind(gvk)
+	namespace := "cluster-network-addons"
```
phoracek, line 52: This should be taken from the operand namespace variable.

https://github.com/kubevirt/cluster-network-addons-operator/blob/master/pkg/network/multus.go#L58

data/nmstate/001-rbac.yaml
```diff
@@ -113,9 +123,16 @@ rules:
   - nodes
   - namespaces
   verbs:
-  - get
+  - '*'
```
phoracek, line 126: We talked about this one on the kubernetes-nmstate PR. Do we need this change?

data/nmstate/003-operator.yaml
```diff
@@ -39,78 +40,74 @@ spec:
...
-      - name: dbus-socket
-        hostPath:
-          path: /run/dbus/system_bus_socket
-          type: Socket
-      - name: nmstate-lock
-        hostPath:
-          path: /var/k8s_nmstate
+        - name: tls-key-pair
+          secret:
+            secretName: {{template "handlerPrefix" .}}nmstate-webhook
```
phoracek, line 76: do we have support for this templating on CNAO?

data/nmstate/001-rbac.yaml
qinqon: I changed for cluster role not for normal role, I prefer to keep here as we have at knmstate.

data/nmstate/003-operator.yaml
qinqon: CNAO is already dealing with it at master.

phoracek:
/lgtm
/approve
/hold

Since this is a big change. Would you please deploy CNAO (at least with `make cluster-up cluster-sync`) and test that you are able to create a basic policy?

qinqon:
/hold cancel
we can deploy nncp

qinqon:
/hold

I am going to disable webhook, there is things to figure out at CNAO related to caBundle and secret.

pkg/network/nmstate.go
```diff
@@ -32,3 +40,36 @@ func renderNMState(conf *opv1alpha1.NetworkAddonsConfigSpec, manifestDir string,
...
 }
+
+func cleanUpNMState(conf *opv1alpha1.NetworkAddonsConfigSpec, ctx context.Context, client k8sclient.Client) []error {
+	if conf.NMState == nil {
+		return nil
+	}
+
+	// Get existing
+	existing := &unstructured.Unstructured{}
+	gvk := schema.GroupVersionKind{Group: "apps", Version: "v1", Kind: "Deployment"}
```
phoracek, line 51: It is a DaemonSet

PR #376, opened 2020-05-12
Bump github-release 0.8.1

qinqon:
The latest version do correct github authentication using the HTTP header to do so instead of using the token at the URL.


phoracek:

Not related.

PR #377, opened 2020-05-12
Bump kubemacpool v0.10.0

RamLavi:
This PR bump kubemacpool to version v0.10.0, to introduce opt-in per namespace functionality, cert-manager upgrade so that kubemacpool can run on openshift, fixing flaky tests, and other infra updates.

```release-note
kubemacpool bumped to v0.10.0
```

### DIFF
--- a/hack/components.sh
+++ b/hack/components.sh
@@ -65,6 +65,7 @@ apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: mac-controller-manager
+  namespace: system
 spec:
   template:
     spec:
@@ -78,7 +79,7 @@ rm -rf data/kubemacpool/*
 (
     cd $KUBEMACPOOL_PATH
     kustomize build config/cnao | \
-        sed 's/name: kubemacpool-system/name: {{ .Namespace }}/' | \
+        sed 's/kubemacpool-system/{{ .Namespace }}/' | \
         sed 's/RANGE_START: .*/RANGE_START: {{ .RangeStart }}/' | \
         sed 's/RANGE_END: .*/RANGE_END: {{ .RangeEnd }}/'
 ) > data/kubemacpool/kubemacpool.yaml
--- a/components.yaml
+++ b/components.yaml
@@ -10,7 +10,7 @@ components:
     commit: "2c01b100bcae0061a622d8b82e2c4db4ac092b4c" # 0.2.0
   kubemacpool:
     url: "https://github.com/k8snetworkplumbingwg/kubemacpool"
-    commit: "b3b7bde391b40c3f3f8446626a1efa0b62cf39c9" # v0.8.3
+    commit: "bd3057c490e011624ba7992af0b4e8d6bf8a5d81" # v0.10.0
   nmstate:
     url: "https://github.com/nmstate/kubernetes-nmstate"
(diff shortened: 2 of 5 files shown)


phoracek:
/hold

I want to release with kubernetes-nmstate first as it blocks our deployments. We will release with this right after it. I just want an option to roll back in case we uncover more issues.

hack/components.sh
```diff
@@ -65,6 +65,7 @@ apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: mac-controller-manager
+  namespace: system
```
phoracek, line 68: We should not hardcode namespace. That should be filled in by kustomize.

RamLavi: I'm afraid it's unavoidable in this case. you see, "system" with the added prefix becomes "kubemacpool-system" by kustomize, and only this way the kustomize knows to patch the yamls together. changing it to {{ .Namespace }} ends up with kustomized erroring.

But I will make sure that in the end yaml there would not be any trace of the namespace.

RamLavi: I'm think it's not possible. kustomize requires the namespace in order to attach the patch to the appropriate objects. I tried substituting it with  `"{{ .Namespace }}"` or something alike, but it results with the kustomized failing. 

We already have a substitutes every instance of the namespace to the parameterized version,  but I also added another sed command that makes sure that in the end yaml we will have on the parameterized parameter.
is that ok with you?

phoracek: I don't understand why are you changing this. It used to work (correct me if I'm wrong) without this set. Kustomize is configured here https://github.com/kubevirt/cluster-network-addons-operator/blob/a7d8735895c4752ee11f628d431366f7fde5b662/hack/components.sh#L56 to set the namespace.

hack/components.sh
```diff
@@ -79,6 +80,7 @@ rm -rf data/kubemacpool/*
     cd $KUBEMACPOOL_PATH
     kustomize build config/cnao | \
         sed 's/name: kubemacpool-system/name: {{ .Namespace }}/' | \
+        sed 's/namespace: kubemacpool-system/namespace: {{ .Namespace }}/' | \
```
phoracek, line 83: Where is this needed?

hack/components.sh
RamLavi: when I use `"{{ .Namespace }}"` ( or just nothing) - I get the following error: 
`Error: no matches for OriginalId apps_v1_Deployment|{{ .Namespace }}|mac-controller-manager; no matches for CurrentId apps_v1_Deployment|{{ .Namespace }}|mac-controller-manager; failed to find unique target for patch apps_v1_Deployment|mac-controller-manager
`
and the output kubemacpol.yaml is empty.

I believe it's because I'm using a newer version of kustomize 3.5.4, and they changed to force specifying an explicit namespace, as mentioned in this [issue](https://github.com/fluxcd/flux/issues/2533)
I'm trying to find a less messier solution..

hack/components.sh
RamLavi: mutatingwebhookconfiguration is not caught by the kustmize so it stays `namespace: kubemacpool-system` unless I add this.

I tried manipulating it in kmp project but this mutatingwebhookconfiguration object does not get attached with the kustomize. investigating

phoracek:
Closing in favor of #386

PR #382, opened 2020-05-13
Preserve webhook configuration caBundle and TLS secrets at reconcile

qinqon:
When webhooks are deployed they also include some webhookconfiguration and secrets
that are now updated by other operator (ej, kube-admission-webhook), this PR ignore
reconcile of Secret with type kubernetes.io/tls and also merges MutatingWebhookConfiguration and
ValidatingWebhookConfiguration preserving previous caBundle.


pkg/apply/merge.go
```diff
@@ -112,6 +116,51 @@ func MergeServiceAccountForUpdate(current, updated *unstructured.Unstructured) e
 	return nil
 }
 
+// MergeWebhookConfiguration ensure caBundle is keept at webhooks's clientConfig
+func MergeWebhookConfiguration(current, updated *unstructured.Unstructured) error {
```
phoracek, line 120: does not need to be public. You can test it through the public `MergeObjectForUpdate`

pkg/apply/merge_test.go
```diff
@@ -326,3 +326,115 @@ metadata:
...
+      - operations: ["CREATE", "UPDATE"]
+        apiGroups: ["*"]
+        apiVersions: ["v1alpha1"]
+        resources: ["nodenetworkconfigurationpolicies", "nodenetworkconfigurationpolicies/status"]
+`)
+
+		It("should copy caBundle from current webhooks to updated ones", func() {
+			err := apply.MergeWebhookConfiguration(current, updated)
+			Expect(err).ToNot(HaveOccurred())
+			updatedWebhooks, found, err := unstructured.NestedSlice(updated.Object, "webhooks")
```
phoracek, line 423: let this breathe. insert empty above this one

pkg/apply/merge_test.go
```diff
@@ -326,3 +326,115 @@ metadata:
...
+			Expect(err).ToNot(HaveOccurred())
+			Expect(found).To(BeTrue())
+			for i, _ := range updatedWebhooks {
```
phoracek, line 426: ditto

pkg/apply/merge_test.go
```diff
@@ -326,3 +326,115 @@ metadata:
...
+    rules:
+      - operations: ["CREATE", "UPDATE"]
+        apiGroups: ["*"]
+        apiVersions: ["v1alpha1"]
+        resources: ["nodenetworkconfigurationpolicies", "nodenetworkconfigurationpolicies/status"]
+`)
+
+		It("should copy caBundle from current webhooks to updated ones", func() {
+			err := apply.MergeWebhookConfiguration(current, updated)
+			Expect(err).ToNot(HaveOccurred())
```
phoracek, line 422: Please use expectation strings everywhere. Here is may be "Should successfully execute the merge function"

pkg/apply/merge_test.go
```diff
@@ -326,3 +326,115 @@ metadata:
...
+        name: nmstate-webhook
+        namespace: nmstate
+        path: "/nodenetworkconfigurationpolicies-timestamp-mutate"
+    rules:
+      - operations: ["CREATE", "UPDATE"]
+        apiGroups: ["*"]
+        apiVersions: ["v1alpha1"]
+        resources: ["nodenetworkconfigurationpolicies", "nodenetworkconfigurationpolicies/status"]
+`)
+		updated := k8s.UnstructuredFromYaml(`
```
phoracek, line 377: It is impossible to spot the difference. And the test itself is little hard to decode for me as well.

Would you please keep a single MutatingWebhookConfiguration as a template and then populate it in before each? The constants you use to populate it may be then used in assertions, so it is clear what needs to change and what remain the same.

pkg/apply/merge.go
```diff
@@ -112,6 +116,51 @@ func MergeServiceAccountForUpdate(current, updated *unstructured.Unstructured) e
...
+		}
+
+		updatedWebhooks, found, err := unstructured.NestedSlice(updated.Object, "webhooks")
+		if err != nil {
+			return errors.Wrap(err, "failed searching for webhooks at updated configuration")
+		}
+		if !found {
+			return nil
+		}
+		for ci, _ := range currentWebhooks {
```
phoracek, line 138: Please insert an empty line above

phoracek, line 138: Do we need to iterate ^2? Would you please create two dictionaries `currentWebhookByName` and `updatedWebhookByName` and then compare those with instant access?

qinqon: super early optimization, we are not going to have a massive number of of webhooks.

phoracek: It is not only to optimize the performance but also readability. Precreating of two maps is a clear and well understood operation. Iterating one of them and looking up keys from the other as well. Having two nested loops with the `if` in it is not that clear IMO.

pkg/apply/merge_test.go
```diff
@@ -325,4 +327,67 @@ metadata:
 			Expect(updated.GetUID()).To(Equal(current.GetUID()))
 		})
 	})
+	Context("when current has non empty casBundle and update is empty", func() {
```
phoracek, line 330: typo, caBundle

pkg/apply/merge_test.go
```diff
@@ -325,4 +327,67 @@ metadata:
...
+`
+		var (
+			updated *unstructured.Unstructured
+			current *unstructured.Unstructured
+		)
+		BeforeEach(func() {
+			updated = k8s.UnstructuredFromYaml(fmt.Sprintf(template, "kubernetes-nmstate-2", "", "", ""))
+			current = k8s.UnstructuredFromYaml(fmt.Sprintf(template, "kubernetes-nmstate-1", "caBundle: ca1", "caBundle: ca2", "caBundle: ca3"))
+		})
+		It("should copy caBundle from current webhooks to updated ones", func() {
```
phoracek, line 385: "should now overwrite existing caBundle in webhook configuration" describes better the expectation without mentioning implementation details

pkg/apply/merge.go
phoracek: @qinqon I want to finish this argument

PR #383, opened 2020-05-13
K8s 1.17

qinqon:
The kubernetes-nmstate that is now at CNAO depends on nm-1.22 for that we have to upgrade also the kubevirtci that has proper centos8-stream nodes.


cluster/up.sh
```diff
@@ -32,8 +32,7 @@ if [[ "$KUBEVIRT_PROVIDER" =~ k8s- ]]; then
     echo 'Install NetworkManager on nodes'
     for node in $(./cluster/kubectl.sh get nodes --no-headers | awk '{print $1}'); do
         ./cluster/cli.sh ssh ${node} sudo -- yum install -y yum-plugin-copr
-        ./cluster/cli.sh ssh ${node} sudo -- yum copr enable -y networkmanager/NetworkManager-1.20
-        ./cluster/cli.sh ssh ${node} sudo -- yum install -y NetworkManager NetworkManager-ovs
```
phoracek, line 36: You dropped the install, so now we stay with whatever was installed on the platform and there is no need to add the repo.

PR #390, opened 2020-05-18
Cover all use cases at caBundle

qinqon:
The code that reconcile CABundle at webhookconfiguration was covering only the
case when webhookcofiguration is updated but not when CABundle is updated by a
rotation or created after a webhook delete.

Signed-off-by: Quique Llorente <ellorent@redhat.com>

PR #391, opened 2020-05-19
fix merge reconcile of mutatingWebhookConfiguration when both current…

RamLavi:
… and updated have no caBundle.

Currently when we reconcile webhookConfiguration instances, we make sure to pass the current caBundle to the updated instance should it not have it, but if both don't have a caBundle then it removes the entire webhook section in the webhookConfiguration instance.
in order to fix this, a fix in the logic was inserted, and also the unit test was expanded to include this scenario.

### DIFF
--- a/pkg/apply/merge_test.go
+++ b/pkg/apply/merge_test.go
@@ -277,29 +277,8 @@ metadata:
 			}))
 		})
 	})
-})
 
-var _ = Describe("IsObjectSupported", func() {
-	Context("when given a ServiceAccount with a secret", func() {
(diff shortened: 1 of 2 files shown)


pkg/apply/merge_test.go
```diff
@@ -277,29 +277,8 @@ metadata:
 			}))
 		})
 	})
-})
-
-var _ = Describe("IsObjectSupported", func() {
```
RamLavi, line 282: moved down to include webhook unit tests in `Describe: MergeObjectForUpdate`

RamLavi:
duplicate of #390. closing

PR #386, opened 2020-05-14
Bump kubemacpool v0.11.0

ormergi:
This PR continues #377 kubemacpool bumping:

> This PR bump kubemacpool to version v0.11.0, to introduce opt-in per namespace functionality, cert-manager upgrade so that kubemacpool can run on openshift, fixing flaky tests, and other infra updates.
> 
> kubemacpool bumped to v0.11.0

```release-note
Bump kubemacpool v0.11.0
```

PR #393, opened 2020-05-19
Ensure components objects deletion and creation at e2e tests

ormergi:
This PR add checks in e2e tests to ensure deletion and deployment of Secrets and MutationWebhookConfiguration objects

PR #395, opened 2020-05-20
Bump kubemacpool v0.12.0

ormergi:
This PR continues #377 and #386 Kubemacpool bump

### DIFF
--- a/test/check/components.go
+++ b/test/check/components.go
@@ -13,10 +13,12 @@ type Component struct {
 
 var (
 	KubeMacPoolComponent = Component{
-		ComponentName:      "KubeMacPool",
-		ClusterRole:        "kubemacpool-manager-role",
-		ClusterRoleBinding: "kubemacpool-manager-rolebinding",
-		Deployments:        []string{"kubemacpool-mac-controller-manager"},
+		ComponentName:                "KubeMacPool",
+		ClusterRole:                  "kubemacpool-manager-role",
+		ClusterRoleBinding:           "kubemacpool-manager-rolebinding",
+		Deployments:                  []string{"kubemacpool-mac-controller-manager"},
+		Secret:                       "kubemacpool-service",
+		MutatingWebhookConfiguration: "kubemacpool-mutator",
 	}
 	LinuxBridgeComponent = Component{
 		ComponentName:              "Linux Bridge",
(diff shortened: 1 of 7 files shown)


ormergi:
/hold 
this PR should merge after #393

ormergi:
> Please note [data/kubemacpool/kubemacpool.yaml](https://github.com/kubevirt/cluster-network-addons-operator/pull/395/commits/91e529cdc6277b4277ce60149e9b3e54448a9ed6#diff-2d881ba098d61860b05bb19019451712) is empty.
> you need to re-run `make components` and see why kustomize is broke

It was missing, we should vendor it.

test/e2e/workflow/deployment_test.go
```diff
@@ -165,6 +165,9 @@ var _ = Describe("NetworkAddonsConfig", func() {
 		//2300
 		It("should be able to remove the config and create it again", func() {
 			DeleteConfig()
+			//TODO: remove this checking after this [1] issue is resolved
```
RamLavi, line 168: not sure why this commit is in this PR. can you explain why not in separate?

test/check/components.go
```diff
@@ -13,10 +13,12 @@ type Component struct {
...
-		ComponentName:      "KubeMacPool",
-		ClusterRole:        "kubemacpool-manager-role",
-		ClusterRoleBinding: "kubemacpool-manager-rolebinding",
-		Deployments:        []string{"kubemacpool-mac-controller-manager"},
+		ComponentName:                "KubeMacPool",
+		ClusterRole:                  "kubemacpool-manager-role",
+		ClusterRoleBinding:           "kubemacpool-manager-rolebinding",
+		Deployments:                  []string{"kubemacpool-mac-controller-manager"},
+		Secret:                       "kubemacpool-service",
+		MutatingWebhookConfiguration: "kubemacpool-mutator",
```
RamLavi, line 21: ditto about this commit

test/e2e/workflow/deployment_test.go
ormergi: Without this check the e2e-workflow lane fails

test/check/components.go
ormergi: We need to add these to the component so when the function CheckComponentsRemoval is called, it will take care of Secret and MutatingWebhookConfiguration objects

RamLavi: I agree, but it's not directly related to the bump.
I think it should be in separate PR, like #393 was.

ormergi: It is directly caused by the bump because the changes that has been done in the new release this is why I did not separate it to another PR.
And the e2e-workflow lane will not pass w/o those.

ormergi:
Close in favor #397

PR #405, opened 2020-05-29
bump ovs-cni 0.12.0

phoracek:
**What this PR does / why we need it**:

Bump OVS CNI to v0.12.0. This new version is pinned to Kubernetes 0.18

**Special notes for your reviewer**:

**Release note**:

```release-note
Pin OVS CNI v0.12.0
```

PR #406, opened 2020-05-29
collect init containers

phoracek:

In the current code we collect information only about deployed
containers. With this patch we extend it with init containers.

It will be needed for https://github.com/kubevirt/cluster-network-addons-operator/pull/405

**Release note**:

```release-note
Collect information about deployed init containers
```


test/check/check.go
```diff
@@ -403,7 +403,7 @@ func checkForSecurityContextConstraintsRemoval(name string) error {
 }
 
 func checkForSecretRemoval(name string) error {
-	err := framework.Global.Client.Get(context.Background(), types.NamespacedName{Name: name,Namespace: components.Namespace}, &corev1.Secret{})
+	err := framework.Global.Client.Get(context.Background(), types.NamespacedName{Name: name, Namespace: components.Namespace}, &corev1.Secret{})
```
maiqueb, line 406: `gofmt` made you do it?

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
```diff
@@ -457,6 +446,17 @@ func (r *ReconcileNetworkAddonsConfig) trackDeployedObjects(objs []*unstructured
 	r.statusManager.SetFromPods()
 }
 
+func extendContainerStats(containersStats *[]opv1alpha1.Container, parentKind string, parentName string, containers []v1.Container) {
```
maiqueb, line 449: I'm not very fond of the idea of creating an empty array in the caller side, and then filling it up in the function; what do you think of creating it within, and returning the array ? 

I just think that the caller shouldn't be bothered to be creating those structs beforehand. Feel free to disagree, it's mostly cosmetics; I can `lgtm` if you don't want to do it.

phoracek: What do you suggest to do instead? Return one list for containers, another for init contaiers and join them at the caller side?

maiqueb: that or 

```golang
func extendContainerStats(containersStats *[]opv1alpha1.Container, parentKind string, parentName string, containers []v1.Container...)
```

and return a single list w/ everything. Is the above possible ?

test/check/check.go
phoracek: Yeah :(

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
phoracek: Check it now

maiqueb:
/lgtm

thanks for listening.

phoracek:
Apes together = stronk.

PR #409, opened 2020-05-31
Vendor kustomize

RamLavi:
It install all the needed tools from ./vendor directory using the installed golang version, it also adds
kustomize Makefile targets so it get installed if does not exist.

This simplify the projects by removing docker container indirections and maintenance and also reduce build time, we can do that since all the tools are golang.

**Release note**:

```release-note
NONE
```


Makefile
```diff
@@ -36,6 +33,11 @@ E2E_SUITES = \
 
 BIN_DIR = $(CURDIR)/build/_output/bin/
 GOBIN = $(BIN_DIR)/go/bin/
+export GOFLAGS=-mod=vendor
+export GO111MODULE=on
+export PATH := $(GOBIN):$(PATH)
```
RamLavi, line 38: not sure this is necessary

qinqon:
/hold

@RamLavi can you first make a PR to create the `make tools` target and move all the deps like GINKGO and friends to do `make tools` ? then when merged (I promise it will be fast) we put this

RamLavi:
> /hold
> 
> @RamLavi can you first make a PR to create the `make tools` target and move all the deps like GINKGO and friends to do `make tools` ? then when merged (I promise it will be fast) we put this

https://github.com/kubevirt/cluster-network-addons-operator/pull/410

RamLavi:
not needed anymore. closing PR

PR #413, opened 2020-06-02
remove over privileged rbac from cnao manifest

RamLavi:

Currently the cnao rbac Policy has a segment that elevates the container privileges to all resources, and all verbs.
This is too much for the cnao's needs, and also in turn changes scc to anyuid in cnv/okd clusters.
Hence, the rbac segment is removed in this commit.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/components/components.go
+++ b/pkg/components/components.go
@@ -308,17 +308,6 @@ func GetClusterRole() *rbacv1.ClusterRole {
 					"watch",
 				},
 			},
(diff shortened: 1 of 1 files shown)


RamLavi:
/retest

prow hiccup

phoracek:
This change prevents us from deploying anything on OpenShift. Reverting.

PR #415, opened 2020-06-04
allow listing of all SCC

phoracek:

In 19291e695e6a293ed395bf42ab23aaaa91e68162 we have dropped the
universal ClusterRole allowing us to handle all objects. That introduced
a bug where we were not able to list SCCs, despite we had a rule for it,
it was limited to a single name and resulted in this bug:
https://bugzilla.redhat.com/show_bug.cgi?id=1844057

**Release note**:

```release-note
Fix bug caused my missing SCC rule, preventing CNAO from starting on OpenShift
```


phoracek:
/hold

This fixes only a part of the issue

PR #417, opened 2020-06-04
Fix false positive on the CI and missed bugs

phoracek:
**What this PR does / why we need it**:

Fix operator-sdk test --namespace flag. It has been changed to --operator-namespace. Using --namespace cause operator-sdk to return immediately with false positive.

**Special notes for your reviewer**:

**Release note**:

```release-note
NONE
```

PR #419, opened 2020-06-08
Fix rbac in CNAO's manifest

RamLavi:
Currently the cnao rbac Policy has a segment that elevates the container privileges to all resources, and all verbs.
This is too much for the cnao's needs, and also in turn changes scc to anyuid in cnv/okd clusters.
Hence, the rbac segment is replaced with more refined rbac rules.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/components/components.go
+++ b/pkg/components/components.go
@@ -273,13 +273,26 @@ func GetClusterRole() *rbacv1.ClusterRole {
 				Resources: []string{
 					"securitycontextconstraints",
 				},
+				Verbs: []string{
+					"get",
+					"list",
+					"watch",
+				},
+			},
+			{
+				APIGroups: []string{
+					"security.openshift.io",
+				},
+				Resources: []string{
+					"securitycontextconstraints",
+				},
 				ResourceNames: []string{
 					"privileged",
 				},
 				Verbs: []string{
 					"get",
-					"list",
-					"watch",
+					"patch",
+					"update",
 				},
 			},
 			{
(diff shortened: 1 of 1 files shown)


RamLavi:
/hold

until #417 is merged - the tests are not to be trusted.

pkg/components/components.go
```diff
@@ -271,15 +271,10 @@ func GetClusterRole() *rbacv1.ClusterRole {
 					"security.openshift.io",
 				},
 				Resources: []string{
-					"securitycontextconstraints",
-				},
-				ResourceNames: []string{
-					"privileged",
+					"*",
```
phoracek, line 274: This is what we are trying to avoid if possible. Could you just drop ResourceNames? I think that should be sufficient for our use-case

pkg/components/components.go
```diff
@@ -301,24 +296,128 @@ func GetClusterRole() *rbacv1.ClusterRole {
...
+					"delete",
+					"patch",
+				},
+			},
+			{
+				APIGroups: []string{
+					"kubevirt.io",
+				},
+				Resources: []string{
+					"virtualmachines",
```
phoracek, line 394: I don't think we are creating any VM from the operator, nor list them. We only create another ClusterRule that allows KMP to do that

pkg/components/components.go
```diff
@@ -301,24 +296,128 @@ func GetClusterRole() *rbacv1.ClusterRole {
...
+					"list",
+					"watch",
+					"create",
+					"update",
+					"patch",
+				},
+			},
+			{
+				APIGroups: []string{
+					"policy",
```
phoracek, line 407: We can drop this now, can't we?

pkg/components/components.go
```diff
@@ -301,24 +296,128 @@ func GetClusterRole() *rbacv1.ClusterRole {
...
+					"mutatingwebhookconfigurations",
+					"validatingwebhookconfigurations",
+				},
+				Verbs: []string{
+					"*",
+				},
+			},
+			{
+				APIGroups: []string{
+					"k8s.cni.cncf.io",
```
phoracek, line 364: This is for NAD, right? We are not creating any network attachment definition from the operator

pkg/components/components.go
```diff
@@ -301,24 +296,128 @@ func GetClusterRole() *rbacv1.ClusterRole {
...
 			},
+			{
+				APIGroups: []string{
+					"apiextensions.k8s.io",
+				},
+				Resources: []string{
+					"customresourcedefinitions",
+				},
+				Verbs: []string{
+					"get",
```
phoracek, line 381: just use "*" then

pkg/components/components.go
```diff
@@ -301,24 +296,128 @@ func GetClusterRole() *rbacv1.ClusterRole {
...
 			},
 			{
 				APIGroups: []string{
+					"",
+				},
+				Resources: []string{
+					"services",
+					"configmaps",
+					"namespaces",
+					"nodes",
```
phoracek, line 316: we don't need to do anything with nodes

pkg/components/components.go
```diff
@@ -301,24 +296,128 @@ func GetClusterRole() *rbacv1.ClusterRole {
...
+					"nodes/status",
+					"serviceaccounts",
+					"pods",
+					"pods/status",
+					"events",
```
phoracek, line 321: not needed

pkg/components/components.go
```diff
@@ -301,24 +296,128 @@ func GetClusterRole() *rbacv1.ClusterRole {
...
+				},
+				Resources: []string{
+					"services",
+					"configmaps",
+					"namespaces",
+					"nodes",
+					"nodes/status",
+					"serviceaccounts",
+					"pods",
+					"pods/status",
```
phoracek, line 320: not needed

pkg/components/components.go
```diff
@@ -301,24 +296,128 @@ func GetClusterRole() *rbacv1.ClusterRole {
...
 			{
 				APIGroups: []string{
+					"",
+				},
+				Resources: []string{
+					"services",
+					"configmaps",
+					"namespaces",
+					"nodes",
+					"nodes/status",
```
phoracek, line 317: not needed

RamLavi:
@phoracek 
```
cluster-network-addons-operator (fix_rbac) $ oc get pods -n cluster-network-addons cluster-network-addons-operator-858cc57c59-g9vqg -oyaml | grep scc -A 5
    openshift.io/scc: restricted
  creationTimestamp: "2020-06-11T07:08:48Z"
  generateName: cluster-network-addons-operator-858cc57c59-
  labels:
    name: cluster-network-addons-operator
    pod-template-hash: 858cc57c59
```

RamLavi:
@phoracek if we want to reduce the cluster rbac further we need to do some work in each cnao sub-project, where we re-examine each rbac to make sure we really need it.
But it's not in the scope of this PR of course

PR #428, opened 2020-06-18
merge some upgrade tests

phoracek:

For each release, we test upgrades from every previous release to the
newly created one. That takes a lot of time and is obviously getting
only worse.

This patch merges two of the three upgrade tests into a single one. They
are both just checking the state of the cluster, so we are not risking
too much by merging them.

This should save us 30 % of the time on the suite.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/test/e2e/lifecycle/upgrade_test.go
+++ b/test/e2e/lifecycle/upgrade_test.go
@@ -42,11 +42,11 @@ var _ = Context("Cluster Network Addons Operator", func() {
 					CheckConfigVersions(expectedOperatorVersion, expectedObservedVersion, expectedTargetVersion, podsDeploymentTimeout, CheckDoNotRepeat)
 				})
 
-				It("it should report expected deployed container images", func() {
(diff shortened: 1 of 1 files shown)


test/e2e/lifecycle/upgrade_test.go
```diff
@@ -42,11 +42,11 @@ var _ = Context("Cluster Network Addons Operator", func() {
...
 				})
 
-				It("it should report expected deployed container images", func() {
+				It("it should report expected deployed container images and leave no leftovers from the previous version", func() {
+					By("Checking reported container images")
 					CheckReleaseUsesExpectedContainerImages(newRelease)
-				})
 
-				It("should run with no leftovers from the original version", func() {
+					By("Checking for leftover objects from the previous version")
```
qinqon, line 49: Can we Sprintf the previous version ?

phoracek: Done in the context

phoracek:

PR #436, opened 2020-06-23
fix version in README

phoracek:

Automation on release 0.37, that left the readme with outdated version
and our scripts were not able to update it since (since they just sed
old version to new). With this patch we set the correct version.

Resolves: #446

**Release note**:

```release-note
NONE
```


phoracek:

PR #439, opened 2020-06-26
test upgrades from 0.27.7 instead of 0.27.0

phoracek:

Use the last z-stream release of that branch to test the real upgrade
flow. Somebody can upgrade from any of the z-stream releases, but in
order to save resources, we will test only the latest one.

**Release note**:

```release-note
Test upgrades from 0.27.7 z-stream instead of 0.27.0
```

### DIFF
--- a/automation/check-patch.e2e-lifecycle-k8s.sh
+++ b/automation/check-patch.e2e-lifecycle-k8s.sh
@@ -27,7 +27,7 @@ main() {
         export RELEASES_DESELECTOR="${to_be_released}"
     else
         # Don't run all upgrade tests in regular PRs, stick to those released under HCO
-        export RELEASES_SELECTOR="{0.18.0,0.23.0,0.27.0,99.0.0}"
+        export RELEASES_SELECTOR="{0.18.0,0.23.0,0.27.7,99.0.0}"
     fi
 
     make cluster-down
--- a/manifests/cluster-network-addons/0.27.7/cluster-network-addons-operator.0.27.7.clusterserviceversion.yaml
+++ b/manifests/cluster-network-addons/0.27.7/cluster-network-addons-operator.0.27.7.clusterserviceversion.yaml
@@ -1,0 +1,206 @@
+apiVersion: operators.coreos.com/v1alpha1
+kind: ClusterServiceVersion
+metadata:
+  name: cluster-network-addons-operator.0.27.7
+  namespace: placeholder
+  annotations:
+    capabilities: "Full Lifecycle"
+    categories: "Networking"
+    alm-examples: |
+      [
+        {
+          "apiVersion":"networkaddonsoperator.network.kubevirt.io/v1alpha1",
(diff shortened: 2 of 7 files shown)


automation/check-patch.e2e-lifecycle-k8s.sh
```diff
@@ -27,7 +27,7 @@ main() {
         export RELEASES_DESELECTOR="${to_be_released}"
     else
         # Don't run all upgrade tests in regular PRs, stick to those released under HCO
-        export RELEASES_SELECTOR="{0.18.0,0.23.0,0.27.0,99.0.0}"
+        export RELEASES_SELECTOR="{0.18.0,0.23.0,0.27.7,99.0.0}"
```
oshoval, line 30: are spaces allowed here between each release tag ?

manifests/cluster-network-addons/0.27.7/cluster-network-addons-operator.0.27.7.clusterserviceversion.yaml
```diff
@@ -0,0 +1,206 @@
...
+            "imagePullPolicy": "IfNotPresent"
+          }
+        }
+      ]
+    description: Additional networking components for Kubernetes
+spec:
+  displayName: Cluster Network Addons
+  description: Deploy additional networking components for Kubernetes
+  keywords:
+    - Networking
```
oshoval, line 35: just making sure, no need here mac v tap or other something that was dropped?

automation/check-patch.e2e-lifecycle-k8s.sh
phoracek: Not in the current implementation AFAIK.

manifests/cluster-network-addons/0.27.7/network-addons-config-example.cr.yaml
```diff
@@ -0,0 +1,12 @@
+---
+apiVersion: networkaddonsoperator.network.kubevirt.io/v1alpha1
+kind: NetworkAddonsConfig
+metadata:
+  name: cluster
+spec:
+  imagePullPolicy: IfNotPresent
+  kubeMacPool: {}
```
oshoval, line 8: aint there was mac v tap around ? or on other version / i am wrong

manifests/cluster-network-addons/0.27.7/cluster-network-addons-operator.0.27.7.clusterserviceversion.yaml
phoracek: Yeah, it should be updated. Not a part of this PR.

manifests/cluster-network-addons/0.27.7/network-addons-config-example.cr.yaml
phoracek: This is version 0.27.7.  The was not macvtap back then.

PR #448, opened 2020-07-05
Split components to sub scripts

RamLavi:
Currently componets.sh script changes all components.
Let's move it to a components folder and split it per component.

**Release note**:

```release-note
NONE
```


hack/components/bump_knmstate.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
+
+set -xe
+
+source hack/components/common_functions.sh
+
+NMSTATE_URL=$(cat components.yaml | shyaml get-value components.nmstate.url)
```
oshoval, line 7: do we need to install shyaml ?

hack/components/components.sh
```diff
@@ -0,0 +1,20 @@
...
+
+source hack/components/common_functions.sh
+CNAO_VERSION=${VERSION} # Exported from Makefile
+
+echo 'Setup temporary Go path'
+export GOPATH=${PWD}/_components/go
+mkdir -p $GOPATH
+export PATH=${GOPATH}/bin:${PATH}
+
+echo 'kubemacpool'
```
oshoval, line 13: maybe worth to create array of the components and a function that will echo and then source them ?
just those 3 we have atm ?

hack/components/bump_kubemacpool.sh
```diff
@@ -0,0 +1,59 @@
...
+
+echo 'Fetch kubemacpool sources'
+fetch_component ${KUBEMACPOOL_PATH} ${KUBEMACPOOL_URL} ${KUBEMACPOOL_COMMIT}
+
+echo 'Configure kustomize for CNAO templates and save the rendered manifest under CNAO data'
+(
+    cd ${KUBEMACPOOL_PATH}
+    mkdir -p config/cnao
+
+    cat <<EOF > config/cnao/kustomization.yaml
```
oshoval, line 20: nit:
just if you want, you can use ident, with tab only, if there arent variables i think
cat <<-EOF
    indented
    EOF
echo Done

hack/components/bump_macvtap.sh
```diff
@@ -0,0 +1,40 @@
...
+MACVTAP_REPO=$(echo ${MACVTAP_URL} | sed 's#https://\(.*\)#\1#')
+MACVTAP_PATH=${GOPATH}/src/${MACVTAP_REPO}
+
+echo 'Fetch macvtap-cni sources'
+fetch_component ${MACVTAP_PATH} ${MACVTAP_URL} ${MACVTAP_COMMIT}
+
+rm -rf data/macvtap/*
+echo 'Copy the templates from the macvtap-cni repo ...'
+cp ${MACVTAP_PATH}/templates/namespace.yaml.in data/macvtap/000-ns.yaml
+echo "{{ if .EnableSCC }}" >> data/macvtap/001-rbac.yaml
```
oshoval, line 18: maybe make those couple lines (appending 001-rbac.yaml) with HEREDOC and variables ?
(in case you can append with HERDOC)

hack/components/common_functions.sh
```diff
@@ -0,0 +1,30 @@
...
+    destination=$1
+    url=$2
+    commit=$3
+
+    if [ ! -d ${destination} ]; then
+        mkdir -p ${destination}
+        git clone ${url} ${destination}
+    fi
+
+    (
```
oshoval, line 15: why sometimes you open this scope ?

hack/components/common_functions.sh
```diff
@@ -0,0 +1,30 @@
...
+}
+
+function get_component_tag() {
+    component_dir=$1
+
+    (
+        cd ${component_dir}
+        git describe --tags
+    )
+}
```
oshoval, line 30: allover, maybe worth to enter at the end ?

hack/components/bump_knmstate.sh
RamLavi: yes. There are alternatives but I had no objections keeping it, as it's a dev tool doing a dev job..

oshoval: maybe then worth to verify it exists in the main script ?
or use a container of it

RamLavi: you mean install it if it's not installed?

oshoval: i dont think people would like the script to install, either just warn + exit 1 before doing anything
or use a containerized version

hack/components/bump_kubemacpool.sh
RamLavi: sounds nice, but I tried but it cancels indents of internal lines, for example:
if I have
```
cat <<-EOF
(tab)X1:
(tab)(tab)X1.1: value
(tab)X2:
(tab)EOF
```

it flatten all tabs to:
```
X1:
X1.1= value
X2:
```

oshoval: (tab)EOF
i think shouldnt be with tab ?
https://unix.stackexchange.com/questions/353076/how-to-indent-an-heredoc-inside-an-heredoc-the-right-way?noredirect=1&lq=1

Makefile
```diff
@@ -157,7 +157,7 @@ gen-k8s-check: $(apis_sources)
 	touch $@
 
 components:
-	./hack/components.sh
+	./hack/components/components.sh
```
qinqon, line 160: I think an verb would be nicer `bump-all.sh` ?

hack/components/bump_macvtap.sh
RamLavi: I'm all forward to making the scripts better, but it's not in the scope of this PR, which only intends to split them to separate scripts.

hack/components/bump_knmstate.sh
qinqon: +1 on containerized

Also having some `get_component_url` and friends at common functions would be nice so we wrap the shyaml call at one place and we decide there how to run it.

hack/components/bump_knmstate.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
+
+set -xe
```
qinqon, line 3: We have a lot of piping add pipefail here.

hack/components/bump_kubemacpool.sh
```diff
@@ -0,0 +1,59 @@
+#!/usr/bin/env bash
+
+set -xe
```
qinqon, line 3: ditto

hack/components/common_functions.sh
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env bash
+
+set -xe
+
+function fetch_component() {
+    destination=$1
+    url=$2
+    commit=$3
```
qinqon, line 8: Is usually good practice to use `local` prefix so they don't get exposure out of the function.

hack/components/common_functions.sh
```diff
@@ -0,0 +1,30 @@
...
+
+    if [ ! -d ${destination} ]; then
+        mkdir -p ${destination}
+        git clone ${url} ${destination}
+    fi
+
+    (
+        cd ${destination}
+        git fetch origin
+        git reset --hard
```
qinqon, line 18: Just do `git reset ${commit} --hard`

hack/components/common_functions.sh
RamLavi: this is not a scope, the `()` opens a subshell that performs all the commands in it.

hack/components/components.sh
```diff
@@ -0,0 +1,20 @@
...
+source hack/components/bump_kubemacpool.sh
+
+echo 'macvtap'
+source hack/components/bump_macvtap.sh
+
+echo 'kubernetes-nmstate'
+source hack/components/bump_knmstate.sh
```
qinqon, line 20: Let's do a loop for this and has a var about components `components="kubemacpool macvtap kubernetes-nmstate` we will need to rename `bump_knmstate.sh` to `bump_kubernetes-nmtate`

RamLavi: np I'll add it in the next PR #449

hack/components/bump_kubemacpool.sh
RamLavi: From what I saw EOF should also be indented if we use <<-EOF

hack/components/common_functions.sh
qinqon: Is about keeping path, since it opens subshell after it finished we will keep the previous path

oshoval: thanks, what about pushd popd ?

hack/components/bump-knmstate.sh
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env bash
+
+set -xe pipefail
```
oshoval, line 3: doesnt pipefail need -o ? or i mix stuff

RamLavi: @qinqon ?

Makefile
EdDev: You need to change this and the location of the script in the same commit.

hack/components/common_functions.sh
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env bash
+
+set -xe
+
+function fetch_component() {
```
EdDev: - The order of the commits seem wrong, as this change comes after the one that calls it.
- Lets try to avoid "common", "utils" and friends, they just point us to the junkyard of all operations we do not know to name. How about calling this one "component"? BTW, what components are these?
- When using `source`, we loose context of where this is coming from. I saw Daniel B. using <filename>::<funcname>, e.g. `component::fetch`.

Makefile
RamLavi: not sure what you mean. I did change the this and the location in the same commit. can you elaborate?

EdDev: Now it looks fine, either github tricked me or I missed it.

hack/components/bump_kubemacpool.sh
EdDev: Not related to this change, but just wandering why the files are build inline and not static?

hack/components/components.sh
```diff
@@ -3,7 +3,6 @@
 set -xe
 
 source hack/components/common_functions.sh
-
```
EdDev: nit: not a related change.

hack/components/components.sh
```diff
@@ -12,59 +11,7 @@ mkdir -p $GOPATH
...
-        sed 's/RANGE_END: .*/RANGE_END: {{ .RangeEnd }}/'
-) > data/kubemacpool/kubemacpool.yaml
-
-echo 'Get kubemacpool image name and update it under CNAO'
-KUBEMACPOOL_TAG=$(get_component_tag ${KUBEMACPOOL_PATH})
-KUBEMACPOOL_IMAGE=quay.io/kubevirt/kubemacpool
-KUBEMACPOOL_IMAGE_TAGGED=${KUBEMACPOOL_IMAGE}:${KUBEMACPOOL_TAG}
-sed -i "s#\"${KUBEMACPOOL_IMAGE}:.*\"#\"${KUBEMACPOOL_IMAGE_TAGGED}\"#" pkg/components/components.go
-sed -i "s#\"${KUBEMACPOOL_IMAGE}:.*\"#\"${KUBEMACPOOL_IMAGE_TAGGED}\"#" test/releases/${CNAO_VERSION}.go
+source hack/components/bump_kubemacpool.sh
```
EdDev: Can you just call it? Is there a real need to source it?

hack/components/components.sh
```diff
@@ -14,40 +14,7 @@ echo 'kubemacpool'
...
-cat ${MACVTAP_PATH}/templates/macvtap.yaml.in >> data/macvtap/002-macvtap-daemonset.yaml
-
-echo 'Get macvtap-cni image name and update it under CNAO'
-MACVTAP_TAG=$(get_component_tag ${MACVTAP_PATH})
-MACVTAP_IMAGE=quay.io/kubevirt/macvtap-cni
-MACVTAP_IMAGE_TAGGED=${MACVTAP_IMAGE}:${MACVTAP_TAG}
-sed -i "s#\"${MACVTAP_IMAGE}:.*\"#\"${MACVTAP_IMAGE_TAGGED}\"#" pkg/components/components.go
-# TODO: uncomment the following line *once* there is macvtap upgrade is supported
-#sed -i "s#\"${MACVTAP_IMAGE}:.*\"#\"${MACVTAP_IMAGE_TAGGED}\"#" test/releases/${CNAO_VERSION}.go
+source hack/components/bump_macvtap.sh
```
EdDev: Same here.

hack/components/common_functions.sh
```diff
@@ -27,4 +27,4 @@ function get_component_tag() {
         cd ${component_dir}
         git describe --tags
     )
-}
\ No newline at end of file
+}
```
EdDev: Change in the wrong commit.

hack/components/components.sh
```diff
@@ -16,22 +16,5 @@ source hack/components/bump_kubemacpool.sh
...
-sed -i "s#\"${NMSTATE_IMAGE}:.*\"#\"${NMSTATE_IMAGE_TAGGED}\"#" \
-    pkg/components/components.go \
-    test/releases/${CNAO_VERSION}.go
-
-echo 'Copy kubernetes-nmstate manifests'
-rm -rf data/nmstate/*
-cp $NMSTATE_PATH/deploy/handler/* data/nmstate/
-cp $NMSTATE_PATH/deploy/crds/*nodenetwork*crd* data/nmstate/
+echo 'kubernetes-nmstate'
+source hack/components/bump_knmstate.sh
```
EdDev: Same here.

hack/components/common_functions.sh
RamLavi: - I rechecked the order of the commits, can you explain what you find ammis with it?
- I agree, I'll change it into 2 scripts: "git_utils.sh" and "yaml_utils.sh"
- I agree, will change the functions names accordingly

hack/components/components.sh
RamLavi: well, the commit states it "split kubernetes-nmstate to separate script", and you can see here that I remove the script and sourced it to the new place. 
Can you explain why you say it's not in the right commit?

hack/components/components.sh
RamLavi: again not sure what you mean. I didn't change this line in this commit. Can you explain what do you mean?

hack/components/common_functions.sh
RamLavi: will fix

Makefile
RamLavi: ack, changing to resolved

hack/components/components.sh
RamLavi: do you mean to do `./hack/components/bump_kubemacpool.sh`?
I can I think, but what's the difference?

hack/components/bump_kubemacpool.sh
RamLavi: I think that using heredocs here removes the need to have a kustomization file that is not directly related to the repo. It keeps thing neat and tidy.

hack/components/components.sh
RamLavi: The commit tries to "split macvtap to separate script", and these lines changes are appropriate to this commit. please recheck

hack/components/components.sh
RamLavi: wil change

oshoval:
Thanks
/lgtm

phoracek:
/hold

Waiting for Edy to ack

hack/components/bump-knmstate.sh
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env bash
+
+set -xeo pipefail
+
+source hack/components/yaml-utils.sh
+source hack/components/git-utils.sh
+
+NMSTATE_URL=$(yaml-utils::get_component_url nmstate)
```
EdDev, line 8: The usage of sub-shelling here is strange.
- If you sub-shell, then you could just do `./hack/components/yaml-utils.sh --get-component-url nmstate` and there is no need for sourcing. The advantage of sourcing as I see it, is the avoidance of sub-shelling.
- Another alternative is for the function to return the output and sub-shell if needed there.

I lost track in this review of what existed and what was actually added, so if it was already like this before, feel free to do this in a following PR.

hack/components/bump-all.sh
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
+
+set -xeo pipefail
+
+echo 'Setup temporary Go path'
+export GOPATH=${PWD}/_components/go
+mkdir -p $GOPATH
+export PATH=${GOPATH}/bin:${PATH}
```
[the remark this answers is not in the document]
RamLavi: it did already existing in the script, so I'll post a fix in a separate PR.

hack/components/bump-knmstate.sh
RamLavi: not sure what you mean by strange. can you elaborate what's wrong with it, so that I could unserstand why your suggestions are better for this situation?

hack/components/bump-all.sh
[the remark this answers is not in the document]
EdDev: ack

hack/components/bump-knmstate.sh
EdDev: If you sub-shell, you do not need function calls, you just need to call the script.
I would expect with `$()` to see shell commands, not calls to functions: `$(ls -la)`.

phoracek: I don't know about performance cost, but when it comes to readability, I am a fan of subshell called on sourced functions.

It allows us to do something close to imports and calling functions from scopes. The alternative would be to either have one shell script per function or to keep a boilerplate to select the right function based on an attribute.

Is there any specific concern regarding this?

RamLavi: Following a talk with Edi, we agreed I will investigate other options in other PR, as it is not crucial for this PR.

EdDev: > I don't know about performance cost, but when it comes to readability, I am a fan of subshell called on sourced functions.
> 
> It allows us to do something close to imports and calling functions from scopes. The alternative would be to either have one shell script per function or to keep a boilerplate to select the right function based on an attribute.
> 
> Is there any specific concern regarding this?

The issue is with sourcing. If you sub-shell, there is little reason to source.

RamLavi:
/hold cancel 
After Edi's ack. 

/retest

phoracek:

PR #449, opened 2020-07-05
move bump-all.sh functionality to Makefile

RamLavi:
We don't need bump-all.sh script, we can just use Makefile.

**Release note**:

```release-note
NONE
```


RamLavi:
/hold 
relies on #448 so will only unheld after it's merged

hack/components/components.sh
```diff
@@ -10,11 +10,27 @@ export GOPATH=${PWD}/_components/go
 mkdir -p $GOPATH
 export PATH=${GOPATH}/bin:${PATH}
 
-echo 'kubemacpool'
-source hack/components/bump_kubemacpool.sh
+function usage() {
+        echo "Usage: COMPONENT=<component-name> components.sh\n"
```
oshoval, line 14: cant run more than 1 at once ?
well its fine anyway

hack/components/components.sh
```diff
@@ -10,11 +10,27 @@ export GOPATH=${PWD}/_components/go
...
+}
 
-echo 'macvtap'
-source hack/components/bump_macvtap.sh
-
-echo 'kubernetes-nmstate'
-source hack/components/bump_knmstate.sh
+#input parameter parsing
+case "$COMPONENT" in
+  knmstate)
```
qinqon, line 20: What about adding `all` ?

Makefile
```diff
@@ -157,7 +157,7 @@ gen-k8s-check: $(apis_sources)
 	touch $@
 
 components:
```
qinqon, line 159: The number of components is low is not like we add new componentes everyweek so we should have something like 
`make kubemacpool-bump` and friends and also `make all-components-bump`

hack/components/components.sh
```diff
@@ -10,11 +10,27 @@ export GOPATH=${PWD}/_components/go
...
+      echo 'knmstate'
+      source hack/components/bump_knmstate.sh
+      ;;
+  kubemacpool)
+      echo 'kubemacpool'
+      source hack/components/bump_kubemacpool.sh
```
qinqon, line 26:
```bash
bump_component() {
local component=$1
echo $component
source hack/components/bump_${component}.sh
}
```

And then use this function directly at cases.

Makefile
RamLavi: I was thinking of changing bump-all.sh to accept COMPONENTS arg so it would itterate on them

Makefile
```diff
@@ -156,8 +156,8 @@ gen-k8s-check: $(apis_sources)
 	./hack/verify-codegen.sh
 	touch $@
 
-bump-all:
-	./hack/components/bump-all.sh
+bump-components:
```
qinqon, line 159: I think we don't need bump-components.sh just small Makefile target is enough.

```Makefile
bump-*:
    CNAO_VERSION=${VERSION} ./hack/components/bump-$@.sh
bump-all: bump-kubemacpool bump-kubernets-nmstate bump ....
```

Or even more exlicit with target per component

RamLavi: sure, do we mind if we use the regular go folder and not the temp as in the [start](https://github.com/kubevirt/cluster-network-addons-operator/pull/449/files#diff-a15786f49370ff37a2c3d11d804b3eeeR5) of the bump_components.sh:

Makefile
```diff
@@ -156,8 +156,8 @@ gen-k8s-check: $(apis_sources)
...
+	COMPONENTS="${COMPONENTS}" ./hack/components/bump-components.sh
```
qinqon, line 160: You don't need this one anymore right ?

hack/components/bump-knmstate.sh
```diff
@@ -5,10 +5,11 @@ set -xeo pipefail
 source hack/components/yaml-utils.sh
 source hack/components/git-utils.sh
 
+echo kubernetes-nmstate
 NMSTATE_URL=$(yaml-utils::get_component_url nmstate)
 NMSTATE_COMMIT=$(yaml-utils::get_component_commit nmstate)
 NMSTATE_REPO=$(yaml-utils::get_component_repo ${NMSTATE_URL})
-NMSTATE_PATH=${GOPATH}/src/${NMSTATE_REPO}
+NMSTATE_PATH=$(pwd)/_components/go/src/${NMSTATE_REPO}
```
qinqon, line 12: Why do you need this under go/src ?

RamLavi: the script needs to clone the repo in order to get the tag and also maybe other stuff like the manifests.
using `${GOPATH}/src/${NMSTATE_REPO}` will change my working kubernetes-nmstate repo which I don't want (luckily I didn't have any unpushed code there ^^ ). 

In the past this was handled on bump.components.sh where we would manually change the GOPATH.. 
but I since we're really using go here then I just changed it manually to `$(pwd)/_components/go/src/${NMSTATE_REPO}`

qinqon: What I mean is that you don't need to do any reference to go you can clone it even at a temporal dir with `mktmp -d` and remove that dir at the end of the scripts some signal catcher.

hack/components/git-utils.sh
```diff
@@ -26,3 +26,8 @@ function git-utils::get_component_tag() {
         git describe --tags
     )
 }
+
+function git-utils::create_temp_path() {
+	TEMP_DIR=$(mktemp -d -p .)
+	trap "rm -rf ${TEMP_DIR}" EXIT
```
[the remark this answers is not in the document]
phoracek: Not sure if we can set a trap inside a function. Moreover, we cannot set multiple traps per EXIT. I'd say that the caller of this function has to ensure cleanup.

Or we can just drop this, it will be removed eventually.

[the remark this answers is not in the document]
RamLavi: It works because I don't run it with a subshell. If it were to do the rm at the exit of the function - then the clone would have failed.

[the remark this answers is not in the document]
RamLavi: Can you explain why you say there are multiple traps? I think each bump script runs standalone bt the MakeFile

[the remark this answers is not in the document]
phoracek: Yes, but you define the trap in utils. If the module importing this function and using it had predefined trap for EXIT, you would shadow it with this one.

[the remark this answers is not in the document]
RamLavi: I can do the trap inside the father script, but I think we shouldn't because:
1. It works since I'm sourcing this script, so as long as I don't run it in a subshell, it should just run as if the lines are in the father script.
2. The point of using this function is to run the same lines in all the scripts, so moving the trap line will reduce maintainabilty

What do you think? 😀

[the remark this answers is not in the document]
RamLavi: > Yes, but you define the trap in utils. If the module importing this function and using it had predefined trap for EXIT, you would shadow it with this one.

Oh, I get it.. You mean this way it's hidden.. OK agree with that

hack/components/bump-knmstate.sh
```diff
@@ -5,10 +5,14 @@ set -xeo pipefail
...
 source hack/components/git-utils.sh
 
+echo kubernetes-nmstate
 NMSTATE_URL=$(yaml-utils::get_component_url nmstate)
 NMSTATE_COMMIT=$(yaml-utils::get_component_commit nmstate)
 NMSTATE_REPO=$(yaml-utils::get_component_repo ${NMSTATE_URL})
-NMSTATE_PATH=${GOPATH}/src/${NMSTATE_REPO}
+
+TEMP_DIR=$(git-utils::create_temp_path)
+trap "rm -rf ${TEMP_DIR}" EXIT
```
oshoval, line 14: i read somewhere that an error in a function wont trigger ERR, unless used set -E (instead of set -e),
well here its EXIT, maybe worth to make sure its indeed caught, even on other file if its the case and so on (there are arguments which control if set args are propagated between files and so on, so to see we are on road)

RamLavi: I added a wrong cmd in `git-utils::get_component_tag` and the trap was indeed triggered

oshoval: Great thx

RamLavi: thank you :)

hack/components/bump-knmstate.sh
```diff
@@ -5,10 +5,14 @@ set -xeo pipefail
 source hack/components/yaml-utils.sh
 source hack/components/git-utils.sh
 
+echo kubernetes-nmstate
```
oshoval, line 8: maybe change to ?
"Bumping kubernetes-nmstate"
(same for other files)

hack/components/git-utils.sh
```diff
@@ -26,3 +26,7 @@ function git-utils::get_component_tag() {
         git describe --tags
     )
 }
+
+function git-utils::create_temp_path() {
+	mktemp -d -p .
```
oshoval, line 31: i think its better to create under /tmp
even that its deleted at the end, 
its cleaner and safer in case script crashes

RamLavi: it's possible and I did that at first but eventually preferred this way. here's my reasoning:

if I create the temp folder inside `./tmp` then the TEMP_PATH  wil be (for example) `./tmp/tmp.JzzNzXFcEh/`
and the the trap `trap "rm -rf ${TEMP_DIR}" EXIT` will not delete the ./tmp folder. 
Of course I can do `trap "rm -rf ./tmp" EXIT`, but I prefer to only delete the folder i created, and not the entire ./tmp folder that may contain other stuff that are running now.

Again - only my personal preferece.

oshoval: totally meant a folder at /tmp
i.e /tmp/XXYY

RamLavi: understood what you mean. will change.

oshoval:
Thanks 
/lgtm

PR #450, opened 2020-07-05
adding linux-bridge bump script

RamLavi:
Currently cnao only bumps linux-bridge manually.
Let's add a script to do it for us.
This PR is a subset of an older PR #363 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/hack/components/bump-linux-bridge.sh
+++ b/hack/components/bump-linux-bridge.sh
@@ -1,0 +1,49 @@
+#!/usr/bin/env bash
+
+set -xe
+
+source hack/components/yaml-utils.sh
+source hack/components/git-utils.sh
+
+echo Bumping linux-bridge
+LINUX_BRIDGE_URL=$(yaml-utils::get_component_url linux-bridge)
+LINUX_BRIDGE_COMMIT=$(yaml-utils::get_component_commit linux-bridge)
+LINUX_BRIDGE_REPO=$(yaml-utils::get_component_repo ${LINUX_BRIDGE_URL})
+
+TEMP_DIR=$(git-utils::create_temp_path linux-bridge)
+trap "rm -rf ${TEMP_DIR}" EXIT
+LINUX_BRIDGE_PATH=${TEMP_DIR}/${LINUX_BRIDGE_REPO}
+
+echo 'Fetch linux-bridge sources'
+git-utils::fetch_component ${LINUX_BRIDGE_PATH} ${LINUX_BRIDGE_URL} ${LINUX_BRIDGE_COMMIT}
+
+echo 'Get linux-bridge tag'
+LINUX_BRIDGE_TAG=$(git-utils::get_component_tag ${LINUX_BRIDGE_PATH})
+
+echo 'Build container image with linux-bridge binaries'
+LINUX_BRIDGE_TAR_CONTAINER_DIR=/usr/src/github.com/containernetworking/plugins/bin
+LINUX_BRIDGE_TAR_FILE=cni-plugins-linux-amd64-${LINUX_BRIDGE_TAG}.tgz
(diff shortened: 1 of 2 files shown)


oshoval:
```
In order to bump it, we add a script that will create a container with the appropriate linux-bridge tagged image, and push it to the registry for local use.
```
long line in description, please try to use max 72 char in body

hack/components/bump_linux_bridge.sh
```diff
@@ -0,0 +1,48 @@
...
+LINUX_BRIDGE_PATH=${GOPATH}/src/${LINUX_BRIDGE_REPO}
+
+echo 'Fetch linux-bridge sources'
+fetch_component ${LINUX_BRIDGE_PATH} ${LINUX_BRIDGE_URL} ${LINUX_BRIDGE_COMMIT}
+
+echo 'Get linux-bridge tag'
+LINUX_BRIDGE_TAG=$(get_component_tag ${LINUX_BRIDGE_PATH})
+
+echo 'Build container image with linux-bridge binaries'
+LINUX_BRIDGE_IMAGE=quay.io/kubevirt/cni-default-plugins
```
oshoval, line 19: generally speaking, maybe this worth to be a variable with default value, so one can override it, and bump the results also to his personal repository (need to be parameterized as well then of course), in order to create a custom bundle?

RamLavi:
/hold
awaits merge of #448 and #449

hack/components/bump_linux_bridge.sh
```diff
@@ -0,0 +1,48 @@
...
+LINUX_BRIDGE_IMAGE_TAGGED=${LINUX_BRIDGE_IMAGE}:${LINUX_BRIDGE_TAG}
+(
+    cd ${LINUX_BRIDGE_PATH}
+    cat <<EOF > Dockerfile
+FROM golang:1.13 AS builder
```
qinqon, line 24: Building and pushing linux-bridge at CNAO feels strange, how were we doing it? maybe it deserve separate repo.

phoracek: We were doing it manually until now.

Separate repository would be a pain to maintain. I'd prefer to keep it here for our own sanity.

I agree that it is weird to build it in the same script we are bumping it. Would `build_linux_bridge.sh` make more sense? I would still call it from this bump script though.

We are consuming U/S components. For some we need to build manifests, for other we need to build whole images.

hack/components/bump_linux_bridge.sh
RamLavi: IIUC there is only 1 image path at a certain point of time, and if it does changes ( like recently changed on multus) then it  should be changed in a proper PR. So I don't see a reason to make it a parameter that is changable from the outside

I mean if it was a common practice to use our own private images then maybe there would be a case for it, but from what I understood so far it's uncommon and not recommended

hack/components/bump-linux-bridge.sh
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/env bash
+
+set -xe
```
oshoval, line 3: others have -o pipefail as well if you want to add here as well

ormergi: I agree, we should keep all bump scripts similar as possible for better maintenance and readability

hack/components/bump_linux_bridge.sh
ormergi: How come there is no built images for CNI plugins like other components?
Should we raise it in containernetworking/plugin repository?

phoracek: Images as a delivery mechanism is a Kubernetes thing. containernetworking/plugins is a generic CNI project that can be used locally with podman etc.

PR #456, opened 2020-07-08
Add multus bump script

RamLavi:
Currently cnao only bumps multus manually.
Let's add a script to do it for us.

**Release note**:

```release-note
NONE
```


RamLavi:
/hold

until #449 is merged

hack/components/bump-multus.sh
```diff
@@ -0,0 +1,26 @@
...
+echo 'Fetch multus sources'
+git-utils::fetch_component ${MULTUS_PATH} ${MULTUS_URL} ${MULTUS_COMMIT}
+
+echo 'Get multus image name and update it under CNAO'
+MULTUS_TAG=$(git-utils::get_component_tag ${MULTUS_PATH})
+MULTUS_IMAGE=nfvpe/multus
+MULTUS_IMAGE_TAGGED=${MULTUS_IMAGE}:${MULTUS_TAG}
+sed -i "s#\"${MULTUS_IMAGE}:.*\"#\"${MULTUS_IMAGE_TAGGED}\"#" \
+    pkg/components/components.go \
+    test/releases/${CNAO_VERSION}.go
```
oshoval, line 26: new line at the end ?

RamLavi:
> /hold
> 
> Can we also fetch Multus manifests?

I think we can use the [image yaml ](https://github.com/intel/multus-cni/blob/8fff5aac6046f4342dc65bda6190ee1172e1d7e7/images/multus-daemonset.yml)with some kustomize/sed tweeks

Though I didn't undersatnd why they have 2 multus daemonsets there

data/multus/001-multus.yaml
```diff
@@ -0,0 +1,181 @@
...
+data:
+  # NOTE: If you'd prefer to manually apply a configuration file, you may create one here.
+  # In the case you'd like to customize the Multus installation, you should change the arguments to the Multus pod
+  # change the "args" line below from
+  # - "--multus-conf-file=auto"
+  # to:
+  # "--multus-conf-file=/tmp/multus-conf/70-multus.conf"
+  # Additionally -- you should ensure that the name "70-multus.conf" is the alphabetically first name in the
+  # /etc/cni/net.d/ directory on each node, otherwise, it will not be used by the Kubelet.
+  cni-conf.json: |
```
phoracek, line 151: We should not use this mode. This assumes you want hard-code config for CNI. What we want to do is to use the dynamic mode. `"--multus-conf-file=auto"`

data/multus/001-multus.yaml
```diff
@@ -0,0 +1,181 @@
+---
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: kube-multus-ds-amd64
```
phoracek, line 5: We currently use `multus` as the name. I would prefer keeping it since it looks so much nicer, but if you decide to just copy what is available from U/S, please add a code to the Multus component to delete old `multus` daemon sets.

hack/components/yaml-utils.sh
```diff
@@ -24,3 +24,11 @@ function yaml-utils::get_component_repo() {
...
 	echo ${url} | sed 's#https://\(.*\)#\1#'
 }
+
+function yaml-utils::split_yaml_by_seperator() {
+	local outout_dir=$1
+	local source_yaml=$2
+
+	cd ${outout_dir}
+	awk '/\-\-\-/{f=NR".yaml"}; {print >f}' ${source_yaml}
+}
```
phoracek, line 34: Please add a new line

hack/components/yaml-utils.sh
```diff
@@ -24,3 +24,11 @@ function yaml-utils::get_component_repo() {
 	#remove the prefix.
 	echo ${url} | sed 's#https://\(.*\)#\1#'
 }
+
+function yaml-utils::split_yaml_by_seperator() {
+	local outout_dir=$1
```
phoracek, line 29: output

hack/components/bump-multus.sh
```diff
@@ -0,0 +1,62 @@
...
+		fi
+	done
+)
+
+echo 'parametize and copy manifests'
+rm -rf data/multus/*
+(
+	cat ${MULTUS_PATH}/config/cnao/001-multus.yaml |
+		sed 's/kube-system/{{ .Namespace }}/' |
+		sed 's/image: .*/image: {{ .MultusImage }}\n        imagePullPolicy: {{ .ImagePullPolicy }}/' |
```
phoracek, line 44: Could we use yq or something proper to parse the YAML and replace values? This would silently break if there is a change in the original YAML.

RamLavi:
/hold 
the multus ds is terminating every ~1 minute.

data/multus/001-multus.yaml
RamLavi: The daemonset object is already set to `--multus-conf-file=auto`, so I will just discard this configMap object.

data/multus/001-rbac.yaml
```diff
@@ -19,23 +19,23 @@ rules:
...
   apiGroup: rbac.authorization.k8s.io
   kind: ClusterRole
   name: multus
 subjects:
-- kind: ServiceAccount
-  name: multus
-  namespace: {{ .Namespace }}
+  - kind: ServiceAccount
+    name: multus
+    namespace: '{{ .Namespace }}'
```
qinqon, line 32: Why do we need single quota, what namespace is failing ?

data/multus/001-rbac.yaml
```diff
@@ -19,23 +19,23 @@ rules:
       - update
 ---
 kind: ClusterRoleBinding
-apiVersion: rbac.authorization.k8s.io/v1beta1
+apiVersion: rbac.authorization.k8s.io/v1
```
qinqon, line 22: This is not related tright ?

hack/components/bump-multus.sh
```diff
@@ -0,0 +1,117 @@
...
+  name: multus
+allowPrivilegedContainer: true
+allowHostDirVolumePlugin: true
+runAsUser:
+  type: RunAsAny
+seLinuxContext:
+  type: RunAsAny
+users:
+- system:serviceaccount:{{ .Namespace }}:multus
+{{ end }}
```
qinqon, line 87: Missing `if` statement, we should test this changes at openshift ASAP.

hack/components/bump-multus.sh
```diff
@@ -0,0 +1,117 @@
...
+---
+EOF
+
+	echo 'rejoin sub-manifests to final manifests'
+	RBAC_YAML=001-rbac.yaml
+	touch ${RBAC_YAML}
```
qinqon, line 93: We don't need to unitfy them, CNAO will works with individual yamls files under data.

hack/components/bump-multus.sh
```diff
@@ -0,0 +1,117 @@
...
+	touch ${DAEMONSET_YAML}
+	cat CustomResourceDefinition_network-attachment-definitions.k8s.cni.cncf.io.yaml >> ${DAEMONSET_YAML} &&
+		cat DaemonSet_kube-multus-ds-amd64.yaml >> ${DAEMONSET_YAML}
+)
+
+echo 'copy manifests'
+rm -rf data/multus/*
+cp ${MULTUS_PATH}/config/cnao/000-ns.yaml data/multus/
+cp ${MULTUS_PATH}/config/cnao/001-rbac.yaml data/multus/
+cp ${MULTUS_PATH}/config/cnao/002-multus.yaml data/multus/
```
qinqon, line 109: As said this this file structure is not needed for example at `kubernetes-nmstate` we follow the file names at knmstate project so it simplifies dumping script, also makes easier to debug problems since mapping manifests from CNAO to component project is easier.

data/multus/001-rbac.yaml
RamLavi: the old meanifest was outdated and static. 
Now that I'm taking the manifest from the tagged version, we get all the new stuff, such as this.

hack/components/bump-multus.sh
RamLavi: yeah but don't we want to gather them to `rbac` and `multus` like it was before?

hack/components/bump-multus.sh
RamLavi: ok, so is my current naming of "kind_name" ok for us?

qinqon: I suppose

data/multus/001-rbac.yaml
RamLavi: since the string I want to change (`{{ .Namespace }}`) has spaces, I need to pass it with `""` all the way to the yq function. This results with the final yq to run like this:
```
__yq w -i ./ServiceAccount_multus.yaml metadata.namespace '{{ .Namespace }}'
```

This is why we get the `''`, but no matter what I tried I couldn't remove the `''` (because like I said, removing the `""` always breaks the argument to words and gives me this `namespace: '{{'`).

Do you have any workaround to this issue?

RamLavi: the only thing I can come up with is to run `sed 's/'{{ .Namespace }}'/{{ .Namespace }}/g'` to the final yaml, or something of that sort, but it's ugly

RamLavi:
/hold

need to resolve the `''` added to the replaced names

RamLavi:
/retest

unrelated flakyness

phoracek:
/hold

There is a real issue:

```
      message: 'could not apply (apps/v1, Kind=DaemonSet) cluster-network-addons/multus:
        could not update object (apps/v1, Kind=DaemonSet) cluster-network-addons/multus:
        DaemonSet.apps "multus" is invalid: spec.selector: Invalid value: v1.LabelSelector{MatchLabels:map[string]string{"name":"multus"},
        MatchExpressions:[]v1.LabelSelectorRequirement(nil)}: field is immutable'
```

RamLavi:
> Seems that the new version of the DaemonSet changes label selectors. We are not allowed to patch that live.

From some reason it doesn't recreate in my local, but I changed it as the error is very specific. Let's see if this solves the issue

PR #465, opened 2020-07-13
Do not trim the version exposed in CR

phoracek:

We started trimming the version so it can fit into labels. We will keep
that but we will stop trimming it for the use in our CRs. That way
whatever was set in CNAO deployment as a version will be exposed in its
CRs. So no unexpected results.

https://bugzilla.redhat.com/show_bug.cgi?id=1856438

**Release note**:

```release-note
Do not trim OPERATOR_VERSION in the CR
```

### DIFF
--- a/pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
+++ b/pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
@@ -36,6 +36,7 @@ import (
 	"github.com/kubevirt/cluster-network-addons-operator/pkg/controller/statusmanager"
 	"github.com/kubevirt/cluster-network-addons-operator/pkg/names"
 	"github.com/kubevirt/cluster-network-addons-operator/pkg/network"
+	"github.com/kubevirt/cluster-network-addons-operator/pkg/util/k8s"
 )
 
 // ManifestPath is the path to the manifest templates
@@ -43,10 +44,12 @@ const ManifestPath = "./data"
 
 var operatorNamespace string
 var operatorVersion string
+var operatorVersionLabel string
 
 func init() {
 	operatorNamespace = os.Getenv("OPERATOR_NAMESPACE")
 	operatorVersion = os.Getenv("OPERATOR_VERSION")
+	operatorVersionLabel = k8s.StringToLabel(operatorVersion)
 }
 
 // Add creates a new NetworkAddonsConfig Controller and adds it to the Manager. The Manager will set fields on the Controller
@@ -305,7 +308,7 @@ func (r *ReconcileNetworkAddonsConfig) renderObjects(networkAddonsConfig *opv1al
 		if labels == nil {
 			labels = map[string]string{}
 		}
-		labels[opv1alpha1.SchemeGroupVersion.Group+"/version"] = operatorVersion
+		labels[opv1alpha1.SchemeGroupVersion.Group+"/version"] = operatorVersionLabel
 		obj.SetLabels(labels)
 	}
 
--- a/pkg/util/k8s/labels.go
+++ b/pkg/util/k8s/labels.go
@@ -1,0 +1,31 @@
+package k8s
+
+import (
+	"log"
+	"regexp"
+)
+
+var labelAntiSelector *regexp.Regexp
+
+func init() {
+	var err error
+
+	// Regex matching all non-allowed label characters
+	labelAntiSelector, err = regexp.Compile("[^a-z0-9A-Z_.-]+")
+	if err != nil {
+		log.Fatal(err)
+	}
+}
+
+// Make given string label-compatible
+func StringToLabel(s string) string {
+	// We need to remove characters that are not allowed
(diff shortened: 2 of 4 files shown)


pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
```diff
@@ -47,6 +48,7 @@ var operatorVersion string
 func init() {
 	operatorNamespace = os.Getenv("OPERATOR_NAMESPACE")
 	operatorVersion = os.Getenv("OPERATOR_VERSION")
+	operatorVersionLabel := k8s.StringToLabel(operatorVersion)
```
nunnatsa, line 51: This is defined in the scope of the init() method. Please define a new var, as done for the other two

pkg/util/k8s/labels_test.go
```diff
@@ -0,0 +1,33 @@
...
+import (
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	"github.com/kubevirt/cluster-network-addons-operator/pkg/util/k8s"
+)
+
+var _ = Describe("StringToLabel", func() {
+	Context("when given a very long string", func() {
+		s := "Avalanches_of water in the midst of a calm/And the distances cataracting toward the abyss!"
```
alonSadan, line 12: Did you mean "Avalanches:of"?, because otherwise I don't think anything will be replaced...

pkg/util/k8s/labels.go
```diff
@@ -0,0 +1,18 @@
+package k8s
+
+import (
+	"strings"
+)
+
+// Make given string label-compatible
+func StringToLabel(s string) string {
+	// We need to remove characters that are not allowed, e.g. :
+	s = strings.Replace(s, ":", "_", -1)
```
alonSadan, line 10: if  `:` is just an example,are there other chars needed to be checked/replaced, or `:` is the only relevant one?

phoracek: There are probably more of those. The usage of this is aimed on version digests being used as the label, so I focused on handling those only instead of making a generic solution fitting all. This is probably worth a doc string. Lemme add it.

alonSadan: got it. Thanks

pkg/util/k8s/labels.go
```diff
@@ -0,0 +1,31 @@
+package k8s
+
+import (
+	"log"
+	"regexp"
+)
+
+var labelAntiselector *regexp.Regexp
```
alonSadan, line 8: maybe `labelAntiSelector` instead of `labelAntiselector` ?

pkg/util/k8s/labels_test.go
```diff
@@ -0,0 +1,46 @@
...
+		It("should trim the text to supported 63 characters", func() {
+			label := k8s.StringToLabel(s)
+			Expect(label).To(HaveLen(63), "The string should be trimmed to 63 characters")
+		})
+	})
+
+	Context("when given a string with special characters", func() {
+		s := "Avalanches&of:water"
+
+		It("should replace special characters with underscrore", func() {
```
alonSadan, line 23: Typo. Should be "underscore"

pkg/util/k8s/labels.go
```diff
@@ -0,0 +1,31 @@
...
+
+func init() {
+	var err error
+
+	// Regex matching all non-allowed label characters
+	labelAntiselector, err = regexp.Compile("[^a-z0-9A-Z_.-]+")
```
oshoval, line 14: why not err := maybe?

oshoval:
Nice,
would suggest 's/do/Do' in commit and description please

pkg/util/k8s/labels.go
nunnatsa: > why not err := maybe?

Won't work - `labelAntiselector` is global, so you can't use `:=`

oshoval: ic thx, is it the favorable method to use globals, or maybe a singleton is preferable ?

nunnatsa: It's actually a kind of singleton, since it's a private variable of a package.

pkg/util/k8s/labels.go
alonSadan, line 8: maybe `labelAntiSelector` instead of `labelAntiselector`

pkg/util/k8s/labels.go
phoracek: It is private and read-only within the package. It is as if it was a constant, except we need to populate it with a function in init.

oshoval: yep thanks, actually its almost the same as the pattern indeed
```
var instantiated *single = nil

func New() *single {
        if instantiated == nil {
                instantiated = new(single);
        }
        return instantiated;
}
```

phoracek:

Flake

phoracek:

Known issue

phoracek:

Flake

PR #469, opened 2020-07-17
Introduce `networkaddonsoperator.network.kubevirt.io/rejectOwner` annotation and drop ownership of KMP secrets

phoracek:

# Introduce rejectOwner annotation

In the current code, we mark all objects created by CNAO (those kept
under data/) as owned by us. As a result, they are garbage collected
upon removal of NetworkAddonsConfig.

This cause complications with objects that we do not want to remove or
that we want to be owned by other objects. The recent example of this
were KubeMacPool secrets that need to be precreated in CNAO in order to
avoid chicken-egg issues, but then handled separately.

In order to provide a generic mechanism to avoid ownership, a new
annotation `networkaddonsoperator.network.kubevirt.io/rejectOwner` was
introduced. When an object is annotated with this, we will not set the
reference.

# Don't set owner reference on KMP secret

Currently, we set owner reference of the secret to NetworkAddonsConfig.
That means that it will be removed upon deletion of that CR.

The issue is that there are KMP controllers recreating this Secret, not
knowing whether the whole setup is being deleted or just the specific
Secret.

In order to make this little clearer, we decided not to own the secret
at all and leave its lifecycle on KMP. We still need to pre-create it
with its dummy values in CNAO to prevent chicken-egg issues.

# Don't set owner reference on knmstate secret

Currently, we set owner reference of the secret to NetworkAddonsConfig.
That means that it will be removed upon deletion of that CR.

The issue is that there are knmstate controllers recreating this Secret, not
knowing whether the whole setup is being deleted or just the specific
Secret.

In order to make this little clearer, we decided not to own the secret
at all and leave its lifecycle on knmstate. We still need to pre-create it
with its dummy values in CNAO to prevent chicken-egg issues.

**Release note**:

```release-note
Introduce `networkaddonsoperator.network.kubevirt.io/rejectOwner` annotation and drop ownership of KMP secrets
```

### DIFF
--- a/hack/components/bump-knmstate.sh
+++ b/hack/components/bump-knmstate.sh
@@ -29,3 +29,10 @@ echo 'Copy kubernetes-nmstate manifests'
 rm -rf data/nmstate/*
 cp $NMSTATE_PATH/deploy/handler/* data/nmstate/
 cp $NMSTATE_PATH/deploy/crds/*nodenetwork*crd* data/nmstate/
+
+echo 'Apply custom CNAO patches on kubernetes-nmstate manifests'
+echo '
+241a242,243
+>   annotations:
+>     networkaddonsoperator.network.kubevirt.io/rejectOwner: ""
+' | patch data/nmstate/operator.yaml
--- a/hack/components/bump-kubemacpool.sh
+++ b/hack/components/bump-kubemacpool.sh
@@ -30,6 +30,7 @@ bases:
 - ../default
 patchesStrategicMerge:
 - cnao_image_patch.yaml
+- cnao_rejectowner_patch.yaml
 EOF
 
     cat <<EOF > config/cnao/cnao_image_patch.yaml
@@ -46,7 +47,18 @@ spec:
         imagePullPolicy: "{{ .ImagePullPolicy }}"
         name: manager
 EOF
+
+    cat <<EOF > config/cnao/cnao_rejectowner_patch.yaml
+apiVersion: v1
+kind: Secret
+metadata:
+  name: service
+  namespace: system
+  annotations:
+    networkaddonsoperator.network.kubevirt.io/rejectOwner: ""
+EOF
 )
+
 rm -rf data/kubemacpool/*
 (
     cd $KUBEMACPOOL_PATH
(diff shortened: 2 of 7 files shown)


hack/components/bump-kubemacpool.sh
```diff
@@ -46,7 +47,18 @@ spec:
...
 EOF
+
+    cat <<EOF > config/cnao/cnao_rejectowner_patch.yaml
+apiVersion: v1
+kind: Secret
+metadata:
+  name: service
+  namespace: system
+  annotations:
+    networkaddonsoperator.network.kubevirt.io/rejectOwner: ""
```
qinqon, line 58: Can we add this annotation only at TLS secret ? 
https://github.com/k8snetworkplumbingwg/kubemacpool/blob/master/config/default/manager/manager.yaml#L201

phoracek: That's what we do, no? The Secret is identified by namespace and name. We can't have two secrets with the same name, so this should work ok.

qinqon: Ahh you right !

qinqon:
/hold

we need to add knmstate secrets too.

qinqon:
@phoracek does not this breaks automatization ? since we are adding a custom annotation that will disappear on `make components`

phoracek:
The patch with the annotation is applied as a part of bump automation.

hack/components/bump-knmstate.sh
```diff
@@ -29,3 +29,10 @@ echo 'Copy kubernetes-nmstate manifests'
 rm -rf data/nmstate/*
 cp $NMSTATE_PATH/deploy/handler/* data/nmstate/
 cp $NMSTATE_PATH/deploy/crds/*nodenetwork*crd* data/nmstate/
+
+echo 'Apply custom CNAO patches on kubernetes-nmstate manifests'
+echo '
+241a242,243
```
oshoval, line 35: would be better to use yq i think?

phoracek: `yq` can't parse the source since it is not a valid YAML, but Go template.

Issue #451, opened 2020-07-06
NetworkAddonsConfig when is the config already deployed should report non-empty list of deployed containers


arnongilboa:
/kind bug
/triage build-officer

**What happened**:

[Fail] NetworkAddonsConfig when is the config already deployed [It] should report non-empty list of deployed containers 
/tmp/cnao/cluster-network-addons-operator/test/e2e/workflow/containers_test.go:27

https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_cluster-network-addons-operator/448/pull-e2e-cluster-network-addons-operator-workflow-k8s/1280102040416030720

PR #474, opened 2020-07-21
Fix test NetworkAddonsConfig shouldn't be empty when config deployed

rhrazdil:
The StatusManager `SetFromPods` function is accesed from both the `pod_controller` and `networkaddonsconfig_controller`.
This PR adds mutex and locks sections that read/write shared variables: containers, deployments, daemonSets.
Also, writing and reading these shared variables is done together atomicallly in single locked sections (`GetAttribures` and `SetAttributes`)

Fix flaky test  https://github.com/kubevirt/cluster-network-addons-operator/issues/451

**Release note**:

```release-note
none
```

### DIFF
--- a/pkg/controller/statusmanager/status_manager.go
+++ b/pkg/controller/statusmanager/status_manager.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"reflect"
 	"strings"
+	"sync"
 	"time"
 
 	conditionsv1 "github.com/openshift/custom-resource-status/conditions/v1"
@@ -54,6 +55,7 @@ type StatusManager struct {
 	deployments []types.NamespacedName
 
 	containers []opv1alpha1.Container
+	mux        sync.Mutex
 }
 
 func New(client client.Client, name string) *StatusManager {
@@ -151,7 +153,7 @@ func (status *StatusManager) set(reachedAvailableLevel bool, conditions ...condi
 	}
 
 	// Make sure to expose deployed containers
-	config.Status.Containers = status.containers
+	_, _, config.Status.Containers = status.GetAttributes()
 
 	// Expose currently handled version
 	config.Status.OperatorVersion = operatorVersion
@@ -212,12 +214,18 @@ func (status *StatusManager) SetNotFailing(level StatusLevel) {
 	status.syncFailing()
 }
 
-func (status *StatusManager) SetDaemonSets(daemonSets []types.NamespacedName) {
+func (status *StatusManager) SetAttributes(daemonSets []types.NamespacedName, deployments []types.NamespacedName, containers []opv1alpha1.Container) {
+	status.mux.Lock()
+	defer status.mux.Unlock()
 	status.daemonSets = daemonSets
+	status.deployments = deployments
+	status.containers = containers
 }
 
-func (status *StatusManager) SetDeployments(deployments []types.NamespacedName) {
-	status.deployments = deployments
+func (status *StatusManager) GetAttributes() ([]types.NamespacedName, []types.NamespacedName, []opv1alpha1.Container) {
+	status.mux.Lock()
+	defer status.mux.Unlock()
+	return status.daemonSets, status.deployments, status.containers
 }
 
 // SetFromPods sets the operator status to Failing, Progressing, or Available, based on
@@ -225,10 +233,11 @@ func (status *StatusManager) SetDeployments(deployments []types.NamespacedName) 
 // no-op if the StatusManager is currently marked as failing due to a configuration error.
 func (status *StatusManager) SetFromPods() {
 	progressing := []string{}
+	daemonSets, deployments, _ := status.GetAttributes()
 
 	// Iterate all owned DaemonSets and check whether they are progressing smoothly or have been
 	// already deployed.
-	for _, dsName := range status.daemonSets {
+	for _, dsName := range daemonSets {
 
 		// First check whether DaemonSet namespace exists
 		ns := &corev1.Namespace{}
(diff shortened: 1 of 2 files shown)


rhrazdil:
@phoracek the test passed, from the log I can see that the test took 6 seconds, compared to 2 seconds in the [failed run](https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_cluster-network-addons-operator/448/pull-e2e-cluster-network-addons-operator-workflow-k8s/1280102040416030720), so the 10 sec limit seems reasonable.

But, but question is, is this the right approach? Or should we focus on the part that populates the list of deployed containers?

phoracek:
@rhrazdil I don't think this is the right approach. In the BeforeEach of the test, we apply the config and then we wait for the Available condition to turn True. By the time we set the condition to True, containers should be already listed. Maybe we need to look into that part of code to see why it is not blocking and why are containers not set before the condition.

RamLavi:
> the test took 6 seconds

Hi @rhrazdil , IIUC the [test](https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_cluster-network-addons-operator/474/pull-e2e-cluster-network-addons-operator-workflow-k8s/1285580824300752896#1:build-log.txt%3A757) that took 6 seconds was not test you're talking about. but the former (notice the ---- separation)
The [test](https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_cluster-network-addons-operator/474/pull-e2e-cluster-network-addons-operator-workflow-k8s/1285580824300752896#1:build-log.txt%3A807) in question took 72 seconds in it's entirety (it is not mentioned how much time each step takes), so it's not comparable to the 2 seconds that took it to fail the the flaky case

rhrazdil:
@RamLavi Ah I see, thanks for pointing this out, the times are indeed not comparable

rhrazdil:
@phoracek @RamLavi  PR is ready for review

pkg/controller/statusmanager/status_manager.go
```diff
@@ -212,23 +214,58 @@ func (status *StatusManager) SetNotFailing(level StatusLevel) {
...
 
+func (status *StatusManager) SetAttributes(daemonSets []types.NamespacedName, deployments []types.NamespacedName, containers []opv1alpha1.Container) {
+	status.mux.Lock()
+	defer status.mux.Unlock()
+	status.daemonSets = daemonSets
+	status.deployments = deployments
+	status.containers = containers
+}
+
 func (status *StatusManager) SetDaemonSets(daemonSets []types.NamespacedName) {
```
phoracek, line 225: We don't need these setters anymore. The only use of them was replaced with SetAttributes. Having one attomic setter and couple of getters looks fine to me.

pkg/controller/statusmanager/status_manager.go
```diff
@@ -335,11 +372,7 @@ func (status *StatusManager) SetFromPods() {
 	status.SetNotFailing(PodDeployment)
 
 	// Finally, if all containers are deployed, mark as Available
-	if len(progressing) == 0 {
+	if len(progressing) == 0 && len(status.GetContainers()) > 0 {
```
phoracek, line 375: I don't think this is valid. Empty configuration is allowed, thus we can be ready in case somebody just creates empty NetworkAddonsConfig CR without any containers listed.

pkg/controller/statusmanager/status_manager.go
```diff
@@ -245,10 +245,12 @@ func (status *StatusManager) GetContainers() []opv1alpha1.Container {
 // no-op if the StatusManager is currently marked as failing due to a configuration error.
 func (status *StatusManager) SetFromPods() {
 	progressing := []string{}
+	daemonSets := status.GetDaemonSets()
```
phoracek, line 248: Don't we need an atomic getter for all 3 of them? Otherwise we may get into a race again.

pkg/controller/statusmanager/status_manager.go
```diff
@@ -354,6 +356,12 @@ func (status *StatusManager) SetFromPods() {
 	// If all pods are being created, mark deployment as not failing
 	status.SetNotFailing(PodDeployment)
 
+	// If there are deployments or daemonSets created, but no pods
```
phoracek, line 359: If we lock the whole getter, we would not need to have this check. Right?

pkg/controller/statusmanager/status_manager.go
rhrazdil: We use the same lock for accessing these, so if there is somebody inside the `SetAttributes`, we'll wait until the function is finished and only then read the value.

pkg/controller/statusmanager/status_manager.go
rhrazdil: If I understand correctly, atomic getter for all 4 attribute would not neccessarily mitigate this.

pkg/controller/statusmanager/status_manager.go
rhrazdil: Yep, I missed the empty spec case. Fixed.

pkg/controller/statusmanager/status_manager.go
rhrazdil: Removed

rhrazdil:
> It seems that the commit contains 3 different topic. Please split it to three different commits to make reviewing easier. Update the PR description as well.

I'm not sure I see the 3rd topic. The `SetAttribute` function seems related to the first one, so I have splitted into 2 commits.

pkg/controller/statusmanager/status_manager.go
phoracek: Following can happen:

1. We get DS
2. Somebody sets new DS and Dep
3. We get Dep

Then we don't have DS and Dep from the same batch.

pkg/controller/statusmanager/status_manager.go
phoracek: If we have atomic getter and setter, then it can never happen that there are any deployments or daemonsets but no containers. So the following condition would be never true.

pkg/controller/statusmanager/status_manager.go
rhrazdil: Ah I see!

rhrazdil: Updated.

pkg/controller/statusmanager/status_manager.go
rhrazdil: Added the atomic getter, removed this condition

PR #479, opened 2020-07-23
Bumper go script infra

RamLavi:
This PR introduces the bumper script with the appropriate function logic flow.
This PR does not include the content of the functions for preparing the PR itself.

- **add bumper script main structure**
Adding bumper script main functions to create the high level logic.
Also add metadata to yaml file, and change params to allow proper parsing of yaml.

- **include bumper folder in fmt,vet,test/unit**
expand fmt, vet, test/unit folders to include the bumper folder

- **adding bumper unit test suite**
adding parser and bumper unit tests.

- **adding README.md for easy usage**
explaining how to use the script locally and via github-Actions

- **add github interface and git interface infra**
adding the github library in a separate interface.
Also adding git clone in order to perform actions on the repo.

- **add components.yaml feilds: update-policy, branch**
expand components.yaml fields to allow CPaaS logic

- **add CPaaS update-policy logic**
introducing 3 methods of bumping: tagged, latest, static.
tagged: updates to the latest tagged commit in the referenced branch
latest: updates to the latest HEAD in the referenced branch
static: disables auto update
Also Adding githubApi fakes and unit tests

- **add isBumpNeeded logic, and unit-test**
Add checks in order to decide if bump is needed for each component

**Release note**:

```release-note
NONE
```


RamLavi:
/hold

still need to add parser unit tests content

automation/bumper-bot/bumper.go
```diff
@@ -0,0 +1,104 @@
+//usr/bin/env go run $0 "$@"; exit
```
phoracek, line 1: We should keep this program under /tool. Automation then should contain any glue needed to be able to run it on an arbitrary CI host.

All the scripts should be available to humans. Automation directory is just to proxy for CI.

tools/bumper-bot/parser/parser.test.go
```diff
@@ -0,0 +1,46 @@
+package parser
+
```
qinqon, line 2: name is usually `parser_test.go` instead of `parser.test.go`

tools/bumper-bot/parser/parser.go
```diff
@@ -0,0 +1,48 @@
+package parser
```
qinqon, line 1: We don't need new package for this this way we don't need to make functions public.

components.yaml
```diff
@@ -1,22 +1,29 @@
 components:
-  multus:
+  - name: "multus"
```
phoracek, line 2: The API needs to remain the same, it is a public API. Is there an issue with adapting the existing structure where key is the name of the component?

components.yaml
```diff
@@ -1,22 +1,29 @@
...
     url: "https://github.com/intel/multus-cni"
-    commit: "bfaf22964b51b93b08dcb1f10cbd8f8cb9a3603a" # v3.4.1
-  linux-bridge:
+    commit: "bfaf22964b51b93b08dcb1f10cbd8f8cb9a3603a"
+    release_tag: "v3.4.1"
```
phoracek, line 5: Thanks for keeping the tag somewhere, I know it can remain as a comment.

Is it a good name though? What if we set our strategy to always take the latest available code on a branch? Maybe a more generic name to keep this metadata would be suitable? It is only for our eyes, not to be processed.

tools/bumper-bot/parser/parser.test.go
```diff
@@ -0,0 +1,46 @@
...
+import (
+	. "github.com/onsi/ginkgo"
+)
+
+var _ = Describe("Testing internal parser functions", func() {
```
phoracek, line 7: drop ` functions`

tools/bumper-bot/parser/parser.test.go
```diff
@@ -0,0 +1,46 @@
...
+	Describe("ParseComponentsYaml function", func() {
+		Context("When yaml file is empty", func() {
```
phoracek, line 9: lower case `w`.

so it would read:

`Testing internal parser ParseComponentsYaml function when yaml file is empty`

tools/bumper-bot/parser/parser.test.go
```diff
@@ -0,0 +1,46 @@
...
+			It("should return an error because there is nothing to validate", func() {
+			})
+		})
+		Context("When yaml file is valid", func() {
+			It("should correctly parse entire file", func() {
+			})
+			It("should correctly parse specific params of specific component", func() {
+			})
+		})
+	})
```
phoracek, line 45: What about tests that Update function really updates?

tools/bumper-bot/parser/parser.go
```diff
@@ -0,0 +1,48 @@
...
+
+type Component struct {
+	Name        string `yaml:"name"`
+	Url         string `yaml:"url"`
+	Commit      string `yaml:"commit"`
+	Release_tag string `yaml:"release_tag"`
+}
+
+type ComponentsConfig struct {
+	Components []Component
```
qinqon, line 20: We are missing `yaml:"components"` here

tools/bumper-bot/parser/parser.go
phoracek, line 1: The module should be just `bumper` there is no bot involved here

tools/bumper-bot/parser/parser.go
```diff
@@ -0,0 +1,48 @@
...
+}
+
+func ParseComponentsYaml(componentsYaml string) ([]Component, error) {
```
qinqon, line 23: Why you don't return ComponentsConfig ?

tools/bumper-bot/parser/parser.go
```diff
@@ -0,0 +1,48 @@
...
+	config := ComponentsConfig{}
+
+	fmt.Println("opening components config", componentsYaml)
+	componentsData, err := ioutil.ReadFile(componentsYaml)
```
qinqon, line 27: Catch error

tools/bumper-bot/parser/parser.go
```diff
@@ -0,0 +1,48 @@
...
+	"github.com/pkg/errors"
+
+	"gopkg.in/yaml.v2"
+)
+
+type Component struct {
+	Name        string `yaml:"name"`
+	Url         string `yaml:"url"`
+	Commit      string `yaml:"commit"`
+	Release_tag string `yaml:"release_tag"`
```
phoracek, line 16: The idiomatic naming would be `ReleaseTag`

tools/bumper-bot/parser/parser.go
```diff
@@ -0,0 +1,48 @@
...
+	fmt.Println("Unmarshaling components config", componentsYaml)
+	err = yaml.Unmarshal(componentsData, &config)
+	if err != nil {
+		return config.Components, errors.Wrapf(err, "failed to Unmarshal %s", componentsYaml)
+	}
+
+	return config.Components, nil
+}
+
+func UpdateComponentsYaml(components []Component) error {
```
qinqon, line 38: Use ComponentsConfig all the way

tools/bumper-bot/parser/parser.go
```diff
@@ -0,0 +1,48 @@
...
+	config := ComponentsConfig{components}
+
+	fmt.Println("Marshaling components config")
+	_, err := yaml.Marshal(config)
+	if err != nil {
+		return errors.Wrap(err, "failed to marshal updated components config")
+	}
+
```
qinqon, line 46: I suppos this is missing writting the file right ?

tools/bumper-bot/parser/parser.go
```diff
@@ -0,0 +1,48 @@
...
+}
+
+type ComponentsConfig struct {
+	Components []Component
+}
+
+func ParseComponentsYaml(componentsYaml string) ([]Component, error) {
+	config := ComponentsConfig{}
+
+	fmt.Println("opening components config", componentsYaml)
```
phoracek, line 26: Capital `O`

tools/bumper-bot/parser/parser.go
phoracek, line 23: componentsConfigPath

It would make it clearer that it is a path to the file. Using Config instead of Yaml would put aside the format of the file which is not too relevant for the caller.

tools/bumper-bot/parser/parser.test.go
```diff
@@ -0,0 +1,46 @@
...
+			It("should correctly specific params of specific component", func() {
+			})
+		})
+	})
+	Describe("UpdateComponentsYaml function", func() {
+		Context("When yaml file is empty", func() {
+			Context("and we try to write valid data to it", func() {
+				It("should NOT return an error because data is valid", func() {
+				})
+			})
```
qinqon, line 29: This is nice, one stuff that I usually do is the use `Fail("not implemented")` At all the `It` so you don't forget any of them and prevent false possitive.

tools/bumper-bot/bumper.go
```diff
@@ -0,0 +1,104 @@
...
+)
+
+func main() {
+	if len(os.Args) != 2 {
+		fmt.Println("Error: must contain path to components yaml", os.Args)
+		return
+	}
+	componentsYaml := os.Args[1]
+
+	fmt.Println("parsing", componentsYaml)
```
phoracek, line 22: Capital P

tools/bumper-bot/bumper.go
```diff
@@ -0,0 +1,104 @@
...
+		}
+	}
+}
+
+func printCurrentComponentParams(component parser.Component) {
+	fmt.Println("current component config:")
+	fmt.Println("Name:", component.Name)
+	fmt.Println("Url:", component.Url)
+	fmt.Println("Commit:", component.Commit)
+	fmt.Println("Release Tag:", component.Release_tag)
```
qinqon, line 91: You can encode it to json and print the json, that's why you don't forget any param.

tools/bumper-bot/bumper.go
```diff
@@ -0,0 +1,104 @@
...
+	"strings"
+
+	"github.com/pkg/errors"
+
+	"github.com/kubevirt/cluster-network-addons-operator/tools/bumper-bot/parser"
+)
+
+func main() {
+	if len(os.Args) != 2 {
+		fmt.Println("Error: must contain path to components yaml", os.Args)
```
phoracek, line 17: It should exit with non-0 return code then, not just return.

Could we use `log` instead? I think it is more suitable to describe the progress of execution. Print is more suitable if you want to return some value to the caller.

Using log would also allow you to call `log.Fatal` here

tools/bumper-bot/bumper.go
```diff
@@ -0,0 +1,104 @@
...
+			return
+		}
+
+		latestReleaseVersion, latestReleaseCommit, err := getLatestReleaseInfo(component.Url)
+		if err != nil {
+			fmt.Println(errors.Wrapf(err, "failed to get latest release version tag from %s", component.Name))
+			return
+		}
+
+		//TODO make updatePolicyDummy dynamic
```
phoracek, line 43: What does this mean? Would be this comment useful to anybody but you?

components.yaml
```diff
@@ -1,22 +1,29 @@
...
+    release_tag: "v0.8.6"
+  - name: "bridge-marker"
     url: "https://github.com/kubevirt/bridge-marker"
-    commit: "049eb6a796b742767a0be838f6ddb7f79c97dd60" # 0.3.0
-  kubemacpool:
+    commit: "049eb6a796b742767a0be838f6ddb7f79c97dd60"
+    release_tag: "0.3.0"
+  - name: "kubemacpool"
     url: "https://github.com/k8snetworkplumbingwg/kubemacpool"
-    commit: "7728339b6c8dcfc9d53cc18ce8e06a910094ed47" # v0.16.0
```
qinqon, line 13: Do we really need this ? it add extra changes here and on the bash scripts

tools/bumper-bot/bumper.go
```diff
@@ -0,0 +1,104 @@
...
+	componentsYaml := os.Args[1]
+
+	fmt.Println("parsing", componentsYaml)
+	components, err := parser.ParseComponentsYaml(componentsYaml)
+	if err != nil {
+		fmt.Println(errors.Wrap(err, "failed to parse components yaml"))
+		return
+	}
+
+	for i, component := range components {
```
qinqon, line 29: You can iterate maps with golang no need to transform components into slice.

components.yaml
RamLavi: The reason of the change is mainly because if I parse in the original format then I loose the component name, since it doesn't have a value.
Also, It was more elegent to go over the components as an array of the same struct (that also holds the name).

phoracek: You can treat `components` as a `map[string]component` and be able to keep this. Would that help?

phoracek: Having the name as a part of an item in a list is more Kubernetes way of doing API. However, this API is already consumed externally, let's not break it.

components.yaml
RamLavi: I agree. I think is good for humans to look at the yaml and know exactly what it holds. The only way I can modify it properly it is to add it as part of the struct.
Will change to Metadata.

tools/bumper-bot/parser/parser.go
RamLavi: I prefer this too but it breaks the parsing of the yaml. First char must be upper case and all the rest needs to be lower.

tools/bumper-bot/bumper.go
RamLavi: This is just a placeholder that be be changed in future task

phoracek: Then it should be verbose enough that it would be clear what needs to be done on this place. Otherwise it is just a headscratching comment.

tools/bumper-bot/bumper.go
```diff
@@ -0,0 +1,104 @@
+//usr/bin/env go run $0 "$@"; exit
```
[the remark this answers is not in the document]
RamLavi: yes. there are a couple of ways  to run go as a script. this is  the equivalent of the bash shebang, I found it  to be the nicest. 
more [here](https://gist.github.com/posener/73ffd326d88483df6b1cb66e8ed1e0bd).

tools/bumper-bot/parser/parser.go
RamLavi: ok I'll move it to the main package

tools/bumper-bot/parser/parser.go
RamLavi: IIUC, it doesn't require a yaml prefix since it only holds a struct, only leaf params (with values) get the "yaml" thing

tools/bumper-bot/parser/parser.go
RamLavi: I preferred returning a component array.

tools/bumper-bot/bumper.go
RamLavi: mmm, ComponentsConfig is not a map, it's a struct that holds structs. I'm thinking of using [this](https://stackoverflow.com/questions/18926303/iterate-through-the-fields-of-a-struct-in-go) to iterate.

tools/bumper-bot/bumper.go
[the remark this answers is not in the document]
RamLavi: scratch that. After I braught parser.go to the main package, I must do a `go build` so shebang won't work anymore..

components.yaml
RamLavi: please note that  I'll have to change the parameters names, because for example `bridge-marker` does not unmarshal well, so I'll use `bridgemarker`. 
So I think will still need to do some adjustments to the midstream script.

tools/bumper-bot/parser/parser.go
RamLavi: eventually turned it to a map[string]component

components.yaml
RamLavi: at the very least we need to change components names so thary they don't have `-` (cannot parse the yaml otherwise)

phoracek: Can we keep the name in quotes to prevent that issue? It is not as elegant but would allow us to move forward without breaking existing code.

components.yaml
phoracek: Python yaml library was able to to cope with that 😬 could we keep the name in quotes then?

components.yaml
RamLavi: do you mean like this?
```
  "macvtap-cni":
    url: "https://github.com/kubevirt/macvtap-cni"
    commit: "14e5a3c7bf516bd4542366a78b7af02f04230ac5"
```

phoracek: You can specify in the structure attribute what is the name of it's name of the equivalent in yaml/json. Doesn't that work? And even then, there is still option to treat this as a map.

components.yaml
RamLavi: yay found how to make it work with the original keys.

tools/bumper-bot/parser/parser.test.go
RamLavi: added

tools/bumper-bot/parser/parser.go
RamLavi: I now return map, so it's a totally different struct.

components.yaml
```diff
@@ -1,22 +1,36 @@
 components:
   multus:
     url: "https://github.com/intel/multus-cni"
-    commit: "bfaf22964b51b93b08dcb1f10cbd8f8cb9a3603a" # v3.4.1
+    commit: "bfaf22964b51b93b08dcb1f10cbd8f8cb9a3603a"
+    metadata:
```
qinqon, line 5: Is the `metadata` level needed ?

tools/bumper/bumper.go
```diff
@@ -0,0 +1,139 @@
...
+	"fmt"
+	"log"
+	"os"
+	"os/exec"
+	"strings"
+
+	"github.com/pkg/errors"
+)
+
+var Logger *log.Logger
```
qinqon, line 15: s/Logger/logger/ It can be confuse with a Logger struct

tools/bumper/bumper.go
```diff
@@ -0,0 +1,139 @@
...
+
+type inputParams struct {
+	componentsConfigPath string
+	gitToken string
+}
+
+func main() {
+	var inputArgs inputParams
+	Logger = initBumper(&inputArgs)
```
qinqon, line 24: The `initBumper` means nothing and you are not going to re-use this, don't add not needed indirections, just put here the logger initialization and call the initFlags

tools/bumper/bumper.go
```diff
@@ -0,0 +1,139 @@
...
+			Logger.Printf("Opening new Branch %s", BranchName)
+
+			//push branch to PR
+
+		}
+	}
+}
+
+func getLatestReleaseInfo(componentUrl string) (string, string, error) {
+	return "", "", nil
```
qinqon, line 91: Return an error with "not implemented" to be sure that fails until we don't finish this PR and we don't forget stuff

tools/bumper/bumper.go
```diff
@@ -0,0 +1,139 @@
...
+}
+
+func initBumper(inputParams *inputParams) (*log.Logger) {
+	Logger = initLog()
+	// cold start message (visibly unique in log)
+	Logger.Printf("~~~~~~~~Init Bumper Script~~~~~~~~")
+	initFlags(inputParams)
+
+	return Logger
+}
```
qinqon, line 110: As say before `initBumper` means nothing, do the logger init at main directly and call initFlags after it, also this functions has as mix of responsabilities that opens the door to turn into a pandora box.

tools/bumper/bumper.go
```diff
@@ -0,0 +1,139 @@
...
+
+func initLog() *log.Logger {
+	var buf bytes.Buffer
+	logger := log.New(&buf, "INFO: ", log.LstdFlags)
+
+	file, err := os.OpenFile("bumper.logs", os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0666)
+	if err != nil {
+		exitWithError(errors.Wrap(err, "failed to open log file bumper.logs"))
+	}
+	logger.SetOutput(file)
```
qinqon, line 120: Don't bump to file, just to stdout/stderr if needed we can bump to file at bump script

tools/bumper/bumper.go
```diff
@@ -0,0 +1,139 @@
...
+	flag.StringVar(&paramArgs.componentsConfigPath, "config-path", "", "Full path to components yaml")
+	flag.StringVar(&paramArgs.gitToken, "token", "", "git Token")
+	flag.Parse()
+	if flag.NFlag() != 2 {
+		exitWithError(fmt.Errorf("Wrong Number of input parameters %d, should be 2. Use --help for usage", flag.NFlag()))
+	}
+}
+
+func exitWithError(err error) {
+	fmt.Fprintf(os.Stderr, "Exiting with Error: %s\n", err)
```
qinqon, line 135: We don't need this is we are doing the logger

tools/bumper/parser.go
```diff
@@ -0,0 +1,114 @@
...
+	Tag		string `yaml:"tag"`
+}
+
+type component struct {
+	Url         string `yaml:"url"`
+	Commit      string `yaml:"commit"`
+	Metadata 	metadata
+}
+
+type ComponentsList struct {
```
qinqon, line 25: Call this `Components` or `ComponentList`

tools/bumper/parser.go
```diff
@@ -0,0 +1,114 @@
...
+)
+
+type metadata struct {
+	Tag		string `yaml:"tag"`
+}
+
+type component struct {
+	Url         string `yaml:"url"`
+	Commit      string `yaml:"commit"`
+	Metadata 	metadata
```
qinqon, line 22: Missing 'yaml:"metadata"' or we can even remove metadata totally and put tag here let's do it as flat as possible.

tools/bumper/parser.go
```diff
@@ -0,0 +1,114 @@
...
+	BridgeMarker component `yaml:"bridge-marker"`
+	Kubemacpool component `yaml:"kubemacpool"`
+	Nmstate component `yaml:"nmstate"`
+	OvsCni component `yaml:"ovs-cni"`
+	MacvtapCni component `yaml:"macvtap-cni"`
+}
+
+type ComponentsConfig struct {
+	Components ComponentsList
+}
```
qinqon, line 37: We don't need this to be public since now it's next to main at the same package
```golang
type componentsConfig struct {
	components Components
}
```

tools/bumper/parser.go
```diff
@@ -0,0 +1,114 @@
...
+		return nil, errors.Wrapf(err, "Failed to open file %s", componentsConfigPath)
+	}
+
+	Logger.Printf("Unmarshaling components config %s", componentsConfigPath)
+	err = yaml.Unmarshal(componentsData, &config)
+	if err != nil {
+		return nil, errors.Wrapf(err, "Failed to Unmarshal %s", componentsConfigPath)
+	}
+
+	if config == (ComponentsConfig{}) {
```
qinqon, line 54: This covers `len(config.components) == 0` ?

tools/bumper/parser.go
```diff
@@ -0,0 +1,114 @@
...
+}
+
+type ComponentsList struct {
+	Multus component `yaml:"multus"`
+	LinuxBridge component `yaml:"linux-bridge"`
+	BridgeMarker component `yaml:"bridge-marker"`
+	Kubemacpool component `yaml:"kubemacpool"`
+	Nmstate component `yaml:"nmstate"`
+	OvsCni component `yaml:"ovs-cni"`
+	MacvtapCni component `yaml:"macvtap-cni"`
```
qinqon, line 32: Why this is not `map[string]component` ?

tools/bumper/parser.go
```diff
@@ -0,0 +1,114 @@
...
+	v := reflect.ValueOf(components)
+
+	componentsMap := make(map[string]component)
+
+	for i := 0; i < v.NumField(); i++ {
+		componentsMap[v.Type().Field(i).Name] = v.Field(i).Interface().(component)
+	}
+
+	return componentsMap
+}
```
qinqon, line 71: if we use a `map` directly at golang struct we don't need this.

tools/bumper/parser.go
```diff
@@ -0,0 +1,114 @@
...
+func printCurrentComponentParams(component component) error {
+	componentJson, err := json.Marshal(component)
+	if err != nil {
+		return errors.Wrap(err, "Failed to print component params")
+	}
+
+	var prettyJSON bytes.Buffer
+	err = json.Indent(&prettyJSON, componentJson, "", "\t")
+	Logger.Printf("Current component config:\n%s", string(prettyJSON.Bytes()))
+	return nil
```
qinqon, line 112: You can use directly https://golang.org/pkg/encoding/json/#MarshalIndent it do Marshall + Indent

tools/bumper/parser_test.go
```diff
@@ -0,0 +1,185 @@
...
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("Testing internal parser", func() {
+
+	var tmpfile *os.File
+
+	CreateTempFileWithContent := func (content []byte) {
```
qinqon, line 16: Don't use mayus is not needed `createTempFileWithContent` is enough

tools/bumper/parser_test.go
```diff
@@ -0,0 +1,185 @@
...
+		var err error
+		tmpfile, err = ioutil.TempFile("", "test")
+		Expect(err).ToNot(HaveOccurred(), "Should successfully create temp file")
+
+		By("Writing and closing the temporary file")
+		_, err = tmpfile.Write(content)
+		Expect(err).ToNot(HaveOccurred(), "Should successfully write content to the temp file")
+
+		err = tmpfile.Close()
```
qinqon, line 25: This has to be done at `defer` or `AfterEach` somehow

components.yaml
RamLavi: We currently have this `# v3.4.1` comment. 
I want to keep it up to date, but in order to change it, it needs to have a proper field in the yaml.
@phoracek noted that since it is not really used, but only for human eyes to understand exactly what version is used, he [suggested](https://github.com/kubevirt/cluster-network-addons-operator/pull/479#discussion_r459917991) I change it to make it clearer it's only metadata. 
What would you recommend as the best way to keep this?

tools/bumper/bumper.go
RamLavi: Done

tools/bumper/bumper.go
RamLavi: Done

tools/bumper/bumper.go
RamLavi: removed `initBumper`
Done

tools/bumper/parser_test.go
RamLavi: `Close` only closed the file descriptor, I already have a `Remove` in AfterEach. is it ok?

tools/bumper/bumper.go
RamLavi: moved logger to stdout, so this is redundant
Done

tools/bumper/bumper.go
RamLavi: Done

tools/bumper/bumper.go
RamLavi: Done

tools/bumper/parser.go
RamLavi: Done

tools/bumper/parser_test.go
RamLavi: Done

tools/bumper/parser_test.go
RamLavi: Done

tools/bumper/parser.go
RamLavi: Done

tools/bumper/parser.go
RamLavi: Done

tools/bumper/parser.go
RamLavi: Done

tools/bumper/parser.go
RamLavi: Done

tools/bumper/parser.go
RamLavi: Done

tools/bumper/parser.go
RamLavi: change to be explicit. Done

phoracek:
/hold

Wait for me please :eyes:

components.yaml
qinqon: Is the bumping bot going to change that ? if this is the case this is not metadata.

tools/bumper/bumper.go
```diff
@@ -0,0 +1,126 @@
...
+		//TODO updatePolicyDummy is a placeholder. In future task will be replaced by proper parameter set in the configYaml
+		updatePolicyDummy := "static"
+		bumpNeeded, err := isBumpNeeded(currentReleaseVersion, latestReleaseVersion, updatePolicyDummy)
+		if err != nil {
+			exitWithError(errors.Wrapf(err, "Failed to discover if Bump need for %s", componentName))
+		}
+
+		if bumpNeeded {
+			logger.Printf("Bumping %s from %s to %s", componentName, currentReleaseVersion, latestReleaseVersion)
+			//reset --hard git repo
```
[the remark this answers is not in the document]
RamLavi: sure, I was just trying to avoid a big PR, After you lgtm I'll move the Jira task to done and continue on this PR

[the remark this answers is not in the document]
phoracek: If we would make something useful or at least testable, it can be merged. I'm not sure how to go about this

tools/bumper/README.md
```diff
@@ -0,0 +1,11 @@
+# Bumper go Script
+
+The Bumper script goes over all of CNAO's components, using the components.yaml config, finds new releases and bumps them in separate PRs. The script can be run locally or via an automation such as git Actions.
```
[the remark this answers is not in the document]
RamLavi: Done

tools/bumper/README.md
```diff
@@ -0,0 +1,11 @@
+# Bumper go Script
```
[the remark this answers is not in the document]
RamLavi: I think I got over excited about doing it in go :D
Done

tools/bumper/README.md
```diff
@@ -0,0 +1,11 @@
...
+
+The Bumper script goes over all of CNAO's components, using the components.yaml config, finds new releases and bumps them in separate PRs. The script can be run locally or via an automation such as git Actions.
+
+## How to run the script
+
+```
+	go build $GOPATH/src/github.com/kubevirt/cluster-network-addons-operator/tools/bumper
```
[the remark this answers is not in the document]
RamLavi: Done

tools/bumper/README.md
```diff
@@ -0,0 +1,11 @@
...
+	./bumper -config-path=<path-to-components.yaml> -token=<git-token>
+
```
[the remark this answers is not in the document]
RamLavi: Done

tools/bumper/bumper.go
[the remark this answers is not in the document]
RamLavi: ack

tools/bumper/parser.go
```diff
@@ -0,0 +1,72 @@
...
+
+func printCurrentComponentParams(component component) error {
+	componentPrettyJson, err := json.MarshalIndent(component, "", "\t")
+	if err != nil {
+		return errors.Wrap(err, "Failed to print component params")
+	}
+
+	logger.Printf("Current component config:\n%s", string(componentPrettyJson))
+	return nil
+
```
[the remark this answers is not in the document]
RamLavi: Done

tools/bumper/parser_suite_test.go
```diff
@@ -0,0 +1,17 @@
+package main
```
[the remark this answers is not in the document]
RamLavi: Done

tools/bumper/parser_suite_test.go
```diff
@@ -0,0 +1,17 @@
...
+import (
+	"testing"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+func TestNetwork(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "parser Suite")
```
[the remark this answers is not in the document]
RamLavi: Done

tools/bumper/parser_suite_test.go
```diff
@@ -0,0 +1,17 @@
...
+}
+
+var _ = BeforeSuite(func() {
+	logger = initLog()
```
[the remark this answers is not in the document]
RamLavi: not sure I understand. 
`var logger *log.Logger` is defined in the main bumper.go script.
defining it also here will be redundant.

tools/bumper/parser_test.go
```diff
@@ -0,0 +1,183 @@
...
+	"fmt"
+	"io/ioutil"
+	"os"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("Testing internal parser", func() {
+
```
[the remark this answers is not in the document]
RamLavi: Done

tools/bumper/parser_test.go
```diff
@@ -0,0 +1,183 @@
...
+		tmpfile, err = ioutil.TempFile("", "test")
+		Expect(err).ToNot(HaveOccurred(), "Should successfully create temp file")
+
+		defer tmpfile.Close()
+
+		By("Writing and closing the temporary file")
+		_, err = tmpfile.Write(content)
+		Expect(err).ToNot(HaveOccurred(), "Should successfully write content to the temp file")
+	}
+	AfterEach(func() {
```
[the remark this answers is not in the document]
RamLavi: Done

tools/bumper/parser_test.go
```diff
@@ -0,0 +1,183 @@
...
+				Expect(componentsConfig.Components["ovs-cni"].Metadata.Tag).To(Equal("v0.12.0"))
+
+				By("checking MacvtapCni params are valid")
+				Expect(componentsConfig.Components["macvtap-cni"].Url).To(Equal("https://github.com/kubevirt/macvtap-cni"))
+				Expect(componentsConfig.Components["macvtap-cni"].Commit).To(Equal("14e5a3c7bf516bd4542366a78b7af02f04230ac5"))
+				Expect(componentsConfig.Components["macvtap-cni"].Metadata.Tag).To(Equal("v0.2.0"))
+			})
+		})
+	})
+	Describe("UpdateComponentsYaml function", func() {
```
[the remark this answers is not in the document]
RamLavi: Done

tools/bumper/parser_test.go
```diff
@@ -0,0 +1,183 @@
...
+		Context("when yaml file is valid", func() {
+			var componentsConfig *ComponentsConfig
+			var err error
+			BeforeEach(func() {
```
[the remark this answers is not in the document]
RamLavi: Done

tools/bumper/parser_test.go
```diff
@@ -0,0 +1,183 @@
...
+				By("Creating a temporary file with empty content")
+				createTempFileWithContent(componentsDummyYaml)
+
+				By("Parsing the yaml")
+				componentsConfig, err = parseComponentsYaml(tmpfile.Name())
+				Expect(err).ToNot(HaveOccurred(), "Should Succeed to parse file")
+			})
+
+			It("should correctly update file", func() {
+				By("changing some params in the yaml")
```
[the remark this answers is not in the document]
RamLavi: Done

tools/bumper/parser_test.go
```diff
@@ -0,0 +1,183 @@
...
+				newComponent := component{Url: "www.google.com", Commit: "12345", Metadata: metadata{Tag: "v1.2.3"}}
+				componentsConfig.Components["MacvtapCni"] = newComponent
+
+				err := updateComponentsYaml(tmpfile.Name(), componentsConfig)
+				Expect(err).ToNot(HaveOccurred(), "Should Succeed to updating the file")
+
+				By("checking params updated in the yaml")
```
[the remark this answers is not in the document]
RamLavi: Done

tools/bumper/parser_test.go
```diff
@@ -0,0 +1,183 @@
...
+				newcomponentsConfig, err := parseComponentsYaml(tmpfile.Name())
+
+				By("checking MacvtapCni params are valid")
```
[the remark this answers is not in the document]
RamLavi: Done

tools/bumper/bumper.go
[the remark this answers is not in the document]
RamLavi: Should I squash the infra commits after you lgtm?

RamLavi:
@phoracek review [fixes](https://github.com/kubevirt/cluster-network-addons-operator/compare/89b24780a3187a5a2cae585060d5c4391b5b0a10..671a3191b35212d03debb275a3f511c22890950e)

tools/bumper/parser_test.go
```diff
@@ -0,0 +1,185 @@
...
+				createTempFileWithContent(componentsDummyYaml)
+			})
+			It("should correctly parse params of all components", func() {
+				By(fmt.Sprintf("Parsing content of temporary file %s", tmpfile.Name()))
+				componentsConfig, err := parseComponentsYaml(tmpfile.Name())
+				Expect(err).ToNot(HaveOccurred(), "Should Succeed to parse file")
+
+				By("Checking map is not empty")
+				Expect(componentsConfig.Components).ToNot(BeEmpty())
+
```
qinqon, line 114: Gomega do a DeepEquals so is safe to compare the maps meaning that you can just compare input and output config instead of "unrolling" it, although maybe is not as clear at error we can fix that doing at gomega custom matcher for Component type.

tools/bumper/parser_test.go
```diff
@@ -0,0 +1,185 @@
...
+				createTempFileWithContent(componentsDummyYaml)
+
+				By("Parsing the yaml")
+				componentsConfig, err = parseComponentsYaml(tmpfile.Name())
+				Expect(err).ToNot(HaveOccurred(), "Should Succeed to parse file")
+			})
+
+			It("should correctly update file", func() {
+				By("Changing some params in the yaml")
+				newComponent := component{Url: "www.google.com", Commit: "12345", Metadata: metadata{Tag: "v1.2.3"}}
```
qinqon, line 169: Use fields value that make sense, it's better for testing purposes.

tools/bumper/bumper.go
```diff
@@ -0,0 +1,126 @@
...
+
+			//PR name
+			PRTitle := fmt.Sprintf("Bump %s to %s", componentName, latestReleaseVersion)
+			logger.Printf("PR title %s", PRTitle)
```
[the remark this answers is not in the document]
RamLavi: Done

tools/bumper/bumper.go
```diff
@@ -0,0 +1,126 @@
...
+			if out, err := cmd.CombinedOutput(); err != nil {
+				exitWithError(errors.Wrapf(err, "Failed to run bump script. StdOut = %s", string(out)))
+			}
+
+			//create a new branch name
+			BranchName := strings.Replace(PRTitle, " ", "_", -1)
+			logger.Printf("Opening new Branch %s", BranchName)
+
+			//push branch to PR
+
```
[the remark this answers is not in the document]
RamLavi: added `exitWithError()` with "X not implemented" message

tools/bumper/bumper.go
```diff
@@ -0,0 +1,126 @@
...
+			}
+
+			//run bump script
+			cmd := exec.Command("make", fmt.Sprintf("bump-%s", componentName))
+			if out, err := cmd.CombinedOutput(); err != nil {
+				exitWithError(errors.Wrapf(err, "Failed to run bump script. StdOut = %s", string(out)))
+			}
+
+			//create a new branch name
+			BranchName := strings.Replace(PRTitle, " ", "_", -1)
```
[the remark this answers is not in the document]
RamLavi: Done

tools/bumper/parser_test.go
```diff
@@ -0,0 +1,185 @@
...
+ Currepted: "data""
+    Shouldnt': "work"
+`)
+
+	var emptyYaml = []byte(``)
+
+	Describe("ParseComponentsYaml function", func() {
+		Context("when yaml file is empty", func() {
+			BeforeEach(func() {
+				By("Creating a temporary file with empty content")
```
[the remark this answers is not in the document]
RamLavi: but it's different every time ("with empty content", "with corrupted content", etc.)

tools/bumper/parser_test.go
RamLavi: Not sure what you mean here.
I don't have anything ready to compare componentsConfig to.
Do you mean I should create a static componentsConfig with the correct data and just do `if componentsConfig == ExpectedcomponentsConfig`?

tools/bumper/parser_test.go
RamLavi: ok, I'll also add another test for adding a new component

RamLavi:
@qinqon review [fixes](https://github.com/kubevirt/cluster-network-addons-operator/compare/671a3191b35212d03debb275a3f511c22890950e..1ad7a496f196ab75472f2c7c7f4ea4baf87e7c09)

components.yaml
RamLavi: it will change it but only for visibility.  no one actually uses this parameter.

tools/bumper/parser.go
RamLavi: Done

components.yaml
qinqon: Removing complexity is always good I see it just as another tag flating this will help and we are not going to need to filter just by `metadata` fields.

tools/bumper/parser_test.go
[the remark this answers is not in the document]
qinqon: fair enough

tools/bumper/parser_test.go
qinqon: Github is doing tricks, this comments was not done here...

tools/bumper/parser_test.go
```diff
@@ -0,0 +1,208 @@
...
+
+		By("Checking Nmstate params are valid")
+		Expect(componentsConfig.Components["nmstate"].Url).To(Equal("https://github.com/nmstate/kubernetes-nmstate"))
+		Expect(componentsConfig.Components["nmstate"].Commit).To(Equal("6f41747788235c3363a314dfae46a4ed90e1846a"))
+		Expect(componentsConfig.Components["nmstate"].Metadata.Tag).To(Equal("v0.25.0"))
+
+		By("Checking OvsCni params are valid")
+		Expect(componentsConfig.Components["ovs-cni"].Url).To(Equal("https://github.com/kubevirt/ovs-cni"))
+		Expect(componentsConfig.Components["ovs-cni"].Commit).To(Equal("edb4754d08f49be54c399c938895fe82aab6aa5a"))
+		Expect(componentsConfig.Components["ovs-cni"].Metadata.Tag).To(Equal("v0.12.0"))
```
qinqon, line 62: Gomega do a DeepEquals so is safe to compare the maps meaning that you can just compare input and output config instead of "unrolling" it, although maybe is not as clear at error we can fix that doing at gomega custom matcher for Component type.

components.yaml
RamLavi: following conversation, will change to a flat `metadata`

tools/bumper/parser_test.go
RamLavi: Done

tools/bumper/parser_test.go
RamLavi: Changed to using an "expected" static struct. there is no need for a matcher imo the output is understandable
Done

RamLavi:
@qinqon see review [fixes](https://github.com/kubevirt/cluster-network-addons-operator/compare/1ad7a496f196ab75472f2c7c7f4ea4baf87e7c09..72338e0e586d384ab245c1c441b394b8259d2b0a)

tools/bumper/parser_test.go
```diff
@@ -0,0 +1,198 @@
...
+				err := updateComponentsYaml(tmpfile.Name(), componentsConfig)
+				Expect(err).ToNot(HaveOccurred(), "Should Succeed to updating the file")
+
+				By("Checking params updated in the yaml")
+				newcomponentsConfig, err := parseComponentsYaml(tmpfile.Name())
+
+				By("Checking MacvtapCni params are valid")
+				Expect(newcomponentsConfig.Components["MacvtapCni"].Url).To(Equal("https://github.com/qinqon/kube-admission-webhook"))
+				Expect(newcomponentsConfig.Components["MacvtapCni"].Commit).To(Equal("f8f7795d05781cdaf869c301e4736adc29a0de28"))
+				Expect(newcomponentsConfig.Components["MacvtapCni"].Metadata).To(Equal("v0.11.0"))
```
qinqon, line 175: Create a component and compare it to reduce the number of Expects.

tools/bumper/parser.go
```diff
@@ -0,0 +1,67 @@
...
+	Url      string   `yaml:"url"`
+	Commit   string   `yaml:"commit"`
+	Metadata string   `yaml:"metadata"`
+}
+
+type componentsConfig struct {
+	Components map[string]component `yaml:"components"`
+}
+
+func parseComponentsYaml(componentsConfigPath string) (*componentsConfig, error) {
```
qinqon, line 22: Don't return a pointer here it's not optimizing anything

RamLavi: I return a pointer mainly because I get a pointer from the allocation `new(componentsConfig)`.
plus it keeps it similar to it's opposite function `updateComponentsYaml` which also gets a pointer:
```
func updateComponentsYaml(componentsConfigPath string, config *componentsConfig) error {
```

tools/bumper/parser_test.go
RamLavi: Done

RamLavi:
@qinqon review [fixes](https://github.com/kubevirt/cluster-network-addons-operator/compare/72338e0e586d384ab245c1c441b394b8259d2b0a..0c0f0ce6b5ef2bc4a66b57811853183aee4aa735)

RamLavi:
@qinqon fix [reviews](https://github.com/kubevirt/cluster-network-addons-operator/compare/0c0f0ce6b5ef2bc4a66b57811853183aee4aa735..d8964bfdd1c2dfd157f8fddc1783a5e99bdd71c4)

tools/bumper/parser.go
qinqon: But why do we need a pointer ? why do we use `new` ?
you can do
```golang
config := componentsConfig{}
yaml.Unmarshal(componentsData, &config)
```

qinqon:
/lgtm
/hold cancel

Has functionality to be implemented by skeleton looks good.

tools/bumper/git.go
```diff
@@ -50,10 +50,126 @@ func (mgr *gitHubManager) verifyClient() error {
...
+
+func getComponentNameFromUrl(componentUrl string) string {
+	urlSlice := strings.Split(componentUrl, "/")
+	return urlSlice[len(urlSlice)-1]
+}
+
+func getComponentOwnerFromUrl(componentUrl string) string {
+	urlSlice := strings.Split(componentUrl, "/")
+	return urlSlice[len(urlSlice)-2]
+}
```
qinqon, line 175: This is not part of git I would say that we have a component.go and we make this pair of functions with component as receiver

tools/bumper/git.go
```diff
@@ -50,10 +50,126 @@ func (mgr *gitHubManager) verifyClient() error {
...
+
+// getting virtual tag of non tagged commits is not implemented in either github/git libraries, so we use console
+func getVirtualTag(updatedCommit, repoDir string) (string, error) {
+	cmd := exec.Command("git", "-C", repoDir, "describe", updatedCommit, "--tags")
+	out, err := cmd.CombinedOutput()
+	if err != nil {
+		return "", errors.Wrapf(err, "Failed to run git describe. StdOut = %s", string(out))
+	}
+	return strings.TrimSuffix(string(out), "\n"), nil
+}
```
qinqon: We can implement this with the golang git library so we don't depend on git version at CI.

tools/bumper/bumper.go
```diff
@@ -0,0 +1,194 @@
...
+	flag.StringVar(&paramArgs.componentsConfigPath, "config-path", "", "Full path to components yaml")
+	flag.StringVar(&paramArgs.gitToken, "token", "", "git Token")
+	flag.Parse()
+	if flag.NFlag() != 2 {
+		exitWithError(fmt.Errorf("Wrong Number of input parameters %d, should be 2. Use --help for usage", flag.NFlag()))
+	}
+}
+
+// since versioning of components can sometimes divert from semver standard, we do some refactoring
+func refactorVersion(version string) (*semver.Version, error) {
```
phoracek, line 167: `canonicalizeVersion` would maybe describe it better

tools/bumper/bumper.go
```diff
@@ -0,0 +1,194 @@
...
+	default:
+		return nil, fmt.Errorf("Failed to refactor version string %s", version)
+	}
+
+	return semver.NewVersion(version)
+}
+
+// check vtag format (example: 0.39.0-32-g1fcbe815)
+func isVtagFormat(tagVersion string) bool {
+	var vtagSyntax = regexp.MustCompile(`^[0-9]\.[0-9]+\.*[0-9]*-[0-9]+-[0-9,a-z]{9}`)
```
phoracek, line 187: It is not a-z, it is hexa, so a-f

tools/bumper/bumper.go
```diff
@@ -0,0 +1,194 @@
...
+
+			//update component's entry in config yaml
+			component.Commit = updatedReleaseCommit
+			component.Metadata = updatedReleaseTag
+			err = updateComponentsYaml(inputArgs.componentsConfigPath, componentsConfig)
+			if err != nil {
+				exitWithError(errors.Wrap(err, "Failed to update components yaml"))
+			}
+
+			// run bump script
```
phoracek, line 100: Redundant

tools/bumper/bumper.go
```diff
@@ -0,0 +1,194 @@
...
+			cmd := exec.Command("make", fmt.Sprintf("bump-%s", componentName))
+			if out, err := cmd.CombinedOutput(); err != nil {
+				exitWithError(errors.Wrapf(err, "Failed to run bump script. StdOut = %s", string(out)))
+			}
+
+			//create a new branch name
```
phoracek, line 106: Redundant

tools/bumper/bumper.go
```diff
@@ -0,0 +1,194 @@
...
+	flag.StringVar(&paramArgs.gitToken, "token", "", "git Token")
+	flag.Parse()
+	if flag.NFlag() != 2 {
+		exitWithError(fmt.Errorf("Wrong Number of input parameters %d, should be 2. Use --help for usage", flag.NFlag()))
+	}
+}
+
+// since versioning of components can sometimes divert from semver standard, we do some refactoring
+func refactorVersion(version string) (*semver.Version, error) {
+	//remove starting "v" if exists
```
phoracek, line 168: trailing "v"

tools/bumper/bumper.go
```diff
@@ -0,0 +1,194 @@
...
+	version = strings.TrimPrefix(version, "v")
+
+	//expand to 2 dotted format
```
phoracek, line 171: Please start all comments with `// ` (note the space after //)

tools/bumper/git.go
```diff
@@ -0,0 +1,175 @@
...
+	"strings"
+
+	"github.com/google/go-github/v32/github"
+	"github.com/pkg/errors"
+	"golang.org/x/oauth2"
+)
+
+// gitHubManager does the Github operations handling.
+type gitHubManager struct {
+	// github client.
```
phoracek, line 16: no point in this comment

tools/bumper/git.go
```diff
@@ -0,0 +1,175 @@
...
+	err := githubMgr.verifyClient()
+	if err != nil {
+		return nil, errors.Wrap(err, "Failed to verify connection with github client API")
+	}
+
+	return githubMgr, nil
+}
+
+func (mgr *gitHubManager) verifyClient() error {
+	//do a small sanity check
```
phoracek, line 44: This comment on its own is quite useless. Describe what you are doing in more detail or make the function self explanatory by its name.

tools/bumper/git.go
```diff
@@ -0,0 +1,175 @@
...
+		return "", "", errors.Wrap(err, "Failed to get release tag refs from github client API")
+	}
+
+	branchCommits, _, err := mgr.client.Repositories.ListCommits(mgr.ctx, owner, repo, &github.CommitsListOptions{SHA: branch})
+	if err != nil {
+		return "", "", errors.Wrap(err, "Failed to get release tag refs from github client API")
+	}
+
+	// look for the first tag that belongs to the chosen branch, going over from last tag
+	for i := len(tagRefs) - 1; i >= 0; i-- {
```
phoracek, line 111: Can't you iterate over the range?

tools/bumper/git.go
```diff
@@ -0,0 +1,175 @@
...
+func (mgr *gitHubManager) verifyClient() error {
+	//do a small sanity check
+	_, _, err := mgr.client.Users.Get(mgr.ctx, "")
+	if err != nil {
+		return errors.Wrapf(err, "Failed to get user from github client API")
+	}
+	return nil
+}
+
+func (mgr *gitHubManager) getCurrentReleaseTag(component component, repoDir string) (string, error) {
```
phoracek, line 52: All the functions in this module are quite complicated. Please add a proper doc string explaining what are they doing, what are the inputs, side effects and outputs

tools/bumper/bumper.go
```diff
@@ -0,0 +1,194 @@
...
+
+func isBumpNeeded(currentReleaseVersion, latestReleaseVersion, updatePolicy string) (bool, error) {
+	logger.Printf("currentReleaseVersion: %s, latestReleaseVersion: %s, updatePolicy: %s\n", currentReleaseVersion, latestReleaseVersion, updatePolicy)
+
+	if updatePolicy == updatePolicyStatic {
+		logger.Printf("updatePolicy is static. avoiding auto bump")
+		return false, nil
+	}
+
+	//if one of the tags is in vtag format (e.g 0.39.0-32-g1fcbe815), and not equal, then always bump
```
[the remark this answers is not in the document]
RamLavi: I return true only if they are equal.
I wanted to avoid handling with vtag version logic, so if they're not the same as before, then I assume we need to bump.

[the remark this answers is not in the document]
phoracek: Aha, the comment is for the body of the condition too. Missed that.

tools/bumper/bumper.go
```diff
@@ -0,0 +1,194 @@
...
+	if isVtagFormat(currentReleaseVersion) || isVtagFormat(latestReleaseVersion) {
+		return currentReleaseVersion == latestReleaseVersion, nil
+	} else {
```
phoracek, line 136: No need for else branch. There is an early return handled by the if. You can continue with the rest of the code on the root of the function.

The idiomatic way here is that the default flow should happen with as little indentation as possible.

tools/bumper/bumper.go
```diff
@@ -0,0 +1,194 @@
...
+		currentVersion, err := refactorVersion(currentReleaseVersion)
+		if err != nil {
+			return false, errors.Wrapf(err, "Failed to digest current Version %s to semver", currentVersion)
+		}
+		LatestVersion, err := refactorVersion(latestReleaseVersion)
```
phoracek, line 141: latestVersion

tools/bumper/bumper.go
```diff
@@ -4,11 +4,14 @@ import (
 	"bytes"
 	"flag"
 	"fmt"
+	"io/ioutil"
 	"log"
 	"os"
 	"os/exec"
 	"strings"
 
+	"gopkg.in/src-d/go-git.v4"
```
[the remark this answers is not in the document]
RamLavi: Done

tools/bumper/bumper.go
RamLavi: Done

tools/bumper/bumper.go
```diff
@@ -93,12 +108,13 @@ func main() {
 	}
 }
 
-func getLatestReleaseInfo(componentUrl string) (string, string, error) {
-	return "", "", fmt.Errorf("getLatestReleaseInfo Not yet implemented")
-}
+func getRepositoryOnBranch(repoDir, repositoryURL string, branchName string) (*git.Repository, error) {
```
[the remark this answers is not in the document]
RamLavi: ack. `component` is already taken so I'll use `gitComponent`

tools/bumper/git.go
```diff
@@ -0,0 +1,59 @@
...
+
+	err := githubMgr.verifyClient()
+	if err != nil {
+		return nil, errors.Wrap(err, "Failed to verify connection with github client API")
+	}
+
+	return githubMgr, nil
+}
+
+func (mgr *gitHubManager) verifyClient() error {
```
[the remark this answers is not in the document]
RamLavi: yeah I'll bring it back to the constructor

tools/bumper/git.go
RamLavi: Done

tools/bumper/git.go
```diff
@@ -0,0 +1,59 @@
...
+}
+
+func (mgr *gitHubManager) getCurrentReleaseTag(component component, repoDir string) (string, error) {
+	return "", fmt.Errorf("getCurrentReleaseTag not implemented")
+}
+
+func (mgr *gitHubManager) getUpdatedReleaseInfo(component component, repoDir string) (string, string, error) {
+	return "", "", fmt.Errorf("getUpdatedReleaseInfo not implemented")
+}
+
```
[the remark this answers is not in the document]
RamLavi: Done.

tools/bumper/git.go
RamLavi: Done

tools/bumper/git.go
```diff
@@ -0,0 +1,59 @@
...
+	"os/exec"
+	"strings"
+
+	"github.com/google/go-github/v32/github"
+	"github.com/pkg/errors"
+	"golang.org/x/oauth2"
+)
+
+// gitHubManager does the Github operations handling.
+type gitHubManager struct {
```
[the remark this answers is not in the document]
RamLavi: Done

tools/bumper/git.go
RamLavi: I'll add description in the commit where I implement it then

RamLavi: Done

tools/bumper/git.go
RamLavi: I can, but the latest tags are always last. 
So in order to find it faster I run from the end (regular golang iteration doesn't have a "reverse" run option)...

phoracek: Why don't we just take the last item? Is one of the steps expected to fail?

RamLavi: the last one may not belong to the branch specified. 
We want the latest tag that belongs to this branch, and no list (tag refs/branch commits) holds them both, so we need to so a cross search to find it.

tools/bumper/git.go
RamLavi: Done

tools/bumper/bumper.go
```diff
@@ -0,0 +1,194 @@
...
+		if err != nil {
+			exitWithError(errors.Wrapf(err, "Failed to discover if Bump need for %s", componentName))
+		}
+
+		if bumpNeeded {
+			//reset --hard git repo
+			exitWithError(fmt.Errorf("reset --hader repo not implemented yet"))
+
+			//PR name
+			PRTitle := fmt.Sprintf("Bump %s to %s", componentName, updatedReleaseTag)
```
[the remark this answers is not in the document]
RamLavi: Done

tools/bumper/bumper.go
RamLavi: Done

tools/bumper/bumper.go
RamLavi: Done

tools/bumper/bumper.go
RamLavi: Done

tools/bumper/bumper.go
RamLavi: Done

tools/bumper/bumper.go
RamLavi: Done

tools/bumper/bumper.go
RamLavi: right. Done.

tools/bumper/bumper.go
```diff
@@ -0,0 +1,180 @@
...
+		if err != nil {
+			exitWithError(errors.Wrapf(err, "Failed to get current release version tag from %s", componentName))
+		}
+
+		updatedReleaseTag, updatedReleaseCommit, err := gitComponent.getUpdatedReleaseInfo()
+		if err != nil {
+			exitWithError(errors.Wrapf(err, "Failed to get latest release version tag from %s", componentName))
+		}
+
+		bumpNeeded, err := isBumpNeeded(currentReleaseTag, updatedReleaseTag, component.Updatepolicy)
```
qinqon, line 72: We have the component fields inside `gitComponent` so we can call `getCurrentRelease` and `getUpdateReleaseInfo` inside `isBumpNeeded` and `isBumpNeeded` can be call `calculateBump` and return a `*bumpInfo` struct with `updatedReleaseCommit` and `updateReleaseTag` if is nill there is no bump needed else you use the fields there.

oshoval:
If you want to consider a progress bar, taken from go playground
```
package main

import (
	"fmt"
	"strings"
	"time"
)

func main() {
	const col = 30
	// Clear the screen by printing \x0c.
	bar := fmt.Sprintf("\x0c[%%-%vs]", col)
	for i := 0; i < col; i++ {
		fmt.Printf(bar, strings.Repeat("=", i)+">")
		time.Sleep(100 * time.Millisecond)
	}
	fmt.Printf(bar+" Done!", strings.Repeat("=", col))
}
```

tools/bumper/bumper.go
RamLavi: I don't see the advantage in putting everything inside `isBumpNeeded` (/`calculateBump`).
Right now I can test `isBumpNeeded()` with fabricated data to test only the bump decision logic.
IMO putting `getCurrentReleaseTag` and `getUpdatedReleaseInfo` complicates testing without real value.

What do you think?

RamLavi: oh.. forgot you're not available. @phoracek ?

RamLavi:
/hold cancel
unit tests added

RamLavi:
> If you want to consider a progress bar, taken from go playground
> 
> ```
> package main
> 
> import (
> 	"fmt"
> 	"strings"
> 	"time"
> )
> 
> func main() {
> 	const col = 30
> 	// Clear the screen by printing \x0c.
> 	bar := fmt.Sprintf("\x0c[%%-%vs]", col)
> 	for i := 0; i < col; i++ {
> 		fmt.Printf(bar, strings.Repeat("=", i)+">")
> 		time.Sleep(100 * time.Millisecond)
> 	}
> 	fmt.Printf(bar+" Done!", strings.Repeat("=", col))
> }
> ```

Thanks!
I found that go-git already has progress ioWriter when doing [git clone](https://github.com/src-d/go-git/blob/1a7db85bca7027d90afdb5ce711622aaac9feaed/_examples/progress/main.go#L26), I will add this in a different PR as this one is already too big

tools/bumper/bumper.go
```diff
@@ -0,0 +1,182 @@
...
+		exitWithError(errors.Wrap(err, "Failed to create github api instance"))
+	}
+
+	logger.Printf("Parsing %s", inputArgs.componentsConfigPath)
+	componentsConfig, err := parseComponentsYaml(inputArgs.componentsConfigPath)
+	if err != nil {
+		exitWithError(errors.Wrap(err, "Failed to parse components yaml"))
+	}
+
+	for componentName, component := range componentsConfig.Components {
```
phoracek, line 48: There is one missing part. Not sure if it should land in this, but I want to discuss it so we know the way forward.

We want to raise one PR per one component's bump right?

And we want to make sure we do not open another PR for a component that is already pending review/tests.

How are you planning to tackle these two points?

tools/bumper/gitFakes.go
```diff
@@ -0,0 +1,305 @@
+package main
```
phoracek, line 1: the name of this module should be git_fakes to match naming in the rest of the project

tools/bumper/component-commands.go
```diff
@@ -0,0 +1,311 @@
+package main
```
phoracek, line 1: The name of the file should be components_commands.go

tools/bumper/bumper.go
RamLavi: >We want to raise one PR per one component's bump right?

Yes. for each components that needs bumping we will open a PR with the githubApi.. do you foresee problems with this?

>make sure we do not open another PR for a component that is already pending review/tests

Yes I plan on adding in the next PRs a function called `isPrAlreadyOpened` that will scan all the current open PRs by title (which should be the same since it's automated).

tools/bumper/gitFakes.go
RamLavi: so git_fakes will be in a separate module/folder, right?

tools/bumper/bumper.go
phoracek: I tend to agree with Ram. It is better to have flat functions dependency tree instead of nesting too much. It helps with testing and keeps my brain RAM sane.

tools/bumper/gitFakes.go
phoracek: Oh sorry, I mean "this file".

tools/bumper/bumper.go
phoracek: > > We want to raise one PR per one component's bump right?
> 
> Yes. for each components that needs bumping we will open a PR with the githubApi.. do you foresee problems with this?
> 

Not really, I just don't see a logic preparing for that here (which is not bad), so I wanted to make sure it is accounted for.

> > make sure we do not open another PR for a component that is already pending review/tests
> 
> Yes I plan on adding in the next PRs a function called `isPrAlreadyOpened` that will scan all the current open PRs by title (which should be the same since it's automated).

Ack

tools/bumper/component-commands.go
RamLavi: I was following Quique's [mention](https://github.com/kubevirt/cluster-network-addons-operator/pull/479#discussion_r459907454) to keep package flat (i.e. nor using packages) in order to avoid using public functions.
What do you think?

phoracek: I'm ok with that, but how is that related to the file name?

tools/bumper/bumper.go
RamLavi: > > > We want to raise one PR per one component's bump right?
> > 
> > 
> > Yes. for each components that needs bumping we will open a PR with the githubApi.. do you foresee problems with this?
> 
> Not really, I just don't see a logic preparing for that here (which is not bad), so I wanted to make sure it is accounted for.
> 

Yeah I just prefer to add all high level logic functions, helps me make sure I didn't forget anything..

tools/bumper/gitFakes.go
RamLavi: Done

tools/bumper/component-commands.go
RamLavi: it's not, my bad

PR #483, opened 2020-07-29
Bump kubernetes-nmstate to 0.26.0

qinqon:
It uses kube-admission-webhook 0.12.0 that implements the CA overlap interval, although this interval
has the expiration time value so behaviour has to be the same.

**Release note**:

```release-note
Bump kubernetes-nmstate to v0.26.0
```

### DIFF
--- a/data/nmstate/operator.yaml
+++ b/data/nmstate/operator.yaml
@@ -46,6 +46,10 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: metadata.name
+            - name: POD_NAMESPACE
+              valueFrom:
+                fieldRef:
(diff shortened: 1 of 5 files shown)


data/nmstate/operator.yaml
```diff
@@ -235,12 +239,12 @@ spec:
       name: {{template "handlerPrefix" .}}nmstate-webhook
 ---
 apiVersion: v1
+  annotations:
```
phoracek, line 242: This must stay under metadata

qinqon: Something is wrong with rejectOwner patching.

PR #485, opened 2020-07-31
Check SetFromPods has current networkAddonsConfig generation

rhrazdil:
Second attempt to fix the infamous #451 

Added generation field to StatusManager, which is set in Reconcile and used in SetFromPods to 
verify that we're checking tha current networkAddonsConfig before setting status to Available.

This is a follow-up to #474
**Release note**:
```release-note
none
```

### DIFF
--- a/pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
+++ b/pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
@@ -190,7 +190,7 @@ func (r *ReconcileNetworkAddonsConfig) Reconcile(request reconcile.Request) (rec
 			// Request object not found, could have been deleted after reconcile request.
 			// Reset list of tracked objects.
 			// TODO: This can be dropped once we implement a finalizer waiting for all components to be removed
-			r.trackDeployedObjects([]*unstructured.Unstructured{})
+			r.stopTrackingObjects()
 
 			// Owned objects are automatically garbage collected. Return and don't requeue
 			return reconcile.Result{}, nil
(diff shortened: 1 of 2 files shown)


pkg/controller/statusmanager/status_manager.go
```diff
@@ -10,6 +10,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/openshift/cluster-network-operator/pkg/names"
```
phoracek, line 13: Wrong repo, should be kubevirt/cluster-network-addons-operator

rhrazdil: Right, fixed, thanks

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
```diff
@@ -190,7 +190,7 @@ func (r *ReconcileNetworkAddonsConfig) Reconcile(request reconcile.Request) (rec
 			// Request object not found, could have been deleted after reconcile request.
 			// Reset list of tracked objects.
 			// TODO: This can be dropped once we implement a finalizer waiting for all components to be removed
-			r.trackDeployedObjects([]*unstructured.Unstructured{})
+			r.trackDeployedObjects([]*unstructured.Unstructured{}, 0)
```
phoracek, line 193: This may introduce a bug. When we remove a NAC, this will be set tp empty list and 0. Let's say somebody then creates a new NAC. It will get generation 0 and couple of containers assigned. We can get into the same race again. Can we use negative number for the intial setting?

Or maybe better, can we introduce r.stopTracking. That would speak for itself.

rhrazdil:
> Thanks Radim.
> Let's hope this will solve it.
> can you add to the commit message a short explanation on how this Pr attempt to resolve the issue?
> I think it is not clear enough from the current message

Updated the commit message

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
rhrazdil: added method `stopTrackingObjects` and used negative value to reset the generation.

alonSadan:
Nice!
/lgtm

PR #487, opened 2020-08-04
Bump kubemacpool to v0.18.0

qinqon:
The kubemacpool v0.18.0 include some manager pod environment variable to control the
certificates rotation behaviour.

**Release note**:

```release-note
Bump kubemacpool to v0.18.0 
```

### DIFF
--- a/data/kubemacpool/kubemacpool.yaml
+++ b/data/kubemacpool/kubemacpool.yaml
@@ -285,6 +285,12 @@ spec:
             configMapKeyRef:
               key: RANGE_END
               name: kubemacpool-mac-range-config
+        - name: CA_ROTATE_INTERVAL
+          value: 8760h0m0s
+        - name: CA_OVERLAP_INTERVAL
+          value: 8760h0m0s
+        - name: CERT_ROTATE_INTERVAL
+          value: 4380h0m0s
         image: '{{ .KubeMacPoolImage }}'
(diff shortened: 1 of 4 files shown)


data/kubemacpool/kubemacpool.yaml
```diff
@@ -285,6 +285,12 @@ spec:
             configMapKeyRef:
               key: RANGE_END
               name: kubemacpool-mac-range-config
+        - name: CA_ROTATE_INTERVAL
+          value: 8760h0m0s
+        - name: CA_OVERLAP_INTERVAL
+          value: 8760h0m0s
```
phoracek, line 291: This is the same as rotate interval, how does that work? Do we always have 2 CAs?

phoracek:
/hold

Petr has a question

data/kubemacpool/kubemacpool.yaml
qinqon: This is the previous behaviour where CAs live at CABunlde until expiration time, but we can change if we want, in fact I suppose we have to make this numbers similar to the other kubevirt projects.

phoracek: Ok, if it matches the previous behavior let's be it. It looks odd, but it will be changed pretty soon anyway.

PR #490, opened 2020-08-05
Add kmp certificates knobs to CRD

qinqon:
This PR add the following fiels to the NetworkAddonsConfig CRD KubeMacPool component

caRotateInterval 
caOverlapInterval
certRotateInterval

**Release note**:

```release-note
Add certificate rotate knobs to the NetworkAddonsConfig CRD kubemacpool component
```


pkg/apis/networkaddonsoperator/v1alpha1/networkaddonsconfig_types.go
```diff
@@ -32,6 +39,7 @@ type NMState struct{}
 
 // +k8s:openapi-gen=true
 type KubeMacPool struct {
+	CertificateKnobs
```
phoracek, line 42: I think we should unify all cert settings under a single root-level structure (like imagePullPolicy) instead of making it specific for each component. Any reason not-to?

qinqon: Are we going to need to set different values per component ?

phoracek: Not under HCO. I'd start with single top-level settings. If somebody comes up with a reason to have those values different per each component, we can make it more granular.

qinqon: One issue I see is for example kmp webhooks are critical but not the case for knmstate, in case we find that there are some downtime during rotation we can increase rotation at kmp and not at knmstate, what do you think ?

phoracek: Good point. Let's think how it will be used.

The user will set some HCO CR. That will be propagated to CNAO CR, to one ore multiple knobs and then down to components.

If the scenario you describe will happen, there will be still only a single set of knobs on HCO, so if we want to change the config to prefent some bug from happening, we would need to hardcode the special value ourselves in HCO or CNAO.

At that point, granular control would make it slightly easier, but is not a must.

qinqon: Ack I will make it common.

pkg/apis/networkaddonsoperator/v1alpha1/networkaddonsconfig_types.go
```diff
@@ -9,13 +9,21 @@ import (
...
-	NMState         *NMState          `json:"nmstate,omitempty"`
-	MacvtapCni      *MacvtapCni       `json:"macvtap,omitempty"`
+	Multus           *Multus           `json:"multus,omitempty"`
+	LinuxBridge      *LinuxBridge      `json:"linuxBridge,omitempty"`
+	Ovs              *Ovs              `json:"ovs,omitempty"`
+	KubeMacPool      *KubeMacPool      `json:"kubeMacPool,omitempty"`
+	ImagePullPolicy  corev1.PullPolicy `json:"imagePullPolicy,omitempty"`
+	NMState          *NMState          `json:"nmstate,omitempty"`
+	MacvtapCni       *MacvtapCni       `json:"macvtap,omitempty"`
+	CertificateKnobs *CertificateKnobs `json:"certificateKnobs,omitempty"`
```
phoracek, line 19: That is pretty informal name. What about  `SelfSignConfiguration`, to follow KubeVirt's API?

pkg/network/certificates_knobs.go
```diff
@@ -0,0 +1,99 @@
...
+
+}
+
+func validateGreaterThan(lhsName string, lhsValue *time.Duration, rhsNamed string, rhsValue *time.Duration, errs *[]error) {
+	if lhsValue == nil || rhsValue == nil {
+		return
+	}
+
+	if *rhsValue > *lhsValue {
+		*errs = append(*errs, errors.Errorf("failed to validate: %s(%s) has to be <= %s(%s)", rhsNamed, *rhsValue, lhsName, *lhsValue))
```
phoracek, line 97: Return list of errors and join them on the caller level. We do it that way for all the other functions

pkg/network/certificates_knobs.go
```diff
@@ -0,0 +1,99 @@
...
+
+	validateGreaterThan("caRotateInterval", caRotateInterval, "caOverlapInterval", caOverlapInterval, &errs)
+	validateGreaterThan("caRotateInterval", caRotateInterval, "certRotateInterval", certRotateInterval, &errs)
+
+	return errs
+}
+
+func fillDefaultsCertificateKnobs(conf, previous *opv1alpha1.NetworkAddonsConfigSpec) []error {
+	if conf.CertificateKnobs == nil {
+		return []error{}
```
phoracek, line 51: This functions should fill in defaults. Could we keep them here? Something close to the current KubeVirt defaults. Otherwise we will have undefined behavior where everything is controlled by the component, I don't think that is desirable.

pkg/network/certificates_knobs.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+
+	if conf.CertificateKnobs.CARotateInterval == "" || conf.CertificateKnobs.CAOverlapInterval == "" || conf.CertificateKnobs.CertRotateInterval == "" {
```
phoracek, line 54: When would this happen? changeSafeCertificateKnobs checks against this option, doesn't it?

pkg/apis/networkaddonsoperator/v1alpha1/networkaddonsconfig_types.go
qinqon: Done.

pkg/network/certificates_knobs.go
qinqon: Done. But with a helper function.

pkg/network/certificates_knobs.go
qinqon: the kube-admission-webhook have some defaults, but we can put here the ones like kubevirt.

qinqon: Done.

pkg/network/certificates_knobs.go
phoracek: Aha, nevermind, it will be useful to bring the previously implicitly configured values. Just like it is done in https://github.com/kubevirt/cluster-network-addons-operator/blob/36f9b55ac1a1f6481960fbca84fe50dcc58b235e/pkg/network/kubemacpool.go#L66

qinqon: If some of the knobs is missing we set to defaults all of them, since we don't want to revalidate here, is kind of all or nothing.

pkg/network/self_sign_configuration.go
```diff
@@ -0,0 +1,130 @@
...
+
+func parseCertificateKnob(name, value string) (*time.Duration, error) {
+	if value == "" {
+		return nil, nil
+	}
+	d, err := time.ParseDuration(value)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to parse %s", name)
+	}
+	return &d, nil
```
phoracek, line 100: We always have a non-nil duration returned here. Why do you return reference? If we drop the reference, we can get rid of all the nil checks in other functions

pkg/network/self_sign_configuration.go
```diff
@@ -0,0 +1,130 @@
...
+}
+
+func fillDefaultsSelfSignConfiguration(conf, previous *opv1alpha1.NetworkAddonsConfigSpec) []error {
+	if conf.SelfSignConfiguration == nil || conf.SelfSignConfiguration.CARotateInterval == "" || conf.SelfSignConfiguration.CAOverlapInterval == "" || conf.SelfSignConfiguration.CertRotateInterval == "" {
+		if previous != nil && previous.SelfSignConfiguration != nil {
+			conf.SelfSignConfiguration = previous.SelfSignConfiguration
+			return []error{}
+		}
+
+		conf.SelfSignConfiguration = &opv1alpha1.SelfSignConfiguration{
```
phoracek, line 75: This may result in unexpected behavior. Let's say a user only configures the overlap, the config is accepted, but then they get completely different intervals.

We need to make sure that if somebody sets conf.SelfSignConfiguration, it always contains all 3 values. Would you please fail hard in parseCertificateKnob when the value is empty?

qinqon: We should validate _after_ setting the defaults if we do that.

phoracek: Why? When would it break if we keep it as is?

qinqon: Imaging that we have the following config
```yaml
caRotateInterval: ""
caOverlapInterval: "500h"
certRotateInterval: ""
```
Clearly `caOverlapInterval` is beyond `caRotateInterval` and that's wrong, so we either use defaults at validation or we set defaults and then validate.

phoracek: If somebody passes this config, it should fail validation with "you must provide all parameters". Would not that solve it?

pkg/network/self_sign_configuration.go
qinqon: that's not true

```golang
if value == "" {                                                            
        return nil, nil                                                         
    }    
```

phoracek: And if we change the validation? Empty config (with "" attributes) will not be allowed. It must be `nil` or fully populated. Then we can depend on it being set to a value.

pkg/network/self_sign_configuration.go
qinqon: Ack that's good.

README.md
```diff
@@ -187,6 +187,23 @@ spec:
   imagePullPolicy: Always
 ```
 
+## Self Signed Certificates Configuration
+
+Administrator can specify [webhook self signed certificates configuration](https://pkg.go.dev/github.com/qinqon/kube-admission-webhook@v0.12.0/pkg/certificate?tab=doc#Options)
```
RamLavi, line 192: I think we should mention which components are currently using this certification manager service

hack/components/bump-kubemacpool.sh
```diff
@@ -46,6 +46,13 @@ spec:
       - image: "{{ .KubeMacPoolImage }}"
         imagePullPolicy: "{{ .ImagePullPolicy }}"
         name: manager
+        env:
+          - name: CA_ROTATE_INTERVAL
+            value: '{{ .CARotateInterval }}'
```
RamLavi, line 51: please use `"` to be consistent with other parameters in the heredoc

hack/components/bump-kubemacpool.sh
```diff
@@ -46,6 +46,13 @@ spec:
...
+          - name: CA_OVERLAP_INTERVAL
+            value: '{{ .CAOverlapInterval }}'
```
RamLavi, line 53: ditto

hack/components/bump-kubemacpool.sh
```diff
@@ -46,6 +46,13 @@ spec:
...
+          - name: CERT_ROTATE_INTERVAL
+            value: '{{ .CertRotateInterval }}'
```
RamLavi, line 55: ditto

pkg/network/self_sign_configuration.go
```diff
@@ -0,0 +1,141 @@
...
+
+	"github.com/pkg/errors"
+
+	opv1alpha1 "github.com/kubevirt/cluster-network-addons-operator/pkg/apis/networkaddonsoperator/v1alpha1"
+)
+
+const (
+	caRotateIntervalDefault   = 7 * 24 * time.Hour // 7 days
+	caOverlapIntervalDefault  = 24 * time.Hour     // 1 day
+	certRotateIntervalDefault = 24 * time.Hour     // 1 day
```
RamLavi, line 15: why is this the default? are we aligned to kubevirt intervals?

qinqon: Yep we are aligned with them

pkg/network/self_sign_configuration_test.go
```diff
@@ -0,0 +1,354 @@
...
+						CARotateInterval:   "48h30m50s",
+						CAOverlapInterval:  "48h30m50s",
+						CertRotateInterval: "",
+					},
+				}
+				errorList := validateSelfSignConfiguration(clusterConfig)
+				Expect(errorList).To(HaveLen(1), "validation failed due to an unexpected error: %v", errorList)
+				Expect(errorList[0].Error()).To(Equal("failed to validate selfSignConfiguration: certRotateInterval is missing"))
+			})
+		})
```
phoracek, line 53: This can be written as a table checking that if one is missing, it return an error.

pkg/network/self_sign_configuration_test.go
```diff
@@ -0,0 +1,354 @@
...
+						CARotateInterval:   "12h",
+						CAOverlapInterval:  "10h",
+					},
+				}
+				errorList := validateSelfSignConfiguration(clusterConfig)
+				Expect(errorList).To(HaveLen(1), "validation failed due to an unexpected error: %v", errorList)
+				Expect(errorList[0].Error()).To(MatchRegexp("failed to validate selfSignConfiguration: error parsing certRotateInterval: time"))
+
+			})
+		})
```
phoracek, line 110: Can be a table

pkg/network/self_sign_configuration_test.go
```diff
@@ -0,0 +1,354 @@
...
+						CAOverlapInterval:  "10h",
+						CertRotateInterval: "0",
+					},
+				}
+				errorList := validateSelfSignConfiguration(clusterConfig)
+				Expect(errorList).To(HaveLen(1), "validation failed due to an unexpected error: %v", errorList)
+				Expect(errorList[0].Error()).To(Equal("failed to validate selfSignConfiguration: certRotateInterval duration has to be > 0"))
+
+			})
+		})
```
phoracek, line 153: Table

PR #501, opened 2020-08-11
Prefix version tag with "v"

phoracek:

Go mod expects versions to start with "v". Without that, go mod is able
to fetch the tag, but it interprets it as:
v0.3.1-0.20200811082954-d0b735298e44

With this change, we prefix release tags with "v" while keeping the
version in our CSVs with a naked version to satisfy OLM's expectations.

**Release note**:

```release-note
Prefix version tags with "v"
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -173,7 +173,7 @@ prepare-major:
 release: $(GITHUB_RELEASE)
 	DESCRIPTION=version/description \
 	GITHUB_RELEASE=$(GITHUB_RELEASE) \
-	TAG=$(shell hack/version.sh) \
+	TAG=v$(shell hack/version.sh) \
 	  hack/release.sh \
 	    manifests/cluster-network-addons/cluster-network-addons.package.yaml \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -173,7 +173,7 @@ prepare-major:
 release: $(GITHUB_RELEASE)
```
RamLavi, line 173: PR title typo: `Previx`-->`Prefix`

RamLavi:
why does OLM expect naked version? When can we align so that all will be with `v`?

phoracek:
I don't know why. The naked version is used as a part of the CSV and also for directory names under https://github.com/kubevirt/cluster-network-addons-operator/tree/master/manifests/cluster-network-addons. Changing it would be costly. I take it as two notations we can use for the same version. One is the version, one is the tag. Maybe we can converge back to naked if go mod ever supports it. Moving OLM may be more disruptive.

PR #505, opened 2020-08-17
add stdout to git-clone process in bumper script

RamLavi:
Currently when running the bumper script,
there is no output to stdout when a component is cloned
to the temporary folder. As this may take a short while,
it could be seen as if the script is stuck.

Let's dump the clone command output to stdout,
thus making it more user-friendly.

**Release note**:

```release-note
NONE
```


phoracek (approved):
/lgtm
/approve

PR #506, opened 2020-08-17
add bumper go executable to .gitignore

RamLavi:
building the bumper.go script creates a bumper executable file.
This may cause a future bumping PR to fail travis (git --porcelain will fail with this unmerged file)
Moreover, we don't want to accidentally commit it.
Hence, we want to add it to .gitignore

**Release note**:
```release-note
NONE
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,3 +26,7 @@ goimports-check
 vet
 whitespace
 whitespace-check
+
+# bumper go exec file
+bumper
+


.gitignore
```diff
@@ -26,3 +26,7 @@ goimports-check
 vet
 whitespace
 whitespace-check
+
+# bumper go exec file
+bumper
+
```
phoracek, line 32: Drop the trailing line

.gitignore
```diff
@@ -26,3 +26,7 @@ goimports-check
 vet
 whitespace
 whitespace-check
+
+# bumper go exec file
```
phoracek, line 30: Can we build it in the _out or build_output directory instead?

RamLavi: I was thinking of ultimately running the auto-bump via the `Makefile` and that runs on the primary folder.
So building it on the _out folder only so that the bumper exec file will be ignored might look awkward if you stick to my makefile paradigm.. 
What do you think?

phoracek: `go run` generates the binary in the root folder?

RamLavi: since our script is comprised out of more than 1 file, we need to build it first. this creates the bumper exec file

phoracek: Ok, but then you can specify the destination for the file. Just like we do it with all the other binaries https://github.com/kubevirt/cluster-network-addons-operator/blob/master/Makefile#L54

RamLavi: ack. ok so this PR is not needed then..

PR #511, opened 2020-08-23
Migrate CNAO to v1

RamLavi:
Currently CNAO CR version is v1alpha1. 
In order to comply with [CFC operator Best practice](https://docs.engineering.redhat.com/display/CFC/Best_Practices), we need to upgrade CNAO to v1.
This PR follows the [operator-SDK migration Guide](https://github.com/operator-framework/operator-sdk/blob/11aa2af7aaa202cb38b2417031e6af2035c7948c/doc/user/migrating-existing-apis.md#golang-api-migrations-types-and-commonalities).

- **generate v1 API Version with Operator-SDK**
Using operator-SDK API, we generate a new API-Version v1 files
This commit only includes the generated fils with no other changes.

- **adding a new shared apis folder that holds shared API**
Following operator-SDK migration manual, and since v1alpha
is the same in it's content to the new v1, we create a shared folder
to hold all the shared API-Vesrion types, to be used by v1alpha and v1 in future commits.
Also, as the shared folder is just a helper version folder, it does not register it's schema.

- **update empty v1 and v1alpha1's networkaddonsconfig_types.go to use shared types**
We want to reuse shared API-Version type structs to be used both on v1alpha1 and v1.
This way both API-Versions use the same structs via aliasing

- **generate gen-k8s deepcopy files**
Using operator-SDK to generate deepcopy files for the v1, v1alpha1, shared API-Versions.

- **update CustomResourceDefinition manifests**
In CNAO repo the CRD is created by the manifest-templator,
so generation of the CRD with the operator-SDK API is not straight forward.
Instead we manually update components.go to introduce both v1, v1alpha1 version to the CRD
Also, we remove the use of Spec.Version which is deprecated, and update the CRD to use
the appropriate CNAO version types.

- **update pkg code to use appropriate cnao library**
We need to change all existing CNAO code to use the shared
API-Version types.

- **Refactor at end of Reconcile of NetworkAddonsConfig obj to v1**
We want the controller to migrate the NetworkAddonsConfig to v1
by doing that, we automatically upgrade v1alpha to v1

- **update test code to use appropriate cnao library**
We need to update all tests code to use the new CNAO
API-Version types.

**Release note**:

```release-note
migrate CNAO to v1
```


pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
```diff
@@ -134,7 +135,7 @@ func add(mgr manager.Manager, r *ReconcileNetworkAddonsConfig) error {
 	}
 
 	// Watch for changes to primary resource NetworkAddonsConfig
-	if err := c.Watch(&source.Kind{Type: &opv1alpha1.NetworkAddonsConfig{}}, &handler.EnqueueRequestForObject{}, pred); err != nil {
+	if err := c.Watch(&source.Kind{Type: &cnaov1.NetworkAddonsConfig{}}, &handler.EnqueueRequestForObject{}, pred); err != nil {
```
qinqon, line 138: Don't we have to watch for v1alpha1 so in case networkaddonsconfig is allready there we Reconcile it ?

rhrazdil (commented):
Looks good to me
/lgtm

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
```diff
@@ -308,15 +309,15 @@ func (r *ReconcileNetworkAddonsConfig) renderObjects(networkAddonsConfig *opv1al
...
+		labels[cnaov1.SchemeGroupVersion.Group+"/version"] = operatorVersionLabel
 		obj.SetLabels(labels)
 	}
 
 	return objs, nil
 }
 
 // Validate and returns the previous configuration spec
-func (r *ReconcileNetworkAddonsConfig) getPreviousConfigSpec(networkAddonsConfig *opv1alpha1.NetworkAddonsConfig) (*opv1alpha1.NetworkAddonsConfigSpec, error) {
+func (r *ReconcileNetworkAddonsConfig) getPreviousConfigSpec(networkAddonsConfig *cnaov1.NetworkAddonsConfig) (*cnaov1.NetworkAddonsConfigSpec, error) {
```
qinqon, line 320: If we pass `name string, spec *cnao.NetworkAddonsSpec` then we don't need to reference cnaov1 here

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
```diff
@@ -277,7 +278,7 @@ func (r *ReconcileNetworkAddonsConfig) Reconcile(request reconcile.Request) (rec
 }
 
 // Render objects for all desired components
-func (r *ReconcileNetworkAddonsConfig) renderObjects(networkAddonsConfig *opv1alpha1.NetworkAddonsConfig, openshiftNetworkConfig *osv1.Network) ([]*unstructured.Unstructured, error) {
+func (r *ReconcileNetworkAddonsConfig) renderObjects(networkAddonsConfig *cnaov1.NetworkAddonsConfig, openshiftNetworkConfig *osv1.Network) ([]*unstructured.Unstructured, error) {
```
qinqon, line 281: If we pass `SchemeGroupVersion` and `cnao.NetworkAddonsConfigSpec` we don't need to refeerence cnaov1

qinqon (changes requested):
First pass is not a line by line review, but I have general comments

We should use `shared` as much as possible so we don't reference version, to do so all the helper functions need to accept and return stuff from `shared`.

I propose to create a `shared.NetworkAddonsConfig` that is created from versioned at `Reconcile` and then at the end of `Reconcile` is converted to versioned, this way if we have to do some kind of migrations we do just there.

pkg/apis/networkaddonsoperator/v1/networkaddonsconfig_types.go
```diff
@@ -0,0 +1,65 @@
...
+type Ovs shared.Ovs
+
+// +k8s:openapi-gen=true
+type NMState shared.NMState
+
+// +k8s:openapi-gen=true
+type KubeMacPool shared.KubeMacPool
+
+// +k8s:openapi-gen=true
+type MacvtapCni shared.MacvtapCni
```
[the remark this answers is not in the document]
RamLavi: This is part of the cnao types.. it's the same for v1alpha1, so I am just keeping consistency

pkg/apis/networkaddonsoperator/v1/networkaddonsconfig_types.go
```diff
@@ -2,30 +2,47 @@ package v1
...
+	"github.com/kubevirt/cluster-network-addons-operator/pkg/apis/networkaddonsoperator/shared"
+)
 
 // NetworkAddonsConfigSpec defines the desired state of NetworkAddonsConfig
-type NetworkAddonsConfigSpec struct {
-	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
-	// Important: Run "operator-sdk generate k8s" to regenerate code after modifying this file
-	// Add custom validation using kubebuilder tags: https://book-v1.book.kubebuilder.io/beyond_basics/generating_crd.html
-}
+// +k8s:openapi-gen=true
```
rhrazdil, line 10: So if we want to add new types in v1, should those be added here as well?

RamLavi: yes

pkg/apis/networkaddonsoperator/v1/networkaddonsconfig_types.go
```diff
@@ -0,0 +1,65 @@
...
+type NetworkAddonsConfigStatus shared.NetworkAddonsConfigStatus
+
+type Container shared.Container
+
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// NetworkAddonsConfig is the Schema for the networkaddonsconfigs API
+// +kubebuilder:subresource:status
+// +kubebuilder:resource:path=networkaddonsconfigs,scope=Namespaced
```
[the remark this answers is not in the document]
RamLavi: This is again copied from v1alpha1 types. 
tbh I don't much understand how these lines work, bu I figured that if they're needed in v1alpha1 then it's also relevant here.

Need I remove it from v1 and v1alpha1?

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
RamLavi: good catch!

pkg/components/components.go
```diff
@@ -3,14 +3,14 @@ package components
...
 
-	cnav1alpha1 "github.com/kubevirt/cluster-network-addons-operator/pkg/apis/networkaddonsoperator/v1alpha1"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	extv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
-	opv1alpha1 "github.com/kubevirt/cluster-network-addons-operator/pkg/apis/networkaddonsoperator/v1alpha1"
+	cnao "github.com/kubevirt/cluster-network-addons-operator/pkg/apis/networkaddonsoperator/shared"
```
alonSadan, line 12: maybe call it `shared` or `cnaoshared`? could be a bit confusing

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
RamLavi: since I added `shared.NetworkAddonsConfig` then I guess it's not needed now right? or do you prefer I do the change?

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
RamLavi: ditto

pkg/components/components.go
```diff
@@ -326,9 +326,8 @@ func GetCrd() *extv1beta1.CustomResourceDefinition {
 			Name: "networkaddonsconfigs.networkaddonsoperator.network.kubevirt.io",
 		},
 		Spec: extv1beta1.CustomResourceDefinitionSpec{
-			Group:   "networkaddonsoperator.network.kubevirt.io",
-			Version: "v1alpha1",
```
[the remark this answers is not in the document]
RamLavi: it's [deprecated](https://github.com/kubernetes/apiextensions-apiserver/blob/master/pkg/apis/apiextensions/types.go#L41). we should only use `Versions` now

pkg/components/components.go
RamLavi: I think the point is to make everyone use the shared one instead of v1/v1alpha1, so I think `cnao` is a more appealing name.

alonSadan: alrighty then

RamLavi:
> I propose to create a `shared.NetworkAddonsConfig` that is created from versioned at `Reconcile` and then at the end of `Reconcile` is converted to versioned, this way if we have to do some kind of migrations we do just there.

Note this specific change is in this [commit](https://github.com/kubevirt/cluster-network-addons-operator/pull/511/commits/28df9ba93f745f99f6a594f7e26c83f6ec1d48ee), tell me if that's what you meant

RamLavi:
@qinqon 
> I propose to create a `shared.NetworkAddonsConfig` that is created from versioned at `Reconcile` and then at the end of `Reconcile` is converted to versioned, this way if we have to do some kind of migrations we do just there.

After playing with it and also [reading](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/) about it, I understood the following:
1. since v1alpha1 and v1 are the same, then the conversion to the `stored` version (aka our latest `v1`) is automatic when doing get/set k8s operations.
2. If I convert the `NetworkAddonsConfig` to shared at start of reconcile (to allow all later types to use the shared) fails - because later in the code we try to `Get/Set` this NetworkAddonsConfig and that fails (since it's using the shared which isn't registered). an example for this is when we `SetControllerReference()` 
3. This leads me to conclude that at the very least we need to keep `NetworkAddonsConfig` at the original stored version. 

what do you think?

qinqon:
@RamLavi so we are going to receive "v1" at `Reconcile` always right ?

So we do the .Get  with `v1` and copy content to `share.NetworkAddonsConfig` and pass to functions right ?

RamLavi:
> @RamLavi so we are going to receive "v1" at `Reconcile` always right ?
> 
> So we do the .Get with `v1` and copy content to `share.NetworkAddonsConfig` and pass to functions right ?

that doesn't work, because somewhere along the way, some function (like `SetControllerReference()`) will try to Get this `NetworkAddonsConfig` and since we converted it to `shared` it will fail..

qinqon:
@RamLavi Then I would say that we version the functions that really need versioned manifests under pkg with something like `pkg/v1` or the like 

But I am not really sure, at the end we want groupd versined and shared stuff.

RamLavi:
> @RamLavi Then I would say that we version the functions that really need versioned manifests under pkg with something like `pkg/v1` or the like
> 
> But I am not really sure, at the end we want group versined and shared stuff.

Well, if k8s does the conversion for us, why don't we just use v1 for all of Reconcile?

qinqon:
@RamLavi if we image an scenario where we have v2 (meaning we are not retrocompatible with v1) then we will have to rename everything again, propose to reduce v1 reference and group shared and versioned stuff together.

RamLavi:
> @RamLavi if we image an scenario where we have v2 (meaning we are not retrocompatible with v1) then we will have to rename everything again, propose to do as much as we can to just reference shared and group together the stuff that needs v1.

I think we can't avoid using `cnaov1.NetworkAddonsConfig`, i.e. not shared type in this case. Let me explain:
If the `Get/Set` functions are made in a third party function (example: `controllerutil.SetControllerReference()`) then passing the v1 means keeping 2 types simultaneously, which I think is a bad idea. 

Moreover, taking your v2 scenario, I think it's unavoidable to change then the code the v1/v2 codes, but right now it's not necessary. but we won't encounter the issue I currently have since v2 will be registered (unlike shared)

So I propose keep what we can as shared, and what we can't (I think currently only `cnaov1.NetworkAddonsConfig`) we don't.
What say you? @qinqon

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
```diff
@@ -308,15 +333,15 @@ func (r *ReconcileNetworkAddonsConfig) renderObjects(networkAddonsConfig *opv1al
 		if labels == nil {
 			labels = map[string]string{}
 		}
-		labels[opv1alpha1.SchemeGroupVersion.Group+"/version"] = operatorVersionLabel
+		labels[cnaov1.SchemeGroupVersion.Group+"/version"] = operatorVersionLabel
```
qinqon, line 336: Can we remove this ference somehow if not we have to use `v1` at the function name

pkg/controller/statusmanager/status_manager.go
```diff
@@ -85,7 +86,7 @@ func (status *StatusManager) Set(reachedAvailableLevel bool, conditions ...condi
 // set updates the NetworkAddonsConfig.Status with the provided conditions
 func (status *StatusManager) set(reachedAvailableLevel bool, conditions ...conditionsv1.Condition) error {
```
qinqon, line 87: We are doing two things here, creating the spected status and storing it at the versioned version, let's split that so we do the versined part ("storing the status") out of here.

pkg/controller/statusmanager/status_manager.go
```diff
@@ -346,7 +347,7 @@ func (status *StatusManager) SetFromPods() {
 	// If all pods are being created, mark deployment as not failing
 	status.SetNotFailing(PodDeployment)
 
-	config := &opv1alpha1.NetworkAddonsConfig{}
+	config := &cnaov1.NetworkAddonsConfig{}
```
qinqon, line 350: Don't know what to do with this one, maybe we need a generic interface for This operations that has a v1 implementation...

pkg/components/components.go
```diff
@@ -428,22 +432,22 @@ func GetCrd() *extv1beta1.CustomResourceDefinition {
 	return crd
 }
 
-func GetCR() *cnav1alpha1.NetworkAddonsConfig {
-	return &cnav1alpha1.NetworkAddonsConfig{
+func GetCR() *cnaov1.NetworkAddonsConfig {
```
qinqon, line 435: `GetCRV1`

qinqon (changes requested):
I still see some cnaov1 references at status_manager, would be nice to have just a few references to it at the _controller.go and the rest of the code managing shared

pkg/apis/networkaddonsoperator/shared/networkaddonsconfig_types.go
```diff
@@ -0,0 +1,88 @@
...
+	Name       string `json:"name"`
+	Image      string `json:"image"`
+}
+
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// NetworkAddonsConfig is the Schema for the networkaddonsconfigs API
+// +kubebuilder:subresource:status
+// +kubebuilder:resource:path=networkaddonsconfigs,scope=Namespaced
+// +k8s:openapi-gen=true
```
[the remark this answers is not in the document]
RamLavi: Done

pkg/apis/networkaddonsoperator/v1/networkaddonsconfig_types.go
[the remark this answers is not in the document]
RamLavi: you're right.. will remove.
Done

pkg/apis/networkaddonsoperator/v1/networkaddonsconfig_types.go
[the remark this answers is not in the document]
RamLavi: ack. keeping.

pkg/components/components.go
RamLavi: Done

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
```diff
@@ -195,10 +201,19 @@ func (r *ReconcileNetworkAddonsConfig) Reconcile(request reconcile.Request) (rec
 			// Owned objects are automatically garbage collected. Return and don't requeue
 			return reconcile.Result{}, nil
 		}
+
+		log.Printf("Error reading NetworkAddonsConfig. err = %v", err)
```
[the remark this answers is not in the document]
RamLavi: I didn't see it while debugging while it failed there. Are you sure it is presented somewhere?

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
```diff
@@ -195,10 +201,19 @@ func (r *ReconcileNetworkAddonsConfig) Reconcile(request reconcile.Request) (rec
...
 		// Error reading the object - requeue the request.
 		return reconcile.Result{}, err
 	}
 
+	networkAddonsConfig, err := r.ConvertNetworkAddonsConfigToShared(networkAddonsConfigStorageVersion)
+	if err != nil {
+		// If failed, set NetworkAddonsConfig to failing and requeue
+		r.statusManager.SetFailing(statusmanager.OperatorConfig, "updateNetworkAddonsConfigToV1", err.Error())
```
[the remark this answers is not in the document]
RamLavi: Done

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
```diff
@@ -276,8 +291,18 @@ func (r *ReconcileNetworkAddonsConfig) Reconcile(request reconcile.Request) (rec
 	return reconcile.Result{RequeueAfter: time.Minute}, nil
 }
 
+// Convert NetworkAddonsConfig to shared type
+func (r *ReconcileNetworkAddonsConfig) ConvertNetworkAddonsConfigToShared(networkAddonsConfig *cnaov1.NetworkAddonsConfig) (*cnao.NetworkAddonsConfig, error) {
```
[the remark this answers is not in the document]
RamLavi: Done

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
RamLavi: I think we need to use it because `SchemeGroupVersion` type is only set in `register.go` which isn't present in shared. 
Will change to `renderObjectsV1()`

pkg/controller/statusmanager/status_manager.go
RamLavi: you mean you want a function like `getNetworkAddonsConfigV1()`?

pkg/controller/statusmanager/status_manager.go
RamLavi: like `etworkAddonsStorageVersion` interface that will just implement stuff like 
`getNetworkAddonsConfig()`
`getSchemeGroupVersion()`
?
I'm liking this more and more by the moment

qinqon: Something like that, and maybe the same to get and set so we can put the  "migration" code there.

RamLavi: ok I'll do it in a separate PR.

pkg/controller/statusmanager/status_manager.go
RamLavi: will do it as part of the following PR using the new `NetworkAddonsStorageVersion` interface

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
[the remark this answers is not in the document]
qinqon: humm now that you say it will re Reconcile let's keep it

PR #514, opened 2020-08-24
Knmstate configure cert env vars

qinqon:
Configure knmstate cert rotation parameters from the CR
    
It will take the config from the selfSignConfiguration field at the networkaddonsconfig CR.

**Release note**:

```release-note
Use selfSignConfiguration to configure kubernetes-nmstate cert rotation parameters.
```


phoracek (commented):
/approve

PR #515, opened 2020-08-24
Add events to status manager and controller

alonSadan:

In order to  bring CNAO to meet operators conventional best practices,
It should use k8s events to describe actions that are being taken.

In this PR:

- A new event emitter is introduced, to easily produce events.
the emitter has to be initiated and then can be used to create events.   
Right now it only support associating the event to a runtime object,
but can be improved in the future to also support associating events to
CSVs and the self pod.

- Add events to log the deployment of all (not each) component.
Three types of events will be logged:  Available, Progressing and Failed.
The events will be associated with the networkAddonsConfig object. and 
will be produced in statusManager.

- Add an event on changes made to networkAddonsConfig.Spec,
to log any addition or removal of a network component.

- Added some info about the newly added events, and how to get them to the readme.

- Removed redundant error formatting from Wrapf() functions.  
   

events output( `kubectl describe kubectl describe networkaddonsconfig/cluste`) : 

Modified: 
```
  Type    Reason    Age   From                             Message
  ----    ------    ----  ----                             -------
  Normal  Modified  92s   cluster-network-addons-operator  Config spec was modified
```

Progressing: 
```
  Type    Reason    Age   From                             Message
  ----    ------    ----  ----                             -------
Normal  Progressing  86s (x4 over 88s)  cluster-network-addons-operator  Operator is progressing

```

Available: 
```
  Type    Reason    Age   From                             Message
  ----    ------    ----  ----                             -------
 Normal  Available    2s (x7 over 63s)   cluster-network-addons-operator  All applied componenets are available
```

Failed( this is an example, could be many Failing event types. in this example I gave illegal range for KMP ):
```
 Type    Reason    Age   From                             Message
  ----    ------    ----  ----                             -------
Warning  InvalidConfiguration  91s (x17 over 7m18s)  cluster-network-addons-operator  Operator failed: Human interaction is 
needed, please fix the desired configuration
```

**Release note**:

```release-note
k8s events were added to log failed, progressing, successful and modified deployment  
```


pkg/names/names.go
```diff
@@ -12,3 +12,6 @@ const APPLIED_PREFIX = "cluster-networks-addons-operator-applied-"
 // assigned with NetworkAddonsConfig as their owner. This can be used to prevent
 // garbage collection deletion upon NetworkAddonsConfig removal.
 const REJECT_OWNER_ANNOTATION = "networkaddonsoperator.network.kubevirt.io/rejectOwner"
+
+// ClusterNetworkOperatorName is the name of the cluster-network-addons-operator resource that will be reconciled
+const ClusterNetworkAddonsOperatorName = "cluster-network-addons-operator"
```
qinqon, line 17: Looks like const are UPPER_CASE at CNAO

pkg/util/k8s/event_emmiter.go
```diff
@@ -0,0 +1,37 @@
+package k8s
```
qinqon, line 1: Move this to `pkg/eventemitter/eventemitter.go` so we do `eventemitter.New()`

Also are we really sure that we need the interface ? I suspect is for unit testing if this is the case maybe we can mock `GetEventRecorderFor` somehow passing a fake manager.

qinqon (changes requested):
First pass

pkg/controller/statusmanager/status_manager.go
```diff
@@ -214,6 +233,12 @@ func (status *StatusManager) SetNotFailing(level StatusLevel) {
 		status.failing[level] = nil
 	}
 	status.syncFailing()
+
+	config := &opv1alpha1.NetworkAddonsConfig{}
+	if err := status.client.Get(context.TODO(), types.NamespacedName{Name: names.OPERATOR_CONFIG}, config); err != nil {
+		return
+	}
+	eventEmitter.EmitEvent(config, corev1.EventTypeNormal, "Progressing", fmt.Sprintf("Operator is progressing; Now on level %d", level))
```
phoracek, line 241: The level is internal detail, it would not provide user any extra information

pkg/eventemitter/eventemitter.go
```diff
@@ -0,0 +1,33 @@
+package eventemitter
+
```
RamLavi, line 2: commit typo `to  bring` (double space)

pkg/eventemitter/eventemitter.go
```diff
@@ -0,0 +1,33 @@
...
+import (
+	"github.com/kubevirt/cluster-network-addons-operator/pkg/names"
```
RamLavi, line 4: please put self imports in a separate section:
```
import (
	"k8s.io/apimachinery/pkg/runtime"
	"k8s.io/client-go/tools/record"
	"sigs.k8s.io/controller-runtime/pkg/manager"

	"github.com/kubevirt/cluster-network-addons-operator/pkg/names"
```

pkg/controller/statusmanager/status_manager.go
```diff
@@ -346,8 +366,8 @@ func (status *StatusManager) SetFromPods() {
 	// If all pods are being created, mark deployment as not failing
 	status.SetNotFailing(PodDeployment)
 
-	config := &opv1alpha1.NetworkAddonsConfig{}
-	if err := status.client.Get(context.TODO(), types.NamespacedName{Name: names.OPERATOR_CONFIG}, config); err != nil {
+	config := &opv1alpha1.NetworkAddonsConfig{ObjectMeta: metav1.ObjectMeta{Name: status.name}}
+	if err := status.client.Get(context.TODO(), types.NamespacedName{Name: status.name}, config); err != nil {
```
RamLavi, line 370: why is this change needed?

RamLavi (commented):
added so,e comments and questions

pkg/eventemitter/eventemitter.go
alonSadan: Done.

pkg/eventemitter/eventemitter.go
alonSadan: Done

pkg/controller/statusmanager/status_manager.go
alonSadan: I don't think there is much difference, I just wanted it be consistent with [this](https://github.com/kubevirt/cluster-network-addons-operator/blob/547ad444cd072db8493ce93c37809937904e83f8/pkg/controller/statusmanager/status_manager.go#L97) Get command. since I am using this `config` object I thought I mught as welll changed also the `Get` , and make it like the other `Get`s in this module. but if you think this is not related  here I can turn it back.

alonSadan:
removed level from message, fixed constant to be uper-case and mvoed the emitert to it's own package and added New() function [here](https://github.com/kubevirt/cluster-network-addons-operator/compare/111a868b7a3c74174617fe332f36f9d130341460..498a82b6208c698050b6a2905392b1a37224c1cf)

fixed typo in commit message. and separated the imports [here](https://github.com/kubevirt/cluster-network-addons-operator/compare/547ad444cd072db8493ce93c37809937904e83f8..e15ba2db9ddd5a59011ab7997fb31045772efb03)

pkg/controller/statusmanager/status_manager.go
RamLavi: no I don't mind a little piggy-banking, just wanted to make sure I'm not missing something of essence.

alonSadan:
added another commit to add events on change in the networkAddonsConfig spec

pkg/util/k8s/event_emmiter.go
qinqon: Nah EventEmitter feels right.

Well maybe the most important part is tu split where do we trigger the events and how do we format them, for that I sould say that we create some high level functions at `EventEmitter` that reads the networkaddons config and format the events as needed

```golang
func EmmitAvailable(configName string)
func EmmitProgressing(configName string)
func EmmitFailing(configName string)
func EmmitWarning(configName string, reason string, message string)
```

qinqon (changes requested):
Let's create some high level functions with event formating so we split responsabilities, added a comment to it.

pkg/util/k8s/event_emmiter.go
alonSadan: > Let's create some high level functions with event formating so we split responsabilities, added a comment to it.

Thanks @qinqon 
I added the external function. I hope that's what you meant.

I didn't understand what is EmitWarning. can you elaborate?

I agree we need to think together about were we wnat to place the events. this is my suggestion, but I am not really sure of it...

pkg/eventemitter/eventemitter.go
```diff
@@ -0,0 +1,91 @@
...
+
+func (ee eventEmitter) EmitEvent(object runtime.Object, eventType, reason, msg string) {
+	if object != nil {
+		ee.recorder.Event(object, eventType, reason, msg)
+	}
+}
+
+func (ee eventEmitter) EmitProgressingForConfig(configName string) {
+	config := ee.getConfigForEmitter(configName)
+	if config != nil {
```
qinqon, line 55: The `config != nil` is already check at `EmitEvent`  we can remove this from this function and it's friends.

qinqon (changes requested):
All good, just last nit.

pkg/eventemitter/eventemitter.go
alonSadan: Tnx. 
Done.

alonSadan:
I made some changes: 
- pods controller: I added spec-Modified-emit to the pods-contoroler, to better notify on changes to spec
- config-controller:  moved the the Modified-emit to the begging of the reconcile loop. Same reason.
-   fixed @qinqon  nit:)

alonSadan:
We need to decide if this is were we want our emitters to be place.

From what I see in my laptop the emitters follow the behavior of the status changes.  ( if status `Progressing` is changed to `true` from `false` then a `Progressing` event will be created i.e)

two points I noticed that I think require discussion:
- spec-modifications-events also created when the Spec is firstly introduced.
- Progressing-event is not created when components are being deleted, only when created.  

I am fine with thees two points, but would love to here more opinions.

there could be other points to discuss but it requires more QA time.

alonSadan:
added some info about the events in the readme

alonSadan:
removed  CLUSTER_NETWORK_ADDONS_OPERATOR_NAME name from names.go. it is redundant.

modified some Wrapf() functions , which used `"...%s",err..`  since it duplicates the error message in the logs

qinqon:
/lgtm
If we need add/modify/delete events we can do at follow up PRs.

test/e2e/workflow/deployment_test.go
```diff
@@ -151,25 +154,31 @@ var _ = Describe("NetworkAddonsConfig", func() {
...
 			UpdateConfig(configSpec)
+			CheckModifiedEvent()
+			CheckProgressingEvent()
 			CheckConfigCondition(ConditionAvailable, ConditionTrue, CheckImmediately, 30*time.Second)
 		})
 		//2281
 		It("should be able to remove all of them by removing the config", func() {
 			DeleteConfig()
 			CheckComponentsRemoval(components)
+			CheckModifiedEvent()
```
phoracek, line 170: We don't need to test it in every single test case. In general, you want to have one test for each feature. When it is done like that, it is easier to identify which feature is broken.

Do we have a suiting tests for status changes? You should make sure that we don't test events multiple times in identical scenarios

Alternatively

phoracek (changes requested):

test/check/events.go
```diff
@@ -0,0 +1,211 @@
+package check
```
phoracek, line 1: What is the source of this module?

alonSadan: Sorry forgot to add it to the commit message. Added it now.
https://github.com/kubevirt/kubevirt/blob/master/tests/utils.go#L234

test/e2e/workflow/deployment_test.go
alonSadan: > We don't need to test it in every single test case. In general, you want to have one test for each feature. When it is done like that, it is easier to identify which feature is broken.
> 

I understand. So I think the recovery test covers all types of event creation. I'll keep only this now
> Do we have a suiting tests for status changes?

I think this is tested along with the integration tests. I don't think there is a  dedicated suit if that's what you have meant.

PR #524, opened 2020-09-02
make it possible to run `kubectl explain crd`

RamLavi:
Let's make make it possible to run `kubectl explain networkaddonsconfigs`

- **add components annotations to types**
`operator-sdk generate crds` generates descriptions from the annotations added to `networkaddonsconfig_types.go`
This commit adds these annotations.

- **fix NetworkAddonsConfig scope to Cluster in api types**
Currently there is a inconsistency between the manifest templator
outputed crd manifest and the type annotations in scope field.
Altough the manifests are controlled by the manifest generator
lets fix the annotation avoid future confusion.

- **add openAPIV3Schema validation parameters**
Let's add openAPIV3Schema detail for "kubectl explain crd"
Since operator-sdk crd cannot run. this commit manually creates the openAPIV3Schema
that should've been generated by operator-sdk crd API.

- **adding test to make sure crd is explainable**

PR depends on #537, #593 to be merged

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/apis/networkaddonsoperator/shared/networkaddonsconfig_types.go
+++ b/pkg/apis/networkaddonsoperator/shared/networkaddonsconfig_types.go
@@ -21,15 +21,22 @@ type NetworkAddonsConfigSpec struct {
 }
 
 // +k8s:openapi-gen=true
+// SelfSignConfiguration defines self sign configuration
 type SelfSignConfiguration struct {
+	// CARotateInterval defines duration for CA and certificate
 	CARotateInterval   string `json:"caRotateInterval,omitempty"`
+	// CAOverlapInterval defines the duration of CA Certificates at CABundle if not set it will default to CARotateInterval
 	CAOverlapInterval  string `json:"caOverlapInterval,omitempty"`
+	// CertRotateInterval defines duration for of service certificate
 	CertRotateInterval string `json:"certRotateInterval,omitempty"`
 }
 
 // +k8s:openapi-gen=true
+// PlacementConfiguration defines node placement configuration
 type PlacementConfiguration struct {
+	// Infra defines placement configuration for master nodes
 	Infra     *Placement `json:"infra,omitempty"`
+	// Workloads defines placement configuration for worker nodes
 	Workloads *Placement `json:"workloads,omitempty"`
 }
 
@@ -40,24 +47,32 @@ type Placement struct {
 	Tolerations  []corev1.Toleration `json:"tolerations,omitempty"`
 }
 
+// Multus plugin enables attaching multiple network interfaces to Pods in Kubernetes
 // +k8s:openapi-gen=true
 type Multus struct{}
 
+// LinuxBridge plugin allows users to create a bridge and add the host and the container to it
 // +k8s:openapi-gen=true
 type LinuxBridge struct{}
 
+// Ovs plugin allows users to define Kubernetes networks on top of Open vSwitch bridges available on nodes
 // +k8s:openapi-gen=true
 type Ovs struct{}
 
+// NMState is a declarative node network configuration driven through Kubernetes API
 // +k8s:openapi-gen=true
 type NMState struct{}
 
+// KubeMacPool plugin manages MAC allocation to Pods and VMs in Kubernetes
 // +k8s:openapi-gen=true
 type KubeMacPool struct {
+	// RangeStart defines the first mac in range
 	RangeStart string `json:"rangeStart,omitempty"`
+	// RangeEnd defines the last mac in range
 	RangeEnd   string `json:"rangeEnd,omitempty"`
 }
 
+// MacvtapCni plugin allows users to define Kubernetes networks on top of existing host interfaces
 // +k8s:openapi-gen=true
 type MacvtapCni struct{}
 
@@ -79,8 +94,7 @@ type Container struct {
 }
 
 // NetworkAddonsConfig is the Schema for the networkaddonsconfigs API
-// This struct is no exposed/registered as part of the CRD, but is used
-// by the v1alpha1 and v1 as kind of inside-helper struct
+// This struct is no exposed/registered as part of the CRD, but is used by the v1alpha1 and v1 as kind of inside-helper struct
 type NetworkAddonsConfig struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
--- a/pkg/components/components.go
+++ b/pkg/components/components.go
@@ -374,6 +374,20 @@ func GetClusterRole() *rbacv1.ClusterRole {
 
 func GetCrd() *extv1.CustomResourceDefinition {
 	subResouceSchema := &extv1.CustomResourceSubresources{Status: &extv1.CustomResourceSubresourceStatus{}}
+	placementProps := map[string]extv1.JSONSchemaProps{
+		"NodeSelector" : extv1.JSONSchemaProps{
+			Type:        "object",
+		},
+		"Affinity" : extv1.JSONSchemaProps{
+			Description: "Affinity is a group of affinity scheduling rules.",
+			Type:        "object",
+		},
+		"Tolerations" : extv1.JSONSchemaProps{
+			Description: "The pod this Toleration is attached to tolerates any taint that matches the triple <key,value,effect> using the matching operator <operator>.",
+			Type:        "object",
+		},
+	}
+
 	validationSchema := &extv1.CustomResourceValidation{
 		OpenAPIV3Schema: &extv1.JSONSchemaProps{
 			Description: "NetworkAddonsConfig is the Schema for the networkaddonsconfigs API",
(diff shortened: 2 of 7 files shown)


RamLavi:
output CRD:
```
---
apiVersion: apiextensions.k8s.io/v1beta1
kind: CustomResourceDefinition
metadata:
  name: networkaddonsconfigs.networkaddonsoperator.network.kubevirt.io
spec:
  group: networkaddonsoperator.network.kubevirt.io
  names:
    kind: NetworkAddonsConfig
    listKind: NetworkAddonsConfigList
    plural: networkaddonsconfigs
    singular: networkaddonsconfig
  scope: Cluster
  subresources:
    status: {}
  validation:
    openAPIV3Schema:
      description: NetworkAddonsConfig is the Schema for the networkaddonsconfigs
        API
      properties:
        apiVersion:
          description: 'APIVersion defines the versioned schema of this representation
            of an object. Servers should convert recognized schemas to the latest
            internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
          type: string
        kind:
          description: 'Kind is a string value representing the REST resource this
            object represents. Servers may infer this from the endpoint the client
            submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
          type: string
        metadata:
          type: object
        spec:
          description: NetworkAddonsConfigSpec defines the desired state of NetworkAddonsConfig
          properties:
            imagePullPolicy:
              description: PullPolicy describes a policy for if/when to pull a container
                image
              type: string
            kubeMacPool:
              description: Plugin that manages MAC allocation to Pods and VMs in Kubernetes
              properties:
                rangeEnd:
                  description: Defines the first mac in range
                  type: string
                rangeStart:
                  description: Defines the last mac in range
                  type: string
              type: object
            linuxBridge:
              description: Plugin that allows users to create a bridge and add the
                host and the container to it
              type: object
            macvtap:
              description: Plugin that allows users to define Kubernetes networks
                on top of existing host interfaces
              type: object
            multus:
              description: Plugin that enables attaching multiple network interfaces
                to Pods in Kubernetes
              type: object
            nmstate:
              description: Declarative node network configuration driven through Kubernetes
                API
              type: object
            ovs:
              description: Plugin that allows users to define Kubernetes networks
                on top of Open vSwitch bridges available on nodes
              type: object
            selfSignConfiguration:
              properties:
                caOverlapInterval:
                  description: Defines the duration of CA Certificates at CABundle
                    if not set it will default to CARotateInterval
                  type: string
                caRotateInterval:
                  description: Defines duration for CA and certificate
                  type: string
                certRotateInterval:
                  description: Defines duration for of service certificate
                  type: string
              type: object
          type: object
        status:
          description: NetworkAddonsConfigStatus defines the observed state of NetworkAddonsConfig
          properties:
            conditions:
              items:
                description: Condition represents the state of the operator's reconciliation
                  functionality
                properties:
                  lastHeartbeatTime:
                    format: date-time
                    type: string
                  lastTransitionTime:
                    format: date-time
                    type: string
                  message:
                    type: string
                  reason:
                    type: string
                  status:
                    type: string
                  type:
                    description: ConditionType is the state of the operator's reconciliation
                      functionality
                    type: string
                required:
                - status
                - type
                type: object
              type: array
            containers:
              items:
                properties:
                  image:
                    type: string
                  name:
                    type: string
                  parentKind:
                    type: string
                  parentName:
                    type: string
                required:
                - image
                - name
                - parentKind
                - parentName
                type: object
              type: array
            observedVersion:
              type: string
            operatorVersion:
              type: string
            targetVersion:
              type: string
          type: object
      type: object
  versions:
  - name: v1
    served: true
    storage: true
  - name: v1alpha1
    served: true
    storage: false

```

RamLavi:
getting `Failed calling status Set 8/10: Failed to update NetworkAddonsConfig "cluster" Status: networkaddonsconfigs.networkaddonsoperator.network.kubevirt.io "cluster" not found` although :
```
kubectl get networkaddonsconfigs
NAME      AGE
cluster   60s
```

pkg/components/components.go
```diff
@@ -368,63 +368,198 @@ func GetClusterRole() *rbacv1.ClusterRole {
 	return role
 }
 
-func GetCrd() *extv1beta1.CustomResourceDefinition {
-	crd := &extv1beta1.CustomResourceDefinition{
+func GetCrd() *extv1.CustomResourceDefinition {
```
qinqon, line 371: Don't we want to annotate the _types with descriptions and generate this with gen-openapi ?

qinqon (changes requested):

pkg/components/components.go
RamLavi: the existence of the `/shared` folder in types cause an error with `operator-sdk generate crds` command.
so, I created  it in an older version for reference and then recreated it in components.go.
the resulting .crd file is identical (except for the version migration changes of course)

RamLavi:
updated crd and explain [outputs](https://github.com/kubevirt/cluster-network-addons-operator/pull/524#issuecomment-685785353)

pkg/apis/networkaddonsoperator/shared/networkaddonsconfig_types.go
```diff
@@ -21,15 +21,22 @@ type NetworkAddonsConfigSpec struct {
...
 	CertRotateInterval string `json:"certRotateInterval,omitempty"`
 }
 
 // +k8s:openapi-gen=true
+// PlacementConfiguration defines node placement configuration
 type PlacementConfiguration struct {
+	// Infra defines placement configuration for master nodes
 	Infra     *Placement `json:"infra,omitempty"`
+	// Workloads defines placement configuration for worker nodes
 	Workloads *Placement `json:"workloads,omitempty"`
```
qinqon, line 40: Placement struc is missing.

pkg/components/components.go
```diff
@@ -372,63 +372,215 @@ func GetClusterRole() *rbacv1.ClusterRole {
...
+	validationSchema := &extv1.CustomResourceValidation{
```
qinqon, line 376: Add we sure we cannot store at the ontainer the stuff generated from operator-sdk and read it as golang struct ?

qinqon (changes requested):

pkg/components/components.go
RamLavi: we currently can't generate it, it doesn't know how to handle the shared folder:
```
cluster-network-addons-operator (add_csv_schema_validation_detail) $ ./build/_output/bin/operator-sdk generate crds
INFO[0000] Running CRD generator.                       

github.com/kubevirt/cluster-network-addons-operator/pkg/apis/networkaddonsoperator/shared:-: CRD for NetworkAddonsConfig.networkaddonsoperator.network.kubevirt.io has no storage version
FATA[0001] error generating CRDs from APIs in pkg/apis: error generating CRD manifests: error running CRD generator: not all generators ran successfully 
```

plus, (correct me if I'm wrong, but) marshaling the automated yaml to  struct will require us to always know the exact struct format. so in fact it could mean taking up the same amount of time.

qinqon: Well it's all about duplicity I see the description on structs and on the Schema, It has to be doable somehow we will se in the future if we do deeper integration of cnao with operator-sdk.

pkg/apis/networkaddonsoperator/shared/networkaddonsconfig_types.go
RamLavi: Done

RamLavi:
updated [crd output](https://github.com/kubevirt/cluster-network-addons-operator/pull/524#issuecomment-685785353)
changed in this [push](https://github.com/kubevirt/cluster-network-addons-operator/compare/173867cef595fef9cba3855039ed362da35e5613..5b3a3509af47be583536e9b5c452095643fbb5cc)

test/e2e/lifecycle/upgrade_test.go
```diff
@@ -32,6 +33,7 @@ var _ = Context("Cluster Network Addons Operator", func() {
 				BeforeEach(func() {
 					UninstallRelease(oldRelease)
 					InstallRelease(newRelease)
+					CheckCrdExplainable()
```
phoracek, line 36: Would it make sense to keep the second one in `It`? Since it is really something we want to test, not a setup of the context

test/e2e/lifecycle/upgrade_test.go
```diff
@@ -18,6 +18,7 @@ var _ = Context("Cluster Network Addons Operator", func() {
 		Context(fmt.Sprintf("when operator in version %s is installed and supported spec configured", oldRelease.Version), func() {
 			BeforeEach(func() {
 				InstallRelease(oldRelease)
+				CheckCrdExplainable()
```
phoracek, line 21: Why do we care about this in before each?

RamLavi:
getting validation error:
```
Failed to update NetworkAddonsConfig "cluster" Status: NetworkAddonsConfig.networkaddonsoperator.network.kubevirt.io "cluster" is invalid: status.conditions.lastHeartbeatTime: Invalid value: "null": status.conditions.lastHeartbeatTime in body must be of type string: "null"
```

test/e2e/lifecycle/upgrade_test.go
RamLavi: the idea was to check before and after the upgrade. 
I agree this would better suited in a specific `It`. tell me what you think
Done

RamLavi:
/hold

@qinqon is working on a PR to auto generate the crd via operator-sdk

RamLavi:
both failed test don't recreate locally, retesting

/retest

qinqon:
/hold cancel

We are not going to depend on auto generated CRDs since looks like we have some non tooling issues there.

RamLavi:
crd explain fails since `kubectl explain networkaddonsconfigs` stays <empty> after upgrade of crd.
When running after crd creation (not upgraded from old version) - explain looks OK.
tried to change upgrade to use `--force`,`--prune` flags but they don't refresh  the explain value.

RamLavi:
/hold

awaiting response from https://github.com/kubernetes/kubernetes/issues/95702 before merging

RamLavi:
/hold cancel

since currently the parameter is not deprecated we can move on with the PR

PR #526, opened 2020-09-03
Add PlacementConfiguration to kubeMacPool

rhrazdil:
Add Infra Placement configuration to kubeMacPool taken from NetworkAddonsConfig CR.

Adding Helm chartutil dependency which provides toYaml method for
expanding affinity/tolerations/nodeSelector structures in the
go templates.

Depends on https://github.com/kubevirt/cluster-network-addons-operator/pull/520

Signed-off-by: Radim Hrazdil <rhrazdil@redhat.com>

**What this PR does / why we need it**:

**Special notes for your reviewer**:

**Release note**:

```release-note
NONE
```

PR #520, opened 2020-08-31
Introduce PlacementConfiguration API

rhrazdil:
The intent of this PR is to introduce an API to allow separation of infra and workload components.
For example, `cluster-network-addons-operator` can be scheduled on infrastructure nodes instead of worker nodes.

The API mimicks changes introduced in https://github.com/kubevirt/hyperconverged-cluster-operator/pull/718
by adding 

spec.infra.affinity
spec.infra.nodeSelector
spec.infra.tolerations
spec.workloads.affinity
spec.workloads.nodeSelector
spec.workloads.tolerations
for affinity, nodeSelector and tolerations as corev1.Affinity,
map[string]string and []corev1.Toleration objects on CNAO CR

__Infra components:__
cluster-network-addons-operator
kubemacpool-mac-controller-manager
nmstate-webhook

__Workload components__:
bridge-marker
kube-cni-linux-bridge-plugin
macvtap-cni
multus
nmstate-handler
ovs-cni-amd64

By default `infra` components are scheduled on master nodes and `workload` components are scheduled on worker nodes using nodeAffinity with matchExpression `node-role.kubernetes.io/master` or `node-role.kubernetes.io/worker` respectively.

Signed-off-by: Radim Hrazdil <rhrazdil@redhat.com>

**What this PR does / why we need it**:
Adding API for configuring CNAO components placement into infra and workload categories.

**Special notes for your reviewer**:

**Release note**:

```release-note
Added API for PlacementCondiguration of Infra and Workloads components
```

PR #531, opened 2020-09-03
Add placement configuration to bridge-marker

RamLavi:
**What this PR does / why we need it**:
Use Infra PlacementConfiguration from NetworkAddonsConfig in bridge-marker
By default, bridge-marker will be scheduled on master nodes.

**Special notes for your reviewer**:
This PR is dependent of PRs: #520 #532

**Release note**:

```release-note
NONE
```

PR #532, opened 2020-09-03
Add PlacementConfiguration for linux-bridge

rhrazdil:
Use Workloads PlacementConfiguration from NetworkAddonsConfig in linux-bridge
By default, linux-bridge pods will be scheduled on all nodes.

This PR depends on : #520 #526

```release-note
NONE
```

### DIFF
--- a/pkg/network/network_test.go
+++ b/pkg/network/network_test.go
@@ -88,7 +88,7 @@ var _ = Describe("Testing network", func() {
 
 	Describe("Render", func() {
 		Context("when given valid arguments", func() {
-			conf := &cnao.NetworkAddonsConfigSpec{ImagePullPolicy: v1.PullAlways, Multus: &cnao.Multus{}, LinuxBridge: &cnao.LinuxBridge{}}
+			conf := &cnao.NetworkAddonsConfigSpec{ImagePullPolicy: v1.PullAlways, Multus: &cnao.Multus{}, LinuxBridge: &cnao.LinuxBridge{}, PlacementConfiguration: &cnao.PlacementConfiguration{Workloads: &cnao.Placement{}}}
 			manifestDir := "../../data"
 			openshiftNetworkConf := &osv1.Network{}
 			clusterInfo := &ClusterInfo{SCCAvailable: true, OpenShift4: false}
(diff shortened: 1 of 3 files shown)


RamLavi:
@rhrazdil dco fails meaning your  commit is not signed
you need to ammend the commit with `-s` flag or just add the `Signed-off-by: Radim Hrazdil <your-redhat-mail@redhat.com>` text manually

RamLavi:
I think the travis could  be solved by [this](https://github.com/kubevirt/cluster-network-addons-operator/pull/531/commits/649b37abae72e2417e8b0d1ebcf002487088755f) (I encountered it on the bridge marker PR since I'm rebased on top of your PR)

pkg/network/network_test.go
```diff
@@ -88,7 +88,7 @@ var _ = Describe("Testing network", func() {
 
 	Describe("Render", func() {
 		Context("when given valid arguments", func() {
-			conf := &cnao.NetworkAddonsConfigSpec{ImagePullPolicy: v1.PullAlways, Multus: &cnao.Multus{}, LinuxBridge: &cnao.LinuxBridge{}}
+			conf := &cnao.NetworkAddonsConfigSpec{ImagePullPolicy: v1.PullAlways, Multus: &cnao.Multus{}, LinuxBridge: &cnao.LinuxBridge{}, PlacementConfiguration: &cnao.PlacementConfiguration{Workloads: &cnao.Placement{}}}
```
phoracek, line 91: Why do we need to add that? The conf should have `valid arguments`. It should be valid even without Placement

RamLavi: it is a change I suggested so taking the liberty to answer :) 

the reason is that since we are using `LinuxBridge` then `renderLinuxBridge()` is called and that needs `PlacementConfiguration` to be configured or code will fail [here](https://github.com/kubevirt/cluster-network-addons-operator/pull/532/files#diff-05bcdfcb4d7a09161593a0118efd2a5dR33):
```
data.Data["Placement"] = conf.PlacementConfiguration.Workloads
```
btw this is the same for `ImagePullPolicy` [here](https://github.com/kubevirt/cluster-network-addons-operator/pull/532/files#diff-05bcdfcb4d7a09161593a0118efd2a5dR26)
 ``
data.Data["ImagePullPolicy"] = conf.ImagePullPolicy`
```

phoracek: Understood, closing my eyes

PR #537, opened 2020-09-09
bump CustomResourceDefinition to v1

RamLavi:
this PR bumps apiextensions.k8s.io to v1.
v1beta1 will be [deprecated](https://github.com/kubernetes/kubernetes/issues/82022) in k8s v1.19, and remove by 1.22
Also, it's needed for #524 to work

- **bump CustomResourceDefinition to v1**
in order to run "kubectl explain crd"
we need to bump CustomResourceDefinition to v1.
Part of this change is to move the deprecated spec.validation and Subresources.status fields
to be per version in spec.versions[].schema

- **Add minimal crd validation schema**
In order for the crd to be validated,
we need to add the minimal schema, including
the cnao components.

**Release note**:
This PR  depends on #566, #570 to be merged
```release-note
bump CustomResourceDefinition to v1
```

### DIFF
--- a/pkg/components/components.go
+++ b/pkg/components/components.go
@@ -7,7 +7,7 @@ import (
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
-	extv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
+	extv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	cnao "github.com/kubevirt/cluster-network-addons-operator/pkg/apis/networkaddonsoperator/shared"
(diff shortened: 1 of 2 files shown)


test/releases/0.16.0.go
```diff
@@ -59,6 +59,7 @@ func init() {
 			Ovs:         &cnao.Ovs{},
 		},
 		Manifests: []string{
+			"network-addons-config.crd.yaml",
```
qinqon, line 62: Looks like manifests added here will be deleted/created instead of applied, this means that the NetworkAddonsConfig will be deleted, we need to find a way to apply CRD instead of re-create, there is different slice for that.

RamLavi: >Looks like manifests added here will be deleted/created instead of applied, this means that the NetworkAddonsConfig will be deleted

can you elaborate on why is uninstall+install (re-create) not good?

>we need to find a way to apply CRD instead of re-create, there is different slice for that

I don't think we have that, do you want me to add one?

qinqon: we don't want the previously populated NetworkAddonsConfig to be deleted we want to our new CRD should be compatible if we delete the CRD the networkaddonsconfig get deleted too.

Let's ask @phoracek about where to add that, if not we need to add though.

phoracek: As Quique said, NAC has to be kept so we don't trigger the garbage collection of all components set to it. One of the main points of the upgrade tests is that we keep running workload.

We are now removing and re-creating the CNAO deployment, but we may change it so it is simply updated and the Deployment controller upgrades it.

phoracek: So maybe we just need to change the function running the upgrade to **not** delete and create, but only apply the manifests listed here.

RamLavi: ack, will also add `CheckCustomResourceDefinitionIsReady` to avoid 404 Errors..

RamLavi:
/hold

config is not becoming ready in upgrade tests

PR #538, opened 2020-09-10
change kubemacpool vms to opt-in

RamLavi:
Currently we are using the default opt-mode
for vms and pods in release configuration of kubemacpool repo.
Specifically, vms was set to opt-out.

Since we didn't have the chance to test this
configuration enough, we want to return the vms to opt-in

**Release note**:

```release-note
NONE
```

### DIFF
--- a/hack/components/bump-kubemacpool.sh
+++ b/hack/components/bump-kubemacpool.sh
@@ -85,8 +85,11 @@ EOF
 
     (
         cd config/cnao
-        cp ../release/mutatepods_opt_mode_patch.yaml .
-        cp ../release/mutatevirtualmachines_opt_mode_patch.yaml .
+
+        echo setting pods to opt-in mode
+        cp ../default/mutatepods_opt_in_patch.yaml mutatepods_opt_mode_patch.yaml
+        echo setting vms to opt-in mode
+        cp ../default/mutatevirtualmachines_opt_in_patch.yaml mutatevirtualmachines_opt_mode_patch.yaml
     )
 
 )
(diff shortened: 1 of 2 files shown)


phoracek (approved):
Clean and simple!

/lgtm
/approve

hack/components/bump-kubemacpool.sh
```diff
@@ -85,6 +85,12 @@ EOF
...
     (
         cd config/cnao
+
+        echo setting pods to opt-in mode
+        cp ../default/mutatepods_opt_in_patch.yaml ../release/mutatepods_opt_mode_patch.yaml
+        echo setting vms to opt-in mode
+        cp ../default/mutatevirtualmachines_opt_in_patch.yaml ../release/mutatevirtualmachines_opt_mode_patch.yaml
+
         cp ../release/mutatepods_opt_mode_patch.yaml .
         cp ../release/mutatevirtualmachines_opt_mode_patch.yaml .
```
qinqon, line 95: We can just do 
```bash
        echo setting pods to opt-in mode
        cp ../default/mutatepods_opt_in_patch.yaml mutatepods_opt_mode_patch.yaml
        echo setting vms to opt-in mode
        cp ../default/mutatevirtualmachines_opt_in_patch.yaml mutatevirtualmachines_opt_mode_patch.yaml
```

qinqon (changes requested):

hack/components/bump-kubemacpool.sh
RamLavi: yeah sure..

the reason it's like that is that I wanted to keep the same as the kubemacpool README instructions.. 
but you're right it's just redundant

PR #546, opened 2020-09-14
test: Ensure NAC does not exists eventually after delete

qinqon:
After calling .Delete on NetworkAddonsConfig it takes some time to be really deleted
this add a Eventually after call delete that check if NAC has being really deleted so
we don't pollute other tests.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/test/operations/operations.go
+++ b/test/operations/operations.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"strings"
+	"time"
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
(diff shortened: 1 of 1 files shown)


test/operations/operations.go
```diff
@@ -68,7 +69,13 @@ func DeleteConfig() {
 	}
 
 	err := framework.Global.Client.Delete(context.TODO(), config)
-	Expect(err).NotTo(HaveOccurred(), "Failed to remove the Config")
+	ExpectWithOffset(1, err).NotTo(HaveOccurred(), "Failed to remove the Config")
+
+	// Wait until the namespaces are terminated
```
alonSadan, line 74: you mean config is terminated?

alonSadan (commented):
Great!
one small comment

test/operations/operations.go
qinqon: Done

PR #551, opened 2020-09-15
Check cnao pod restarts in CI tests

RamLavi:
add check if CNAO pod performs unexpected restart during the CI tests.

- **add check that cnao pod has not restarted**
Currently there is no chck in the tests that
the CNAO pod has gone through a restart.
Let add a check to catch those restarts as soon as they occur.

- **add cnao pod stability check ci tests**
We want to make sure that CNAO pod does not perform
unwanted restart.
To do that we add a consistency check at the end of workflow
test suite to make sure no restart have been made
during the tests.
Also , after each test we want to add a print to signal
where the pod started restarting

- **TEMP COMMIT temporarily induce the cnao pod to restart.**
the Exit was added in order to be induced in test It("should remain at Available condition")

**Release note**:

```release-note
NONE
```

### DIFF
--- a/test/check/check.go
+++ b/test/check/check.go
@@ -353,6 +353,47 @@ func checkForDeployment(name string) error {
 	return nil
 }
 
+// CheckOperatorPodStability makes sure that the CNAO pod has not restarted since it started working
+func CheckOperatorPodStability(continuesDuration time.Duration) {
+	By("Checking that cnao operator pod hasn't performed any resets")
+	if continuesDuration != CheckImmediately {
+		Consistently( func() error {
+				return CalculateOperatorPodStability()
+			}, continuesDuration, time.Second).ShouldNot(HaveOccurred(), "CNAO operator pod should not restart consistently")
+	} else {
+		Expect(CalculateOperatorPodStability()).ShouldNot(HaveOccurred(), "Operator is not ready")
+	}
+}
+
+func PrintOperatorPodStability() {
+	if err := CalculateOperatorPodStability(); err != nil {
+		fmt.Fprintln(GinkgoWriter, "WARNING: CNAO operator pod is not stable: " + err.Error())
+	}
+	return
+}
+
+func CalculateOperatorPodStability() error {
+	pods := corev1.PodList{}
+	listOptions := []client.ListOption{
+		client.MatchingLabels(map[string]string{"name": components.Name}),
+		client.InNamespace(components.Namespace),
+	}
+
+	err := framework.Global.Client.List(context.Background(), &pods, listOptions...)
+	Expect(err).NotTo(HaveOccurred(), "should succeed getting the cnao pod")
+
+	if len(pods.Items) != 1 {
+		return fmt.Errorf("cnao operator pod should only have 1 replica")
+	}
+
+	for _, containerStatus := range pods.Items[0].Status.ContainerStatuses {
+		if containerStatus.RestartCount != 0 {
+			return fmt.Errorf("cnao operator pod restarted %d time since created", containerStatus.RestartCount)
+		}
+	}
+	return nil
+}
+
 func checkForDaemonSet(name string) error {
 	daemonSet := appsv1.DaemonSet{}
 
--- a/test/e2e/workflow/main_test.go
+++ b/test/e2e/workflow/main_test.go
@@ -2,6 +2,7 @@ package test
 
 import (
 	"context"
+	"time"
 	"testing"
 
 	. "github.com/onsi/ginkgo"
@@ -47,7 +48,12 @@ var _ = BeforeSuite(func() {
 	Expect(err).ToNot(HaveOccurred())
 })
 
+var _ = AfterSuite(func() {
+	CheckOperatorPodStability(time.Minute)
+})
+
 var _ = AfterEach(func() {
+	PrintOperatorPodStability()
 	By("Performing cleanup")
 	if GetConfig() != nil {
 		DeleteConfig()


test/check/check.go
```diff
@@ -353,6 +353,26 @@ func checkForDeployment(name string) error {
...
+	pods := corev1.PodList{}
+	listOptions := client.ListOptions{}
+	labelSelector, err := k8slabels.Parse(fmt.Sprintf("name, name == %s", components.Name))
+	Expect(err).NotTo(HaveOccurred(), "should succeed parsing label selector")
+	listOptions.LabelSelector = labelSelector
+	listOptions.Namespace = components.Namespace
+	err = framework.Global.Client.List(context.Background(), &pods, &listOptions)
+	Expect(err).NotTo(HaveOccurred(), "should succeed getting the cnao pod")
+
+	Expect(len(pods.Items)).To(BeNumerically("==", 1), fmt.Sprintf("should only have 1 cnao pod"))
```
alonSadan, line 367: If we know that there is only one pod, why not just get  the CNAO pod,  instead of get a list of pods?

alonSadan (commented):
awesome.
one question added.

Also, can you explain why you chose to add the restart check specifically were you placed it, like what's the rational behind it?

p.s: 
I think you have a typo in the commit message ANSO <->CNAO?

test/check/check.go
RamLavi: Done

test/check/check.go
```diff
@@ -353,6 +353,26 @@ func checkForDeployment(name string) error {
 	return nil
 }
 
+// CheckOperatorPodStability makes sure that the CNAO pod has not restarted since it started working
+func CheckOperatorPodStability() {
+	pods := corev1.PodList{}
+	listOptions := client.ListOptions{}
+	labelSelector, err := k8slabels.Parse(fmt.Sprintf("name, name == %s", components.Name))
+	Expect(err).NotTo(HaveOccurred(), "should succeed parsing label selector")
+	listOptions.LabelSelector = labelSelector
```
qinqon, line 362: Selecting labels with controller-runtime is easier
```golang
import "sigs.k8s.io/controller-runtime/pkg/client"
List(ontext.Background(), &pods, &client.MatchingLabels(map[string]string{})) 		
```

test/check/check.go
qinqon: We don't have the pod name since it's generated so we cannot use get on it we have to list and filter.

test/check/check.go
```diff
@@ -353,6 +353,26 @@ func checkForDeployment(name string) error {
...
+	listOptions.Namespace = components.Namespace
+	err = framework.Global.Client.List(context.Background(), &pods, &listOptions)
+	Expect(err).NotTo(HaveOccurred(), "should succeed getting the cnao pod")
+
+	Expect(len(pods.Items)).To(BeNumerically("==", 1), fmt.Sprintf("should only have 1 cnao pod"))
+
+	for _, pod := range pods.Items {
```
qinqon: we have only one pod no need to iterate.

qinqon (changes requested):

RamLavi:
> Also, can you explain why you chose to add the restart check specifically were you placed it, like what's the rational behind it?
Added to commit message

> 
> p.s:
> I think you have a typo in the commit message ANSO <->CNAO?
DONE

test/check/check.go
RamLavi: but Alon is right in the manner that we don't need to iterate over a pod list with len 1. fixed below

test/check/check.go
RamLavi: I think `frameworkClient` List needs ListOptions, while : 
```
func (f *frameworkClient) List(gCtx goctx.Context, list runtime.Object, opts ...dynclient.ListOption)
```

test/check/check.go
RamLavi: Done

test/check/check.go
qinqon: ListOptions is an interface and MatchingLabels implement it https://github.com/kubernetes-sigs/controller-runtime/blob/e63948755ba14b3dd8506fafbef4065fb9457817/pkg/client/options.go#L41

RamLavi:
`AfterEach` print will look like  this:
```
WARNING: CNAO operator pod is not stable: cnao operator pod restarted x times since created
```

RamLavi:
test check caught induced fault : [prow](https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_cluster-network-addons-operator/551/pull-e2e-cluster-network-addons-operator-workflow-k8s/1305889753157079040#1:build-log.txt%3A918)

removing the temp commit

phoracek:
I really like how you approach this and that we are introducing this check. No more random restarts on production.

alonSadan (commented):
nice!
/lgtm

qinqon (changes requested):

pkg/network/network.go
```diff
@@ -182,6 +182,9 @@ func RenderObjsToRemove(prev, conf *cnao.NetworkAddonsConfigSpec, manifestDir st
 
 	if conf.NMState == nil {
 		o, err := renderNMState(prev, manifestDir, clusterInfo)
+		log.Print("!!!inducing Exit")
+		os.Exit(1)
```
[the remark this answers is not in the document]
RamLavi: Done

test/check/check.go
```diff
@@ -353,6 +353,47 @@ func checkForDeployment(name string) error {
 	return nil
 }
 
+// CheckOperatorPodStability makes sure that the CNAO pod has not restarted since it started working
+func CheckOperatorPodStability(onlyPrint bool, continuesDuration time.Duration) {
```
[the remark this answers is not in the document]
RamLavi: Done. I left `CalculateOperatorPodStability` to return error so that I can see exactly how many times it has restarted.

qinqon:
@RamLavi I don't see the changes

test/check/check.go
```diff
@@ -353,6 +353,47 @@ func checkForDeployment(name string) error {
...
+	By("checking that cnao operator pod hasn't performed any resets")
+	if onlyPrint {
+		if err := CheckForOperatorPod(); err != nil {
+			fmt.Printf("WARNING: CNAO operator pod is not stable: %s\n", err)
```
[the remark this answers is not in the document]
RamLavi: Done

PR #566, opened 2020-09-22
Change config to use unstructured

RamLavi:
This PR is changes config operations to support both v1, v1alpha1 NetworkAddonsConfig

- **refactor operations to accept unstructured**
Currently, functions that create/update/remove NetworkAddonsConfig type
only use the currently stored NetworkAddonsConfig version:v1
This is not an issue during upgrade tests since we're not upgrading
the crd itself, so we always support v1.
Let's support both APIs by using unstructred objects
when calling k8s client.

- **update e2e tests to use new config changes**
this commit introduces the change in config API to
all the relevant e2e tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/test/operations/operations.go
+++ b/test/operations/operations.go
@@ -8,74 +8,123 @@ import (
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+
 	framework "github.com/operator-framework/operator-sdk/pkg/test"
 	"gopkg.in/yaml.v2"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 
 	cnao "github.com/kubevirt/cluster-network-addons-operator/pkg/apis/networkaddonsoperator/shared"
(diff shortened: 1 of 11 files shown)


test/operations/operations.go
```diff
@@ -8,74 +8,117 @@ import (
...
+	cnaoV1alpha1 := cnaov1.NetworkAddonsConfig{}
+	switch gvk {
+	case cnaoV1.GroupVersionKind():
+		return &convertToConfigV1(config).Status
+	case cnaoV1alpha1.GroupVersionKind():
+		return &convertToConfigV1alpha1(config).Status
+	}
+
+	Fail(fmt.Sprintf("gvk %v not supported", gvk))
+	return nil
```
qinqon, line 97: We cannot retrieve status from unstructured ?

RamLavi: `framework.Global.Client.Status().` doesn't support `Get`,  only `Update\Patch`.
I did not find a way to get the status object out of the struct

test/operations/operations.go
```diff
@@ -8,74 +8,123 @@ import (
...
-			Name: names.OPERATOR_CONFIG,
-		},
-	}
+	config := &unstructured.Unstructured{}
+	config.SetGroupVersionKind(gvk)
+	config.SetName(names.OPERATOR_CONFIG)
 
 	err := framework.Global.Client.Delete(context.TODO(), config)
-	ExpectWithOffset(1, err).NotTo(HaveOccurred(), "Failed to remove the Config")
+	Expect(err).NotTo(HaveOccurred(), "Failed to remove the Config")
```
qinqon, line 73: This is not related.

qinqon (changes requested):
Just one nit

test/operations/operations.go
RamLavi: Done

PR #565, opened 2020-09-22
bump kubemacpool to v0.20.2

RamLavi:
**What this PR does / why we need it**:
This PR bumps kubemacpool to v0.20.2

**Special notes for your reviewer**:

**Release note**:

```release-note
bump kubemacpool to v0.20.2
```

PR #567, opened 2020-09-23
skip KMP opt-mode tests not maching CNAO config

RamLavi:
This PR add a skip label to the tier1 KMP tests 
to avoid flaky tests on irrelevant opt mode Context.

- **kmp functests, skip KMP opt-mode tests not maching CNAO config**
KMP functests include tests for both opt-in and opt-out KMP webhook modes
and since CNAO deploys KMP with opt-in mode,
it is necessary to skip tests of the opposite mode (opt-out tests).
This commit adds a function that checks what is the current configuration of
KMP in the cluster and skip the relevant tests accordingly.

dependent of #565 to pass
**Release note**:

```release-note
NONE
```

### DIFF
--- a/automation/check-patch.e2e-kubemacpool-functests.sh
+++ b/automation/check-patch.e2e-kubemacpool-functests.sh
@@ -15,6 +15,7 @@ function __get_skipped_tests() {
 
 	# We can't test all KMP opt-mode CNAO context, as the operator will reconcile
 	# back to the configured opt-mode when the test tries to change it.
+	# So we check KMP webhook's opt-mode and skip tests accordingly
 	if grep 'default/mutatevirtualmachines_opt_out_patch.yaml' hack/components/bump-kubemacpool.sh; then
 		echo "KMP VM webhook is set to opt-out mode. Skipping opt-in Context"
 		skipped_regex="${skipped_regex} \(opt-in\smode\)"
--- a/automation/check-patch.e2e-kubemacpool-functests.sh
+++ b/automation/check-patch.e2e-kubemacpool-functests.sh
@@ -9,6 +9,23 @@
 # yum -y install automation/check-patch.packages
 # automation/check-patch.e2e-kubemacpool-functests.sh
 
+function __get_skipped_tests() {
+	local  __resultvar=$1
+	local skipped_regex=""
+
+	# We can't test all KMP opt-mode CNAO context, as the operator will reconcile
+	# back to the configured opt-mode when the test tries to change it.
+	if grep 'default/mutatevirtualmachines_opt_out_patch.yaml' hack/components/bump-kubemacpool.sh; then
+		echo "KMP VM webhook is set to opt-out mode. Skipping opt-in Context"
+		skipped_regex="${skipped_regex} \(opt-in\smode\)"
+	elif grep 'default/mutatevirtualmachines_opt_in_patch.yaml' hack/components/bump-kubemacpool.sh; then
+		echo "KMP VM webhook is set to opt-in mode. Skipping opt-out Context"
+		skipped_regex="${skipped_regex} \(opt-out\smode\)"
+	fi
+
+	eval $__resultvar="'$skipped_regex'"
+}
+
 teardown() {
     make cluster-down
     rm -rf "${TMP_COMPONENT_PATH}"
@@ -28,9 +45,13 @@ main() {
     echo "Deploy KubeVirt latest stable release"
     ./hack/deploy-kubevirt.sh
 
+    echo "Get skip tests regex"
+    __get_skipped_tests SKIPPED_TESTS
+
     # Run KubeMacPool functional tests
     cd ${TMP_COMPONENT_PATH}
-    KUBECONFIG=${KUBECONFIG} E2E_TEST_ARGS="--junit-output=$ARTIFACTS/junit.functest.xml" make functest
+
+    KUBECONFIG=${KUBECONFIG} E2E_TEST_ARGS="--junit-output=$ARTIFACTS/junit.functest.xml --ginkgo.skip $SKIPPED_TESTS" make functest
 }
 
 [[ "${BASH_SOURCE[0]}" == "$0" ]] && main "$@"
--- a/pkg/components/components.go
+++ b/pkg/components/components.go
@@ -436,13 +436,13 @@ func GetCrd() *extv1.CustomResourceDefinition {
 									Type: "object",
 									Properties: map[string]extv1.JSONSchemaProps{
 										"lastHeartbeatTime": extv1.JSONSchemaProps{
-											Format: "date-time",
-											Type:   "object",
+											Format:   "date-time",
+											Type:     "object",
 											Nullable: true,
 										},
 										"lastTransitionTime": extv1.JSONSchemaProps{
-											Format: "date-time",
-											Type:   "object",
+											Format:   "date-time",
+											Type:     "object",
 											Nullable: true,
 										},
 										"message": extv1.JSONSchemaProps{


RamLavi:
/hold

waits on #565 to be merged

automation/check-patch.e2e-kubemacpool-functests.sh
```diff
@@ -28,9 +29,46 @@ main() {
...
+}
+
+function __get_skipped_tests() {
+	local skipped_regex=""
+
+	if [ "$(__is_opt_out)" == "true" ]; then
+		skipped_regex="${skipped_regex} \(opt-in\smode\)"
+	elif [ "$(__is_opt_in)" == "true" ]; then
+		skipped_regex="${skipped_regex} \(opt-out\smode\)"
+	fi
```
qinqon, line 49: skipped_regex is local and we have `if/elif` does not make sense to append her.

automation/check-patch.e2e-kubemacpool-functests.sh
```diff
@@ -28,9 +29,46 @@ main() {
...
+
+function __is_opt_in() {
+	if [ $(yaml-utils::read_multiple_docs data/kubemacpool/kubemacpool.yaml webhooks[1].namespaceSelector.matchExpressions[2].key) == 'mutatevirtualmachines.kubemacpool.io' ] && \
+	   [ $(yaml-utils::read_multiple_docs data/kubemacpool/kubemacpool.yaml webhooks[1].namespaceSelector.matchExpressions[2].operator) == 'In' ] && \
+	   [ $(yaml-utils::read_multiple_docs data/kubemacpool/kubemacpool.yaml webhooks[1].namespaceSelector.matchExpressions[2].values[0]) == 'allocate' ]; then
+	     	echo "true"
+	     	return
+	fi
+	echo "false"
+}
```
qinqon, line 62: What about just greping https://github.com/kubevirt/cluster-network-addons-operator/blob/master/hack/components/bump-kubemacpool.sh#L92 ?

`grep opt_in hack/components/bump-kubemacpool` so we don't need to know how it's done.

qinqon (changes requested):

automation/check-patch.e2e-kubemacpool-functests.sh
RamLavi: it's intentional, the idea was to create a general function `__get_skipped_tests` that adds regexes to skip. Currently it only adds the opt-mode but maybe later we'll want to concatenate others.
but if you prefer I can remove this part

automation/check-patch.e2e-kubemacpool-functests.sh
RamLavi: I like it! :)

automation/check-patch.e2e-kubemacpool-functests.sh
```diff
@@ -17,6 +17,7 @@ teardown() {
 main() {
     # Setup CNAO and artifacts temp directory
     source automation/check-patch.setup.sh
+    source hack/components/yaml-utils.sh
```
ormergi: The reason automation/check-patch.setup.sh is sourced at this place is because it do what the comment suggests and not acting just like a library, maybe changing `source` to `./` is a better fit.
In any case you can move this line to top of the file

ormergi (commented):
Hey about the second commit `add skipped test label to KMP tier1 tests` 
If I understand correctly on CNAO deployment, KMP set works at `opt-in` mode for allocating MACs for VMs, but KMP functests assume it works on `opt-out` mode.
So you check in what KMP mode works and skip tests based on that, is that right?

If so we could introduce a variable (e.g: `GINKGO_PARAMS`) and pass it to this script at[ this lane prow job yaml](https://github.com/kubevirt/project-infra/blob/master/github/ci/prow/files/jobs/kubevirt/cluster-network-addons-operator/cluster-network-addons-operator-presubmits.yaml#L159) until CNAO and KMP will run on the same mode.
Then we could run this script with skipping out-out test like this: 
`GINKGO_PARAMS="--ginkgo.skip=\(opt-in\smode\)"  automation/check-patch.e2e-kubemacpool-functests.sh`

This way we wont need add extra logic to this script, and we could control better how we run tests suite w/o change anything
What you think?

automation/check-patch.e2e-kubemacpool-functests.sh
qinqon: Fair enough

automation/check-patch.e2e-kubemacpool-functests.sh
RamLavi: I removed it entirely (this line and the first commit) thanks to @qinqon 's  great suggestion.

RamLavi:
> Hey about the second commit `add skipped test label to KMP tier1 tests`
> If I understand correctly on CNAO deployment, KMP set works at `opt-in` mode for allocating MACs for VMs, but KMP functests assume it works on `opt-out` mode.

that's actually the other way around. KMP repo is opt-out, and CNAO repo is opt-in..

> So you check in what KMP mode works and skip tests based on that, is that right?

yep

> If so we could introduce a variable (e.g: `GINKGO_PARAMS`) and pass it to this script at[ this lane prow job yaml](https://github.com/kubevirt/project-infra/blob/master/github/ci/prow/files/jobs/kubevirt/cluster-network-addons-operator/cluster-network-addons-operator-presubmits.yaml#L159) until CNAO and KMP will run on the same mode.
> Then we could run this script with skipping out-out test like this:
> `GINKGO_PARAMS="--ginkgo.skip=\(opt-in\smode\)" automation/check-patch.e2e-kubemacpool-functests.sh`
> 
> This way we wont need add extra logic to this script, and we could control better how we run tests suite w/o change anything
> What you think?

my problem with your suggestion is that it's static. my approach will always skip the unneeded opt-mode tests, no matter what the opt mode is  changed to (in/out)
Regarding to put it on [project-infra](https://github.com/kubevirt/project-infra/blob/master/github/ci/prow/files/jobs/kubevirt/cluster-network-addons-operator/cluster-network-addons-operator-presubmits.yaml#L159) I think the logic should be in CNAO.

ormergi:
> my problem with your suggestion is that it's static. my approach will always skip the unneeded opt-mode tests, no matter what the opt mode is changed to (in/out)

I am always up for generic code, in this case its about the configuration of KMP in CNAO and I think its better to manage this form "outside" instead of checking how KMP is configured on the cluster in run time and skip tests based on that.
If you want to stick with this approach I think it should be done KMP functests code (checking if KMP is configured with opt-it/out) 

> Regarding to putting it on [project-infra](https://github.com/kubevirt/project-infra/blob/master/github/ci/prow/files/jobs/kubevirt/cluster-network-addons-operator/cluster-network-addons-operator-presubmits.yaml#L159) I think the logic should be in CNAO. it's not general enough to put it there IMO

I meant to put this line `GINKGO_PARAMS="--ginkgo.skip=\(opt-in\smode\)" automation/check-patch.e2e-kubemacpool-functests.sh` in the project-infra job, and change the line that run test to `E2E_TEST_ARGS="... GINKGO_PARAMS" make functest`.
Since changing KMP mode in CNAO is not something we change often its good enough in my opinion.

RamLavi:
> > my problem with your suggestion is that it's static. my approach will always skip the unneeded opt-mode tests, no matter what the opt mode is changed to (in/out)
> 
> I am always up for generic code, in this case its about the configuration of KMP in CNAO and I think its better to manage this form "outside" instead of checking how KMP is configured on the cluster in run time and skip tests based on that.
> If you want to stick with this approach I think it should be done KMP functests code (checking if KMP is configured with opt-it/out)

I don't think we should put it on KMP repo: KMP repo tests both opt-in&optout Contexes - there, the webhook is manipulated to be opt-in/opt-out, so that all possibilities are always kept in check
The reason this is not possible to do the same in CNAO repo is that the operator reconcile doesn't allow us to change the webhook configuration (which is ok in general, but make testing both opt-modes impossible in CNAO).
For this reason, I think the code should be specific to CNAO's  KMP-tier1 script.
This way, CNAO always checks the configuration that it'll use (aka the release config).
 
> 
> > Regarding to putting it on [project-infra](https://github.com/kubevirt/project-infra/blob/master/github/ci/prow/files/jobs/kubevirt/cluster-network-addons-operator/cluster-network-addons-operator-presubmits.yaml#L159) I think the logic should be in CNAO. it's not general enough to put it there IMO
> 
> I meant to put this line `GINKGO_PARAMS="--ginkgo.skip=\(opt-in\smode\)" automation/check-patch.e2e-kubemacpool-functests.sh` in the project-infra job, and change the line that run test to `E2E_TEST_ARGS="... GINKGO_PARAMS" make functest`.
> Since changing KMP mode in CNAO is not something we change often its good enough in my opinion.

Putting it in project-infra will enforce us to change project-infra every time we change release opt-mode. 
keeping it in CNAO make's it easier to understand what went wrong if something changes (opt mode) or breaks (unknown future thing). I think it'll make our future selves an easier life debugging :)

automation/check-patch.e2e-kubemacpool-functests.sh
```diff
@@ -28,9 +28,33 @@ main() {
     echo "Deploy KubeVirt latest stable release"
     ./hack/deploy-kubevirt.sh
 
+    echo "Get skip tests regex"
```
ormergi, line 31: Please add a comment or change this message, to explain we need to skip tests that check the opposite KMP webhook mode

automation/check-patch.e2e-kubemacpool-functests.sh
```diff
@@ -28,9 +28,33 @@ main() {
...
+
+	echo $skipped_regex
+}
+
+function __is_vm_opt_in() {
+	grep 'default/mutatevirtualmachines_opt_in_patch.yaml' hack/components/bump-kubemacpool.sh
+}
+
+function __is_vm_opt_out() {
+	grep 'default/mutatevirtualmachines_opt_out_patch.yaml' hack/components/bump-kubemacpool.sh
```
ormergi, line 57: Could you please move this to code to another file so we could keep this script simple as possible?

ormergi:
> I don't think we should put it on KMP repo: KMP repo tests both opt-in&optout Contexes - there, the webhook is manipulated to be opt-in/opt-out, so that all possibilities are always kept in check
> The reason this is not possible to do the same in CNAO repo is that the operator reconcile doesn't allow us to change the webhook configuration (which is ok in general, but make testing both opt-modes impossible in CNAO).
> For this reason, I think the code should be specific to CNAO's KMP-tier1 script.
> This way, CNAO always checks the configuration that it'll use (aka the release config).
>

Got it thanks for explaining.
 
> Putting it in project-infra will enforce us to change project-infra every time we change release opt-mode.
> keeping it in CNAO keeps it dynamic, and make's it easier to understand what went wrong if something changes (opt mode) or breaks (unknown future thing). I think it'll make our future selves an easier life debugging :)

It is ok because those configurations shouldn't be change often if at all, and it we could save code to maintain.

I wont insist on it this way is also good, just that we should to move the code to another file to keep that script simple at possible 
I will continue my review :)

ormergi (changes requested):

automation/check-patch.e2e-kubemacpool-functests.sh
RamLavi: I made it  so that `__get_skipped_tests` is general, so I'll add a specific comment inside it
Done

automation/check-patch.e2e-kubemacpool-functests.sh
RamLavi: I was actually thinking about to move it inside the `__get_skipped_tests`.
It's simple enough and doesn't qualify a function of it's  own. + now with the comment it's well documented.

Tell me what you think

alonSadan (commented):
Clean and simple:)
/lgtm

automation/check-patch.e2e-kubemacpool-functests.sh
```diff
@@ -28,9 +28,27 @@ main() {
...
     cd ${TMP_COMPONENT_PATH}
-    KUBECONFIG=${KUBECONFIG} E2E_TEST_ARGS="--junit-output=$ARTIFACTS/junit.functest.xml" make functest
+
+    KUBECONFIG=${KUBECONFIG} E2E_TEST_ARGS="--junit-output=$ARTIFACTS/junit.functest.xml --ginkgo.skip $SKIPPED_TESTS" make functest
+}
+
+function __get_skipped_tests() {
+	local skipped_regex=""
+
+	if [ ! -z "$(grep 'default/mutatevirtualmachines_opt_in_patch.yaml' hack/components/bump-kubemacpool.sh)" ]; then
```
qinqon, line 43: this is checking opt-out to skip opt-in so 

```bash
if grep 'default/mutatevirtualmachines_opt_out_patch.yaml' hack/components/bump-kubemacpool.sh; then
```
```

File: automation/check-patch.e2e-kubemacpool-functests.sh [hunk 7]
```diff
@@ -28,9 +28,27 @@ main() {
...
+		echo "KMP VM webhook is set to opt-out mode. skipping opt-in Context"
+		skipped_regex="${skipped_regex} \(opt-in\smode\)"
+	elif [ ! -z "$(grep 'default/mutatevirtualmachines_opt_in_patch.yaml' hack/components/bump-kubemacpool.sh)" ]; then
```
qinqon, line 46: ditto

qinqon (changes requested):
The first check for filtering is wrong it has to check opt_out to skip opt-in.

automation/check-patch.e2e-kubemacpool-functests.sh
RamLavi: notice the `!`
I am actually asking: if `default/mutatevirtualmachines_opt_out_patch.yaml` is mentioned on `hack/components/bump-kubemacpool.sh` then...

automation/check-patch.e2e-kubemacpool-functests.sh
RamLavi: [same](https://github.com/kubevirt/cluster-network-addons-operator/pull/567#discussion_r493764717)

automation/check-patch.e2e-kubemacpool-functests.sh
```diff
@@ -28,9 +28,27 @@ main() {
...
+		echo "KMP VM webhook is set to opt-in mode. skipping opt-out Context"
+		skipped_regex="${skipped_regex} \(opt-out\smode\)"
+	fi
+
+	echo $skipped_regex
```
oshoval, line 51: the function return its value via print, but you have more prints, 
how come it will work ?
i will suggest to return a value by setting a global for example


another point that will do problem is that if someone add debug prints in the function, it will change the logic and the return value in its current state

automation/check-patch.e2e-kubemacpool-functests.sh
RamLavi: wow great catch!!
Done

automation/check-patch.e2e-kubemacpool-functests.sh
RamLavi: Done

automation/check-patch.e2e-kubemacpool-functests.sh
RamLavi: ok. tell me what you think of [this change](https://github.com/kubevirt/cluster-network-addons-operator/compare/266058b750e06c72c3a4cf48b64496aacb8cba05..c9034d2d0005f3c0a2955a6f92251232a9b5d8f0) then

oshoval: i would prefer to keep it simple please
eval is $%%^
looks like we looked on the same site https://www.linuxjournal.com/content/return-values-bash-functions

unless you insist not to use global i would suggest this

```
function myfunc()
{
    myresult='some value'
}

myfunc
echo $myresult
```

RamLavi: I don't want to use globals unless necessary. 
what non-global option would you suggest?

oshoval: btw i dont think on bash is such a bad idea,

anyhow i would suggest to remove the other echos, or if you want to echo them just make the function body non function and just embed it in the main file imo

automation/check-patch.e2e-kubemacpool-functests.sh
ormergi: Please add short explanation in this line why we need to skip test of the opposite KMP webhook mode, something like 
"Run only tests of the current configured KMP webhooks mode (opt-in/opt-out)"

automation/check-patch.e2e-kubemacpool-functests.sh
```diff
@@ -28,9 +28,28 @@ main() {
...
+    __get_skipped_tests SKIPPED_TESTS
+
     # Run KubeMacPool functional tests
     cd ${TMP_COMPONENT_PATH}
-    KUBECONFIG=${KUBECONFIG} E2E_TEST_ARGS="--junit-output=$ARTIFACTS/junit.functest.xml" make functest
+
+    KUBECONFIG=${KUBECONFIG} E2E_TEST_ARGS="--junit-output=$ARTIFACTS/junit.functest.xml --ginkgo.skip $SKIPPED_TESTS" make functest
+}
+
+function __get_skipped_tests() {
```
ormergi, line 40: If you dont mind would you move this code to another file? or at least move this to line 11 to align with other functions

ormergi (changes requested):

automation/check-patch.e2e-kubemacpool-functests.sh
RamLavi: I'm good with using eval. I want to keep this generic function, and seems like eval makes the best out of all worlds.

automation/check-patch.e2e-kubemacpool-functests.sh
RamLavi: I wouldn't want to move it to another file, as it's a very short function that will only be used here.. 
I'll move it to line 11, thanks!

automation/check-patch.e2e-kubemacpool-functests.sh
RamLavi: `__get_skipped_tests` is meant to be generic, so a generic comment is ok here imo.

ormergi: you could also add this link https://github.com/k8snetworkplumbingwg/kubemacpool#kubemacpool-service

automation/check-patch.e2e-kubemacpool-functests.sh
RamLavi: [Done](https://github.com/kubevirt/cluster-network-addons-operator/compare/c9034d2d0005f3c0a2955a6f92251232a9b5d8f0..985f5fc3e5fed1598b977d971dbad37ee43e0550)

automation/check-patch.e2e-kubemacpool-functests.sh
ormergi: Yes I saw that you add those echos its is great 
I think we should add comment about skipping tests, since its not straight forward to understand why we need to skip in this case
And its good reminder for our future self's why we did it :)

oshoval: i am against comments, the code is / should be self explanatory.
i wont block, its up to you two

RamLavi: >I think we should add comment about skipping tests, since its not straight forward to understand why we need to skip in this case
And its good reminder for our future self's why we did it :)

but why isn't the current echo's enough?

ormergi: @oshoval We discussed this few times now, I know what is you opinion about comments and generally I agree, but there are cases when you need more and explain why you need to do something
like in this case.

@RamLavi The echos explains what are we doing, I think we need to add comment about why we need to do this.
It took me a few minutes to understand why we need this and I am familiar with KMP code, for someone who doesn't it will be hard to understand it in my opinion.

RamLavi: @ormergi this is why the commit message is so important.
We can't explain everything in the code, but we make sure that anyone who want to further understand a change - could do so by reading the commit message (via blame or github GUI)

do you think the commit message needs further clarification?

ormergi: I understand, in this case it would be much better to explain shortly why is it necessary to skip specific tests instead of the current echo message.
Since this is a bash script its not documenting itself well like other languages.

For the commit message, I would change the subject to `kmp functests, skip redundant KMP webhook mode tests`
and for the body, 
```
KMP functests includes tests for both opt-in and opt-out KMP webhook modes
and since CNAO deploys KMP with opt-in mode,
it is necessary to skip tests of the opposite mode (opt-out tests).

This commit adds a function that checks what is the current configuration of KMP in the cluster
 and skip the relevant tests accordingly.
```

ormergi (changes requested):

automation/check-patch.e2e-kubemacpool-functests.sh
RamLavi: Done. 
redundant is not the correct wording, but change to fit

ormergi: Thanks

automation/check-patch.e2e-kubemacpool-functests.sh
```diff
@@ -9,6 +9,23 @@
 # yum -y install automation/check-patch.packages
 # automation/check-patch.e2e-kubemacpool-functests.sh
 
+function __get_skipped_tests() {
+	local  __resultvar=$1
+	local skipped_regex=""
+
+	# We can't test all KMP opt-mode CNAO context, as the operator will reconcile
```
ormergi, line 16: Hey what about this comment, I think you cant understand form this that KMP functests configures the opt-mode
could you change it to something like `check KMP webhook's opt-mode and skip tests accordingly`?

RamLavi: added your comment below mine.
Done

ormergi: Thanks

ormergi (approved):

PR #569, opened 2020-09-23
Change upgrade test

RamLavi:
This PR change the upgrade test to work more like OLM performs upgrade

- **change upgrade test to also update crd manifest**
Currently upgrade test only updates operator.yaml manifest,
while keeping the currently deployed crd.
This is not how OLM performs upgrade, and since we want to
test the actual upgrade scenario then we should also upgrade the crd

- **remove UninstallRelease from upgrade test**
Currently manifests are updated by
uninstalling the old and intalling the new.
Now that we are also uninstalling the crd,
we need to change this behavior to only update the manifests
We also keep operator tests while crd is not kept deployed.

This PR depends on #537 to be merged

**Release note**:

```release-note
NONE
```

### DIFF
--- a/test/e2e/lifecycle/upgrade_test.go
+++ b/test/e2e/lifecycle/upgrade_test.go
@@ -17,6 +17,7 @@ var _ = Context("Cluster Network Addons Operator", func() {
 	testUpgrade := func(oldRelease, newRelease Release) {
 		Context(fmt.Sprintf("when operator in version %s is installed and supported spec configured", oldRelease.Version), func() {
 			BeforeEach(func() {
+				UninstallRelease(newRelease)
 				oldReleaseGvk := GetCnaoV1alpha1GroupVersionKind()
 				InstallRelease(oldRelease)
 				CheckOperatorIsReady(podsDeploymentTimeout)
@@ -32,7 +33,6 @@ var _ = Context("Cluster Network Addons Operator", func() {
 			Context("and it is upgraded to the latest release", func() {
 				newReleaseGvk := GetCnaoV1GroupVersionKind()
 				BeforeEach(func() {
-					UninstallRelease(oldRelease)
 					InstallRelease(newRelease)
 					UpdateConfig(newReleaseGvk, newRelease.SupportedSpec)
 					CheckOperatorIsReady(podsDeploymentTimeout)
@@ -56,7 +56,6 @@ var _ = Context("Cluster Network Addons Operator", func() {
 			It(fmt.Sprintf("should transition reported versions while being upgraded to version %s", newRelease.Version), func() {
 				newReleaseGvk := GetCnaoV1GroupVersionKind()
 				// Upgrade the operator
-				UninstallRelease(oldRelease)
 				InstallRelease(newRelease)
 
 				// Check that operator and target versions will be set to the newer. Ignore observed version, since it
--- a/test/releases/0.41.0.go
+++ b/test/releases/0.41.0.go
@@ -65,6 +65,7 @@ func init() {
 			Ovs:         &cnao.Ovs{},
 		},
 		Manifests: []string{
+			"network-addons-config.crd.yaml",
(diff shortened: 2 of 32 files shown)


test/releases/0.41.0.go
```diff
@@ -65,6 +65,7 @@ func init() {
 			Ovs:         &cnao.Ovs{},
 		},
 		Manifests: []string{
+			"network-addons-config.crd.yaml",
```
rhrazdil, line 68: What about 0.42.0? Isn't it missing?

RamLavi: totally! good catch! :)

test/releases/releases.go
```diff
@@ -112,6 +112,24 @@ func UninstallRelease(release Release) {
 	}
 }
 
+// Installs given release (RBAC and Deployment)
+func InstallOperator(release Release) {
+	manifestName := "operator.yaml"
+	By(fmt.Sprintf("Installing operator %s", release.Version))
+	out, err := Kubectl("apply", "-f", "_out/cluster-network-addons/"+release.Version+"/"+manifestName)
+	Expect(err).NotTo(HaveOccurred(), out)
+
```
rhrazdil, line 121: Extra newline?

test/releases/releases.go
```diff
@@ -112,6 +112,24 @@ func UninstallRelease(release Release) {
...
+}
+
+// Removes given release from cluster
+func UninstallOperator(release Release) {
+	manifestName := "operator.yaml"
+	By(fmt.Sprintf("Uninstalling operator %s", release.Version))
+	out, err := Kubectl("delete", "--ignore-not-found", "-f", "_out/cluster-network-addons/"+release.Version+"/"+manifestName)
+	Expect(err).NotTo(HaveOccurred(), out)
+
```
rhrazdil, line 130: ditto

test/releases/releases.go
RamLavi: Done

test/releases/releases.go
RamLavi: Done

RamLavi:
@rhrazdil fixes in [here](https://github.com/kubevirt/cluster-network-addons-operator/compare/c2022324b09a8d6a9ce2e0bdf98793e730333d67..bc37b3047493456b3f08ad5c5b85966c96406409), also changed the commit message

rhrazdil:
Thanks, very clear now.
/lgtm

RamLavi:
retest KMP teir1: `KubeVirt not ready in time`
/retest

test/e2e/lifecycle/upgrade_test.go
```diff
@@ -17,6 +17,7 @@ var _ = Context("Cluster Network Addons Operator", func() {
 	testUpgrade := func(oldRelease, newRelease Release) {
 		Context(fmt.Sprintf("when operator in version %s is installed and supported spec configured", oldRelease.Version), func() {
 			BeforeEach(func() {
+				UninstallRelease(newRelease)
```
alonSadan, line 20: question: Why it is needed to uninstall newRelease here? Where it is being installed in the first place?

alonSadan (commented):
Looks good overall. Thanks.
one question added.

test/e2e/lifecycle/upgrade_test.go
RamLavi: the test is initiated after performing make cluster up+sync, i.e. in the moment of starting the test, the new release is already deployed.
if we try to do that we'll get an error, since downgrade process is not supported.

alonSadan: > the test is initiated after performing make cluster up+sync, i.e. in the moment of starting the test, the new release is already deployed.
> if we try to do that we'll get an error, since downgrade process is not supported.

Thank you.
/lgtm

PR #580, opened 2020-09-30
fix AddToScheme to add all served versions

RamLavi:
We are currently using the old shceme version when
adding resources to schema.
Let's change it to use all served versions v1 and v1alpha1

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/apis/addtoscheme_networkaddonsoperator_v1alpha1.go
+++ b/pkg/apis/addtoscheme_networkaddonsoperator_v1alpha1.go
@@ -1,0 +1,10 @@
+package apis
+
+import (
+	cnaov1alpha1 "github.com/kubevirt/cluster-network-addons-operator/pkg/apis/networkaddonsoperator/v1alpha1"
+)
+
+func init() {
+	// Register the types with the Scheme so the components can map objects to GroupVersionKinds and back
+	AddToSchemes = append(AddToSchemes, cnaov1alpha1.SchemeBuilder.AddToScheme)
+}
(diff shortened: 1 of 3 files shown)


pkg/apis/addtoscheme_networkaddonsoperator_v1alpha1.go
```diff
@@ -0,0 +1,10 @@
+package apis
+
+import (
+	cnaov1alpha1 "github.com/kubevirt/cluster-network-addons-operator/pkg/apis/networkaddonsoperator/v1alpha1"
+)
+
+func init() {
+	// Register the types with the Scheme so the components can map objects to GroupVersionKinds and back
+	AddToSchemes = append(AddToSchemes, cnaov1alpha1.SchemeBuilder.AddToScheme)
```
qinqon, line 9: This is raceful in case `init` is not sequencially called.

qinqon (commented):
Question related to golang "init" function

pkg/apis/addtoscheme_networkaddonsoperator_v1alpha1.go
RamLavi: well it's used in the same manner in knmstate and KMP, but I didn't find solid evidence that it runs sequentially, so let's be on the safe side.
Done

RamLavi: it's done the same in knmstate

PR #585, opened 2020-09-30
Fix manifest link versioning in README.md

rhrazdil:
Instead of using absolute path for linking manifests in
README.md, use links to released assets.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -249,16 +249,16 @@ spec:
 First install the operator itself:
 
 ```shell
-kubectl apply -f https://raw.githubusercontent.com/kubevirt/cluster-network-addons-operator/master/manifests/cluster-network-addons/0.42.0/namespace.yaml
-kubectl apply -f https://raw.githubusercontent.com/kubevirt/cluster-network-addons-operator/master/manifests/cluster-network-addons/0.42.0/network-addons-config.crd.yaml
-kubectl apply -f https://raw.githubusercontent.com/kubevirt/cluster-network-addons-operator/master/manifests/cluster-network-addons/0.42.0/operator.yaml
+kubectl apply -f https://github.com/kubevirt/cluster-network-addons-operator/releases/download/v0.42.0/namespace.yaml
+kubectl apply -f https://github.com/kubevirt/cluster-network-addons-operator/releases/download/v0.42.0/network-addons-config.crd.yaml
+kubectl apply -f https://github.com/kubevirt/cluster-network-addons-operator/releases/download/v0.42.0/operator.yaml
 ```
 
 Then you need to create a configuration for the operator [example
 CR](manifests/cluster-network-addons/0.4.0/network-addons-config-example.cr.yaml):
 
 ```shell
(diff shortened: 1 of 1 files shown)


RamLavi:
@rhrazdil we would also need to change the branch `release-0.42` in the link, to be dynamic. if not then the next stable branch this link will be broken.
`https://raw.githubusercontent.com/kubevirt/cluster-network-addons-operator/release-0.42/manifests/cluster-network-addons/0.42.2/namespace.yaml`

maybe we can move to this link format?
https://github.com/kubevirt/cluster-network-addons-operator/releases/download/v0.42.1/namespace.yaml

phoracek:
@RamLavi yeah, linking the release would be more durable. +1

rhrazdil:
@RamLavi @phoracek Moved to links that point to the release asset files.
I'm linking to version v0.42.2 which [is on the way](https://github.com/kubevirt/cluster-network-addons-operator/pull/584). I tested that the sed command updates the version correctly.

hack/prepare-release.sh
```diff
@@ -14,8 +14,8 @@ VERSION=${released_version} IMAGE_TAG=${prefixed_released_version} make gen-mani
 git add manifests/cluster-network-addons/${released_version}
 
 echo 'Upgrade README.md with the released manifests'
-sed -i "s/\(.*kubectl apply.*\)${prefixed_previous_version}\(.*\)/\1${prefixed_released_version}\2/g" README.md
-sed -i "s/\(.*startingCSV.*\)${prefixed_previous_version}\(.*\)/\1${prefixed_released_version}\2/g" README.md
+sed -i "s/\(.*kubectl apply.*\)${previous_version}\(.*\)/\1${released_version}\2/g" README.md
```
RamLavi, line 17: if we're using the release version link (example: https://github.com/kubevirt/cluster-network-addons-operator/releases/download/v0.42.2/namespace.yaml) 
then I think the `v` **is** needed, right?
if so then this specific commit is not needed.

RamLavi (changes requested):

hack/prepare-release.sh
rhrazdil: > then I think the v is needed, right?

You're correct, of course. I'm leaving the prepare-release.sh as it was and just changing the link in README.
Also, after testing `make prepare-minor` properly, I'm using `v0.42.0` version as that's what we have in https://github.com/kubevirt/cluster-network-addons-operator/blob/043348fec507cab45501c5d859b9fd76ddac5ecc/version/version.go#L4

RamLavi:
restarting travis: not related error
`ADD failed: failed to GET https://github.com/containernetworking/plugins/releases/download/v0.8.7/cni-plugins-linux-amd64-v0.8.7.tgz with status 429 too many requests: <html>`

RamLavi:
@rhrazdil please update the PR description

rhrazdil:
@RamLavi Updated, thanks

phoracek:
/lgtm
/approve

Thanks!

PR #587, opened 2020-10-01
crdCleanUp, Cleaning release after upgrade test

RamLavi:
Currently, after cnao is uninstalled, the crd's
of it's components are not deleted.
This causes a "downgrade" situation when
upgrade test move from 1 test to the other.
For example, in the first test, release is upgraded
from old version to current version, but in the next test
the old version is not been able to be set because the new crd's
are still there.

In order to solve this, and since CNAO is not supporting downgrades,
we are attempting to remove the crds manually after each upgrade test.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/test/releases/0.16.0.go
+++ b/test/releases/0.16.0.go
@@ -62,6 +62,13 @@ func init() {
 			"network-addons-config.crd.yaml",
 			"operator.yaml",
 		},
+		CrdCleanUp: []string{
+			"network-attachment-definitions.k8s.cni.cncf.io",
+			"networkaddonsconfigs.networkaddonsoperator.network.kubevirt.io",
+			"nodenetworkconfigurationenactments.nmstate.io",
+			"nodenetworkconfigurationpolicies.nmstate.io",
+			"nodenetworkstates.nmstate.io",
+		},
 	}
 	releases = append(releases, release)
 }
--- a/test/releases/0.17.0.go
+++ b/test/releases/0.17.0.go
@@ -62,6 +62,13 @@ func init() {
 			"network-addons-config.crd.yaml",
 			"operator.yaml",
 		},
+		CrdCleanUp: []string{
+			"network-attachment-definitions.k8s.cni.cncf.io",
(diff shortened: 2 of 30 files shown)


cmd/manager/main.go
```diff
@@ -91,6 +92,10 @@ func main() {
 		log.Printf("failed adding openshift scheme to the client: %v", err)
 		os.Exit(1)
 	}
+	if err := apiextensionsv1.AddToScheme(mgr.GetScheme()) ; err != nil {
```
RamLavi, line 95: This doesn't seem to work: getting error:
```
failed to Clean Up outdated objects: invalid configuration:
no kind is registered for the type apiextensions.CustomResourceDefinition in scheme "k8s.io/client-go/kubernetes/scheme/register.go:69"
```

pkg/network/nmstate.go
```diff
@@ -72,8 +70,63 @@ func cleanUpNMState(conf *cnao.NetworkAddonsConfigSpec, ctx context.Context, cli
 
 	} else if apierrors.IsNotFound(err) {
 		// object not found, no need for action.
-		return nil
+		return []error{}
 	}
+	return []error{}
+}
+
+func removeOldStoredVersion(ctx context.Context, client k8sclient.Client) []error {
```
phoracek, line 78: I haven't went over the code, but this looks scary. We must make sure that CRs created by users are not deleted during the upgrade. Is this code safe?

RamLavi: this change is only removing a part of the crd's status, that declares which stored version is used.
a similar thing is done [when droping support to deprecated versions](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#upgrade-existing-objects-to-a-new-stored-version)

normally there shouldn't be more than 1 storedversion AFAIK

RamLavi: but maybe you're right, checking if this is necessary

RamLavi: you're totally right.

PR #588, opened 2020-10-04
bumper, Fix update of component in components.yaml

RamLavi:
This PR fixed a bug in bumper script update to components.yaml

- **bumper, Fix update of component in components.yaml**
Currently the yaml is not really changed
as we are not updating the real items when iteration, but their copy instance.
Lets fix this by updating the real items.
Also, once the components.yaml is changed the components are reformatted.
So Let's align to the way updateComponentsYaml updates it.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/components.yaml
+++ b/components.yaml
@@ -1,43 +1,43 @@
 components:
-  multus:
-    url: "https://github.com/intel/multus-cni"
-    commit: "bfaf22964b51b93b08dcb1f10cbd8f8cb9a3603a"
-    branch: "master"
-    update-policy: "tagged"
-    metadata: "v3.4.1"
-  linux-bridge:
-    url: "https://github.com/containernetworking/plugins"
-    commit: "9b8de6a613657df14a3a5958642872fa9855d570"
-    branch: "master"
-    update-policy: "tagged"
-    metadata: "v0.8.7"
   bridge-marker:
-    url: "https://github.com/kubevirt/bridge-marker"
-    commit: "26d1084453e9d4aa066cf4390326eefa4f8aa5eb"
-    branch: "master"
-    update-policy: "tagged"
-    metadata: "0.4.0"
+    url: https://github.com/kubevirt/bridge-marker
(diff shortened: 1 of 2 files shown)


RamLavi:
@phoracek do you think this change could break the M/S scripts or any other third part project using this? (I mainly mean to the removal of the quotes)

oshoval (commented):
Please capitalize the word "fix" in the title



"This PR fixed a bug in bumper script write to components.yaml"
i would change it to updating instead of write to

"fix update of component in components.yaml
Currently the yaml is not really changed
since we don't update the correct struct, but its for loop copy."

can you explain better the last 5 words ? "but .."

RamLavi:
> can you explain better the last 5 words ? "but .."

by `loop copy instance`, I mean that when you `i, instance := range listObject {...` then instance is only a copy of listObject[i], so if you update it, it's not really updating listObject[i]

RamLavi:
> Please capitalize the word "fix" in the title
> 
> "This PR fixed a bug in bumper script write to components.yaml"
> i would change it to updating instead of write to

done

oshoval:
> > can you explain better the last 5 words ? "but .."
> 
> by `loop copy instance`, I mean `instance` in this example:
> 
> ```
> i, instance := range listObject {...
> ```
> 
> Notice that when you run this loop, then `instance` is only a copy of `listObject[i]`, so if you update it, it's not really updating `listObject[i]`

Thanks, please explain it by few more words in the commit message please if possible, same as you said here.

oshoval:
if you can align PR title with commit title please (missing first word, capital Fix)

RamLavi:
> if you can align PR title with commit title please (missing first word, capital Fix)

Done

RamLavi:
/hold

let's wait for @phoracek to make sure this doesn't break third party scripts

phoracek:
It should work ok. I used pyyaml on tier2. And after your change, it is still a valid yaml.

/hold cancel

PR #592, opened 2020-10-05
bumper, Add bumper make target

RamLavi:
Currently running bumper script requires 2 commands.
The bumper exec output is added to the main directory
and may be mistakenly added to the project.

Let's change the command to run with 'go run',
and move it to a Makefile target for better usability
README is updated accordingly.
Also we add a link to how to produce a github token in README

Also added a link to token creation in case
of running script manually

**Release note**:

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -182,6 +182,9 @@ vendor: $(GO)
 	$(GO) mod tidy
 	$(GO) mod vendor
 
+auto-bumper:
+	$(GO) run $(shell ls tools/bumper/*.go | grep -v test) ${ARGS}
+
 bump-%:
 	CNAO_VERSION=${VERSION} ./hack/components/bump-$*.sh
 bump-all: bump-knmstate bump-kubemacpool bump-macvtap bump-linux-bridge bump-multus bump-ovs-cni bump-bridge-marker
@@ -212,6 +215,7 @@ bump-all: bump-knmstate bump-kubemacpool bump-macvtap bump-linux-bridge bump-mul
 	prepare-minor \
 	prepare-major \
 	vendor \
+	auto-bumper \
 	bump-% \
 	bump-all \
 	gen-k8s \
--- a/tools/bumper/README.md
+++ b/tools/bumper/README.md
@@ -2,9 +2,12 @@
 
 The Bumper script goes over all of CNAO's components, using the components.yaml config, finds new releases and bumps them in separate PRs. The script can be run locally or via an automation such as GitHub Actions.
 
+## Running the script manually
+
+In order to run the script manually, you need to have a github token. To create a token in your github user, follow this [guide](https://docs.github.com/en/free-pro-team@latest/github/authenticating-to-github/creating-a-personal-access-token).
+
 ## How to run the script
 
 ```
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -182,6 +182,10 @@ vendor: $(GO)
 	$(GO) mod tidy
 	$(GO) mod vendor
 
+auto-bumper:
+	$(GO) install ${GOPATH}/src/github.com/kubevirt/cluster-network-addons-operator/tools/bumper
+	${GOBIN}/bumper ${ARGS}
```
qinqon, line 187: Since we are with go modules we can just do `go run tools/bumper ${ARGS}`

qinqon (changes requested):
Let's just use go run

Makefile
RamLavi: AFAIK, since bumper is a script comprised of more than 1 file, building it is mandatory

qinqon: nope, it will compile and use them, since they are under the same project (now is when  you test it and it does not work :-).

Makefile
```diff
@@ -182,6 +182,9 @@ vendor: $(GO)
 	$(GO) mod tidy
 	$(GO) mod vendor
 
+auto-bumper:
+	$(GO) run $(shell ls tools/bumper/*.go |grep -v test) ${ARGS}
```
oshoval, line 186: consider please 2nd answer here
https://stackoverflow.com/questions/23695448/how-to-run-all-go-files-within-current-directory-through-the-command-line-mult

oshoval, line 186: consider please 2nd answer here
https://stackoverflow.com/questions/23695448/how-to-run-all-go-files-within-current-directory-through-the-command-line-mult

or this one `go run !(*_test).go` from there

RamLavi: Done

qinqon: I think `!(*_test).go` does not play nice with Makefile.

oshoval: if its in a shell there or $$ as you said ?

RamLavi: let's keep it simple. I prefer @qinqon 's suggestion.

oshoval: well if it works, is the same but syntactic sugar
up to you, curious if it can works with a simple chance in the Makefile and then we will know the new method
"|grep" -> "| grep" :)

oshoval (approved):
Sw33t

tools/bumper/README.md
```diff
@@ -2,9 +2,12 @@
...
+## Running the script manually
+
+In order to run the script manually, you need to have a github token. To create a token in your github user, follow this [guide](https://docs.github.com/en/free-pro-team@latest/github/authenticating-to-github/creating-a-personal-access-token).
+
 ## How to run the script
 
 ```
-go build $GOPATH/src/github.com/kubevirt/cluster-network-addons-operator/tools/bumper
-./bumper -config-path=<path-to-components.yaml> -token=<git-token>
+make ARGS="config-path=<path-to-components.yaml> -token=<git-token>" auto-bumper
```
qinqon, line 12: `-config-path`

RamLavi: Done

PR #595, opened 2020-10-07
Fix installation of bridge and tuning CNIs

phoracek:

We have two serious bugs in the bridge CNI.

One, we never installed our own bridge and tuning CNI, since the source
of our copy was not matching directories used in the CNI plugins
container image. The only reason why this worked so far is that we were
able to consume these plugins from the Multus deployment.

Second, the symlink making sure that we always have an alias to these
CNIs prefixed with cnv- was not working on OpenShift. The reason is that
on OpenShift, the directory containing CNI binaries is different, but
our `ln` did not respect that.

**Release note**:

```release-note
Fix installation of bridge CNI on OpenShift
```

### DIFF
--- a/data/linux-bridge/002-linux-bridge.yaml
+++ b/data/linux-bridge/002-linux-bridge.yaml
@@ -32,14 +32,15 @@ spec:
             - /bin/bash
             - -c
             - |
-              cp -rf /usr/src/containernetworking/plugins/bin/*bridge /opt/cni/bin/
-              cp -rf /usr/src/containernetworking/plugins/bin/*tuning /opt/cni/bin/
+              echo 'Installing bridge and tuning CNIs'
+              cp -f /usr/src/containernetworking/plugins/bin/*bridge /opt/cni/bin/
+              cp -f /usr/src/containernetworking/plugins/bin/*tuning /opt/cni/bin/
               # Some projects (e.g. openshift/console) use cnv- prefix to distinguish between
               # binaries shipped by OpenShift and those shipped by KubeVirt (D/S matters).
               # Following two lines make sure we will provide both names when needed.
-              find /opt/cni/bin/cnv-bridge || ln -s /opt/cni/bin/bridge /opt/cni/bin/cnv-bridge
-              find /opt/cni/bin/cnv-tuning || ln -s /opt/cni/bin/tuning /opt/cni/bin/cnv-tuning
-              echo "Entering sleep... (success)"
+              find /opt/cni/bin/cnv-bridge &>/dev/null || ln -s /opt/cni/bin/bridge {{ .CNIBinDir }}/cnv-bridge
+              find /opt/cni/bin/cnv-tuning &>/dev/null || ln -s /opt/cni/bin/tuning {{ .CNIBinDir }}/cnv-tuning
+              echo 'Entering sleep... (success)'
               sleep infinity
           resources:
             requests:
@@ -49,7 +50,7 @@ spec:
             privileged: true
           volumeMounts:
             - name: cnibin
-              mountPath: /opt/cni/bin
+              mountPath: {{ .CNIBinDir }}
       volumes:
         - name: cnibin
           hostPath:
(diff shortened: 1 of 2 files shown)


phoracek:
/hold

Once this has an initial review, I will need to update our build of the CNI container image

data/linux-bridge/002-linux-bridge.yaml
```diff
@@ -32,14 +32,15 @@ spec:
...
+              cp -f /usr/src/containernetworking/plugins/bin/*bridge /opt/cni/bin/
+              cp -f /usr/src/containernetworking/plugins/bin/*tuning /opt/cni/bin/
               # Some projects (e.g. openshift/console) use cnv- prefix to distinguish between
               # binaries shipped by OpenShift and those shipped by KubeVirt (D/S matters).
               # Following two lines make sure we will provide both names when needed.
-              find /opt/cni/bin/cnv-bridge || ln -s /opt/cni/bin/bridge /opt/cni/bin/cnv-bridge
-              find /opt/cni/bin/cnv-tuning || ln -s /opt/cni/bin/tuning /opt/cni/bin/cnv-tuning
-              echo "Entering sleep... (success)"
+              find /opt/cni/bin/cnv-bridge &>/dev/null || ln -s /opt/cni/bin/bridge {{ .CNIBinDir }}/cnv-bridge
+              find /opt/cni/bin/cnv-tuning &>/dev/null || ln -s /opt/cni/bin/tuning {{ .CNIBinDir }}/cnv-tuning
```
qinqon, line 42: Let's just re-create the link if it exists.
```suggestion
              ln -fs /opt/cni/bin/bridge {{ .CNIBinDir }}/cnv-bridge
              ln -fs /opt/cni/bin/tuning {{ .CNIBinDir }}/cnv-tuning
```

But current code is wrong since we are checking the link it has to be 
```suggestion
              find {{ .CNIBinDir }}/cnv-bridge &>/dev/null || ln -s /opt/cni/bin/bridge {{ .CNIBinDir }}/cnv-bridge
              find {{ .CNIBinDir }}/cnv-tuning &>/dev/null || ln -s /opt/cni/bin/tuning {{ .CNIBinDir }}/cnv-tuning
```

qinqon (changes requested):

data/linux-bridge/002-linux-bridge.yaml
phoracek: Merged this too quickly. We cannot recreate it. The issue is that on D/S, the binary name is cnv-bridge, on U/S it is bridge. In order to make the cnv-bridge binary available on D/S, we just dump it on the host. In order to make it available on U/S (for UI testing), we make a symlink to bridge called cnv-bridge

hack/components/bump-linux-bridge.sh
```diff
@@ -22,7 +22,7 @@ echo 'Get linux-bridge tag'
 LINUX_BRIDGE_TAG=$(git-utils::get_component_tag ${LINUX_BRIDGE_PATH})
 
 echo 'Build container image with linux-bridge binaries'
-LINUX_BRIDGE_TAR_CONTAINER_DIR=/usr/src/github.com/containernetworking/plugins/bin
+LINUX_BRIDGE_TAR_CONTAINER_DIR=/usr/src/containernetworking/plugins/bin
```
rhrazdil, line 25: It this fixing the first part of the problem? eg:
```
One, we never installed our own bridge and tuning CNI, since the source
of our copy was not matching directories used in the CNI plugins
container image. The only reason why this worked so far is that we were
able to consume these plugins from the Multus deployment.
```

phoracek: I wonder though whether the result is better then. We would ship our own CNIs, but when Multus gets installed, we will never know which one is being used. I'm worried that it would make debugging harder.

I'm open to suggestions on how to fix this for good

RamLavi:
closing in favor of https://github.com/kubevirt/cluster-network-addons-operator/pull/948

PR #611, opened 2020-10-13
bumper, Fix flaky in bumper unit test

RamLavi:
In unit test checking getUpdatedReleaseInfo
entries sometimes failed. This is because
getUpdatedReleaseInfo assumes that the tags list
received by ListMatchingRefs is sorted by time inserted.
However fake ListMatchingRefs method doesn't explicitly
sort it, so sometimes we get different order that causes
the test to fail.
Let's sort the tag list to fix this.
A comment to keep the tags to be inserted in
alphabetic-chronological order is also added.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tools/bumper/git_fakes.go
+++ b/tools/bumper/git_fakes.go
@@ -3,6 +3,7 @@ package main
 import (
 	"encoding/json"
 	"fmt"
+	"sort"
 	"io/ioutil"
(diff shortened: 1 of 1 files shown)


phoracek (approved):
/lgtm
/approve

PR #616, opened 2020-10-13
bumper, Add bumping repo instance

RamLavi:
This PR adds a bumping repo component instance in the bumper script.

- **bumper, Add a cnao repo instance and cnaoRepoCommands.go**
Currently, functions that run on the cnao repo
are run with the component api, instead of on it's own instance.
To make things neater, Let's create a cnaoRepo instance,
and use it to run cnao repo related actions.
Additionaly, all cnao repo related function move to cnaoRepoCommands.go

- **bumper, Move cnap repo related tests to cnaoRepoCommands_test.go**
We purely* move the unit-tests related to cnaoRepo to the dedicated
test file cnaoRepoCommands_test.go
* only exception is a small fix in isComponentBumpNeeded where we also
update the url so that the test won't fail.

- **bumper, Move config parsing actions to cnaoRepo instance**
Let's move all actions regarding components.yaml (aka config file)
to run on the cnaoRepo instance.
To do so, we change the config-path input param to be
relative to the cnaoRepo, so that it'll be easy to concat it
to the cnaoRepo instance.
We also add another reading of the config file to
refresh it's content after each bump.

- **bumper, Move running bump-<component> scripts to cnaoRepo instance**
Let's move bump component script actions to cnaoRepo instance

**Release note**:

```release-note
NONE
```


alonSadan (commented):
Thanks Ram.
/lgtm

RamLavi:
/hold cancel

since working with clone may cause debugging issues in the future, I change the repo instance to run from the current cnao repo.

tools/bumper/bumper.go
```diff
@@ -80,18 +87,23 @@ func main() {
...
 			componentsConfig.Components[componentName] = component
-			err = updateComponentsYaml(inputArgs.componentsConfigPath, componentsConfig)
+			err = updateComponentsYaml(configPath, componentsConfig)
 			if err != nil {
 				exitWithError(errors.Wrap(err, "Failed to update components yaml"))
 			}
 
 			logger.Printf("Running bump-%s script", componentName)
-			cmd := exec.Command("make", fmt.Sprintf("bump-%s", componentName))
+			cmd := exec.Command("make", "-C", bumpingRepo.gitRepo.localDir, fmt.Sprintf("bump-%s", componentName))
```
qinqon, line 106: This should be encapsulated at cnaoRepo `cnaoRepo.Bump(componentName)`

tools/bumper/bumper.go
```diff
@@ -118,7 +130,7 @@ func initLog() *log.Logger {
 }
 
 func initFlags(paramArgs *inputParams) {
-	flag.StringVar(&paramArgs.componentsConfigPath, "config-path", "", "Full path to components yaml")
+	flag.StringVar(&paramArgs.componentsConfigPath, "config-path", "", "relative path to components yaml from bumping repo")
```
qinqon, line 133: ditto, let's remove this argument.

tools/bumper/component_commands.go
```diff
@@ -67,10 +67,28 @@ type gitRepo struct {
 }
 
 const (
-	cnaoOwner = "kubevirt"
-	cnaoRepo  = "cluster-network-addons-operator"
+	bumpingRepoUrl = "https://github.com/kubevirt/cluster-network-addons-operator"
```
qinqon, line 70: we are not going to create new clone we don't need this

qinqon (changes requested):
First pass: Don't use `Component` for CNAO just work with the `gitRepo` from it also we can totally remove the config-path param and stick with components.yaml this is not a generic tool.

tools/bumper/bumper.go
```diff
@@ -37,8 +38,14 @@ func main() {
 		exitWithError(errors.Wrap(err, "Failed to create github api instance"))
 	}
 
+	bumpingRepo, err := getBumpingRepoComponent(githubApi)
```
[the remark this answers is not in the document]
RamLavi: I actually think making it generic is the preferred practice. 
Not because maybe someone will want to use it in other repo (although who knows)
but because this way I keep the repo specific stuff in specific const variables so it'll be easier to modify stuff later on.

Considering the minimal effort, I think keeping it generic is the preferred way in this case.

tools/bumper/bumper.go
```diff
@@ -68,7 +75,7 @@ func main() {
 		}
 
 		proposedPrTitle := fmt.Sprintf("bump %s to %s", componentName, updatedReleaseTag)
-		bumpNeeded, err := gitComponent.isBumpNeeded(currentReleaseTag, updatedReleaseTag, component.Updatepolicy, proposedPrTitle)
+		bumpNeeded, err := bumpingRepo.isBumpNeeded(currentReleaseTag, updatedReleaseTag, component.Updatepolicy, proposedPrTitle)
```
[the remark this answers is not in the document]
RamLavi: The main reason why `isBumpNeeded` is a gitComponent function is that we want to access `isPrAlreadyOpened` inside it...

`isPrAlreadyOpened` specifically checks if there is no open bump PR in **CNAO** repo.. so keeping it under bumpingRepo is the right thing in my opinion..

tools/bumper/component_commands.go
```diff
@@ -67,10 +67,28 @@ type gitRepo struct {
...
 )
 
+func getBumpingRepoComponent(api *githubApi) (*gitComponent, error) {
```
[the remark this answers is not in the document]
RamLavi: `gitComponent` is a struct that allows us go-git functions (like `status()`, `Add()` etc.. ) and githubAPI functions (like future functions `createGithubBranch` `creategithubPR`, etc...)
all those are fit the bumping repo needs, so I don't agree... keeping it in `gitComponent` is the best fit.

btw, we technically can call all functions (like `getUpdatedReleaseInfo`) on the bumping component, it'll work, we just don't want to do it in the purpose of this script.

tools/bumper/bumper.go
```diff
@@ -80,18 +87,23 @@ func main() {
 
 			// Create PR
 			exitWithError(fmt.Errorf("create PR not implemented yet"))
+			// update components yaml in the bumping repo instance
+			componentsConfig, err = parseComponentsYaml(configPath)
+			if err != nil {
+				exitWithError(errors.Wrap(err, "Failed to parse components yaml during bump"))
+			}
```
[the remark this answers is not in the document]
RamLavi: componentsConfig is not refreshed when resetting the cnao repo, it's on RAM.
we need to refresh the componentsConfig, otherwise `components.yaml` will accumulate past bump changes.
For example, if first bump was knmstate to 0.32.0, then `componentsConfig` will have this change..
and then next component we need to bump multus to 3.6, the `componentsConfig` will hold both changes..

tools/bumper/bumper.go
RamLavi: ditto [comment](https://github.com/kubevirt/cluster-network-addons-operator/pull/616#discussion_r504485958)

tools/bumper/component_commands.go
RamLavi: now that cnao is kept in component struct, we can use `getComponentNameFromUrl`, `getComponentOwnerFromUrl` to get `cnaoOwner`, `cnaoRepo` from there. keeping this in url makes it more robust.

tools/bumper/bumper.go
```diff
@@ -37,8 +38,14 @@ func main() {
...
 	}
 
+	bumpingRepo, err := getBumpingRepoComponent(githubApi)
+	if err != nil {
+		exitWithError(errors.Wrap(err, "Failed to clone bumping repo"))
+	}
+
 	logger.Printf("Parsing %s", inputArgs.componentsConfigPath)
-	componentsConfig, err := parseComponentsYaml(inputArgs.componentsConfigPath)
+	configPath := filepath.Join(bumpingRepo.gitRepo.localDir, inputArgs.componentsConfigPath)
```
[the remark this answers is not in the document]

tools/bumper/bumper.go

tools/bumper/bumper.go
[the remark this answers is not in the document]
qinqon: You can still use constants but don't use generic names for CNAO like BumpingRepo or the like, is really missleading and you know to understand that BumpingRepo is CNAO itself, also CNAO itself is not a network component from CNAO.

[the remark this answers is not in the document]
RamLavi: changed all `BumpingRepo` to `cnaoRepo`
 Done

tools/bumper/cnaoRepoCommands.go
```diff
@@ -0,0 +1,168 @@
...
+func (cnaoRepoOps *gitCnaoRepo) isComponentBumpNeeded(currentReleaseVersion, latestReleaseVersion, updatePolicy, proposedPrTitle string) (bool, error) {
+	logger.Printf("currentReleaseVersion: %s, latestReleaseVersion: %s, updatePolicy: %s\n", currentReleaseVersion, latestReleaseVersion, updatePolicy)
+
+	if updatePolicy == updatePolicyStatic {
+		logger.Printf("updatePolicy is static. avoiding auto bump")
+		return false, nil
+	}
+
+	// check if PR not already opened
+	isAlreadyOpened, err := cnaoRepoOps.isPrAlreadyOpened(cnaoRepoOps.getCnaoRepoOwnerFromUrl(), cnaoRepoOps.getCnaoRepoNameFromUrl(), proposedPrTitle)
```
qinqon, line 91: We can reduce this to `cnaoRepoOps.isPrAlreadyOpened(proposedPrTitle)`

tools/bumper/cnaoRepoCommands.go
```diff
@@ -0,0 +1,168 @@
...
+	latestVersion, err := canonicalizeVersion(latestReleaseVersion)
+	if err != nil {
+		return false, errors.Wrapf(err, "Failed to digest latest Version %s to semver", latestVersion)
+	}
+
+	return currentVersion.LessThan(*latestVersion), nil
+}
+
+// check if there is an already open PR with the same title in the repo.
+func (cnaoRepoOps *gitCnaoRepo) isPrAlreadyOpened(owner, repo, proposedPrTitle string) (bool, error) {
```
qinqon, line 118: owner and repo are going to be the ones at CNAO no need to pass them

tools/bumper/cnaoRepoCommands.go
```diff
@@ -0,0 +1,168 @@
...
+
+func (cnaoRepoOps *gitCnaoRepo) getCnaoRepoNameFromUrl() string {
+	urlSlice := strings.Split(cnaoRepoOps.configParams.Url, "/")
+	return urlSlice[len(urlSlice)-1]
+}
+
+func (cnaoRepoOps *gitCnaoRepo) getCnaoRepoOwnerFromUrl() string {
+	urlSlice := strings.Split(cnaoRepoOps.configParams.Url, "/")
+	return urlSlice[len(urlSlice)-2]
+}
```
qinqon, line 143: They are only use for `isPrAlradyOpenned` so we can just do the split once there and access the position.

tools/bumper/cnaoRepoCommands.go
```diff
@@ -0,0 +1,168 @@
+package main
```
qinqon, line 1: In golang go sources fails are usually lowercase, let's move this to just `cnaorepocommands.go`

tools/bumper/cnaoRepoCommands.go
```diff
@@ -0,0 +1,168 @@
...
+	}
+
+	return currentVersion.LessThan(*latestVersion), nil
+}
+
+// check if there is an already open PR with the same title in the repo.
+func (cnaoRepoOps *gitCnaoRepo) isPrAlreadyOpened(owner, repo, proposedPrTitle string) (bool, error) {
+	logger.Printf("checking if there is an already open bump PR for this release")
+
+	prList, _, err := cnaoRepoOps.githubInterface.ListPullRequests(owner, repo)
```
qinqon, line 121: we can encapsulate things a little more and implement this as `cnaoRepoOps.ListPullRequests()` so we don't need to pass owner and repo to `isPrAlreadyOpenned`

tools/bumper/cnaoRepoCommands_test.go
```diff
@@ -0,0 +1,277 @@
+package main
```
qinqon, line 1: move this file to cnaorepocommaneds_test.go

qinqon (changes requested):
It's starting to look super nice, the unit tests are awesome, let's encapsulate things a little more.

tools/bumper/cnaoRepoCommands.go
RamLavi: in future PRs I'm going to add more functions that use this so it's better to keep it.

tools/bumper/cnaoRepoCommands.go
RamLavi: Done

tools/bumper/cnaoRepoCommands.go
RamLavi: Done

tools/bumper/cnaoRepoCommands.go
RamLavi: `ListPullRequests` is a method of githubInterface that implements githubApi & appropriate fakes,  so I shoudn't move it to cnaoRepoOps.

tools/bumper/cnaoRepoCommands.go
RamLavi: I'll keep it like the other files in the bumper folder: cnao_repo_commands.go
Done

tools/bumper/cnaoRepoCommands_test.go
RamLavi: changed to cnao_repo_commands_test.go
Done

PR #617, opened 2020-10-13
bumper, Add resetRepoInAllowedList to remove bump outputs

RamLavi:
This method will be performed on the cnaoRepo instance
between each bumping of a component, to remove any bumping residue
Since bump output files are located in specific locations,
we add an allowList, to make sure only the bumping outputs
get removed.
Also adding a final resetRepoInAllowedList after all components
got bumped to finish the bumper script with no residual files.
In order to use the resetRepo function with defer, we move the
bump operation to a function.

PR depends on #616 to merge

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tools/bumper/bumper.go
+++ b/tools/bumper/bumper.go
@@ -79,42 +79,55 @@ func main() {
 
 		if componentBumpNeeded {
 			logger.Printf("Bumping %s from %s to %s", componentName, currentReleaseTag, updatedReleaseTag)
-			// reset --hard git repo
-			exitWithError(fmt.Errorf("reset --hader repo not implemented yet"))
 
-			// Create PR
-			exitWithError(fmt.Errorf("create PR not implemented yet"))
-
-			// update components yaml in the bumping repo instance
-			componentsConfig, err := cnaoRepo.getComponentsConfig(inputArgs.componentsConfigPath)
-			if err != nil {
-				exitWithError(errors.Wrap(err, "Failed to get components config during bump"))
-			}
-
-			// update component's entry in config yaml
-			component.Commit = updatedReleaseCommit
-			component.Metadata = updatedReleaseTag
-			componentsConfig.Components[componentName] = component
-			err = cnaoRepo.updateComponentsConfig(inputArgs.componentsConfigPath, componentsConfig)
-			if err != nil {
-				exitWithError(errors.Wrap(err, "Failed to update components yaml"))
-			}
-
-			err = cnaoRepo.bumpComponent(componentName)
+			err = handleBump(cnaoRepo, component, componentName, inputArgs.componentsConfigPath, updatedReleaseTag, updatedReleaseCommit, proposedPrTitle)
 			if err != nil {
 				exitWithError(errors.Wrapf(err, "Failed to bump component %s", componentName))
 			}
-
-			// create a new branch name
-			branchName := strings.Replace(strings.ToLower(proposedPrTitle), " ", "_", -1)
-			logger.Printf("Opening new Branch %s", branchName)
-
-			// push branch to PR
-			exitWithError(fmt.Errorf("push branch to PR not implemented yet"))
 		} else {
 			logger.Printf("Bump not needed in component %s", componentName)
 		}
 	}
+}
+
+func handleBump(cnaoRepo *gitCnaoRepo, component component, componentName, componentsConfigPath, updatedReleaseTag, updatedReleaseCommit, proposedPrTitle string) error {
+	defer func() {
+		err := cnaoRepo.reset()
+		if err != nil {
+			exitWithError(errors.Wrapf(err, "Failed to bump component %s: reset repo failed", componentName))
+		}
+	}()
+
+	// Create PR
+	exitWithError(fmt.Errorf("create PR not implemented yet"))
+
+	// update components yaml in the bumping repo instance
+	componentsConfig, err := cnaoRepo.getComponentsConfig(componentsConfigPath)
+	if err != nil {
+		return errors.Wrap(err, "Failed to get components config during bump")
+	}
+
+	// update component's entry in config yaml
+	component.Commit = updatedReleaseCommit
+	component.Metadata = updatedReleaseTag
+	componentsConfig.Components[componentName] = component
+	err = cnaoRepo.updateComponentsConfig(componentsConfigPath, componentsConfig)
+	if err != nil {
+		return errors.Wrap(err, "Failed to update components yaml")
+	}
+
+	err = cnaoRepo.bumpComponent(componentName)
+	if err != nil {
+		return errors.Wrap(err, "Failed to bump component")
+	}
+
+	// create a new branch name
+	branchName := strings.Replace(strings.ToLower(proposedPrTitle), " ", "_", -1)
+	logger.Printf("Opening new Branch %s", branchName)
+
+	// push branch to PR
+	exitWithError(fmt.Errorf("push branch to PR not implemented yet"))
+	return nil
 }
 
 func initLog() *log.Logger {
--- a/tools/bumper/cnao_repo_commands.go
+++ b/tools/bumper/cnao_repo_commands.go
@@ -22,7 +22,8 @@ type gitCnaoRepo struct {
 }
 
 const (
-	repoUrl = "https://github.com/kubevirt/cluster-network-addons-operator"
+	repoUrl         = "https://github.com/kubevirt/cluster-network-addons-operator"
+	allowListString = "components.yaml,data/*,test/releases/99.0.0.go,pkg/components/components.go"
 )
 
 func getCnaoRepo(api *githubApi) (*gitCnaoRepo, error) {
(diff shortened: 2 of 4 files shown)


tools/bumper/bumper.go
```diff
@@ -83,11 +83,14 @@ func main() {
 
 		if bumpNeeded {
 			logger.Printf("Bumping %s from %s to %s", componentName, currentReleaseTag, updatedReleaseTag)
-			// reset --hard git repo
-			exitWithError(fmt.Errorf("reset --hader repo not implemented yet"))
 
 			// Create PR
 			exitWithError(fmt.Errorf("create PR not implemented yet"))
+			err = bumpingRepo.resetRepo()
```
rhrazdil, line 89: Why is this not before the `// Create PR` part?

tools/bumper/component_commands_test.go
```diff
@@ -460,4 +460,43 @@ var _ = Describe("Testing internal git", func() {
...
+		func(r resetRepoParams) {
+			defer os.RemoveAll(gitComponent.gitRepo.localDir)
+			worktree, err := gitComponent.gitRepo.repo.Worktree()
+
+			By("Modifying files in the Repo")
+			modifyFiles(gitComponent.gitRepo.localDir, r.files)
+			status, err := worktree.Status()
+			Expect(err).ToNot(HaveOccurred(), "should succeed to get repo status")
+			isFilesAdded := len(r.files) != 0
+			Expect(status.IsClean()).To(Equal(!isFilesAdded), "repo should have expected modified files")
```
rhrazdil, line 478: This expect is hard to understand, if we're checking that there are newly modified files, it would be more readable to expect
`!status.IsClean()` and `isFilesAdded`

RamLavi: do you mean this?
```
Expect(!status.IsClean()).To(Equal(isFilesAdded), "repo should have expected modified files")
```

tools/bumper/bumper.go
RamLavi: it shouldn't be.
Done

tools/bumper/component_commands_test.go
rhrazdil: Yes

RamLavi: Done

tools/bumper/bumper.go
```diff
@@ -79,8 +79,11 @@ func main() {
 
 		if componentBumpNeeded {
 			logger.Printf("Bumping %s from %s to %s", componentName, currentReleaseTag, updatedReleaseTag)
-			// reset --hard git repo
-			exitWithError(fmt.Errorf("reset --hader repo not implemented yet"))
+
+			err = cnaoRepo.resetRepoInAllowedList(getAllowedList())
+			if err != nil {
+				exitWithError(errors.Wrapf(err, "Failed to reset bumping repo before bumping component %s", componentName))
+			}
```
qinqon, line 86: What do you think about moving the component bumping to a function so we can put the resetRepoInAllowedList into a `defer` so we don't need the last call to it ?

tools/bumper/bumper.go
```diff
@@ -115,6 +118,12 @@ func main() {
 			logger.Printf("Bump not needed in component %s", componentName)
 		}
 	}
+
+	logger.Printf("Finished bumping components. resetting repo")
+	err = cnaoRepo.resetRepoInAllowedList(getAllowedList())
+	if err != nil {
+		exitWithError(errors.Wrap(err, "Failed to reset bumping repo"))
+	}
```
qinqon, line 126: If we implement this with defer we don't need this call.

tools/bumper/cnaoRepoCommands.go
```diff
@@ -22,7 +22,8 @@ type gitCnaoRepo struct {
 }
 
 const (
-	repoUrl = "https://github.com/kubevirt/cluster-network-addons-operator"
+	repoUrl         = "https://github.com/kubevirt/cluster-network-addons-operator"
+	allowListString = "components.yaml,data/*,test/releases/99.0.0.go,pkg/components/components.go"
```
qinqon, line 26: If we convert this into a string slice we don't need `getAllowedList`

tools/bumper/cnaoRepoCommands.go
```diff
@@ -67,6 +68,30 @@ func (cnaoRepoOps *gitCnaoRepo) updateComponentsConfig(relativeConfigPath string
 	return updateComponentsYaml(configPath, componentsConfig)
 }
 
+// resetRepoInAllowedList resets all the files in AllowedList.
+func (cnaoRepoOps *gitCnaoRepo) resetRepoInAllowedList(allowList []string) error {
```
qinqon, line 72: I now this functions is good for unit test since you pass the allowList, but can we create also the following to encapsulate things a little more ? 

```golang
func (cnaoRepoOps *gitCnaoRepo) resetRepo()  error {
    return resetRepoInAllowedList(allowList)
}
```

tools/bumper/cnaoRepoCommands.go
```diff
@@ -166,3 +191,20 @@ func isVtagFormat(tagVersion string) bool {
...
 }
+
+func runExternalGitCommand(args []string) error {
+	// TODO replace this when go-git adds advanced checkout/restore/clean abilities so we could checkout specific paths
+	cmd := exec.Command("git", args...)
+	cmd.Stderr = os.Stderr
+	cmd.Stdout = os.Stdout
+	err := cmd.Run()
+	if err != nil {
+		return errors.Wrapf(err, "Failed to run git command: git %s", strings.Join(args, " "))
```
qinqon, line 202: Can we add Stderr and Stdout to the returned error ?

tools/bumper/cnaoRepoCommands.go
```diff
@@ -166,3 +191,20 @@ func isVtagFormat(tagVersion string) bool {
...
+	}
+	return nil
+}
+
+// AllowedList is a string array of file globs, used to fine-pick the changes we want to bump.
+func getAllowedList() []string {
+	return strings.Split(allowListString, ",")
+}
```
qinqon, line 210: Not needed if we convert allowList string into a []string

tools/bumper/git_fakes.go
```diff
@@ -344,3 +345,31 @@ func createCommitWithLightweightTag(w *git.Worktree, repo *git.Repository, tagCo
 func getFakePrWithTitle(prTitle string) *github.NewPullRequest {
 	return &github.NewPullRequest{Title: &prTitle}
 }
+
+func fileExists(filename string) bool {
+	info, err := os.Stat(filename)
+	if os.IsNotExist(err) {
+		return false
+	}
```
qinqon, line 353: In case of error different from `IsNotExist` it will return true.

qinqon (changes requested):
Looks petty good, maybe we can use defer to reset stuff.

tools/bumper/cnaoRepoCommands.go
RamLavi: the idea was to keep the repo-specific stuff (like allowListString content and the repo url) in one place. I didn't want to hide it in a functin, and since string slice cannot be a const, I preferred using a string and converting in `getAllowedList` to string slice.

tools/bumper/cnaoRepoCommands.go
RamLavi: Like! =)

tools/bumper/cnaoRepoCommands.go
RamLavi: Done

tools/bumper/git_fakes.go
RamLavi: Done

tools/bumper/bumper.go
RamLavi: is `handleBump` sound good?

RamLavi: `handleBump` sound good?

tools/bumper/bumper.go
RamLavi: Done

tools/bumper/bumper.go
```diff
@@ -79,44 +79,54 @@ func main() {
...
 		} else {
 			logger.Printf("Bump not needed in component %s", componentName)
 		}
 	}
 }
 
+func handleBump(cnaoRepo *gitCnaoRepo, component component, componentName, componentsConfigPath, updatedReleaseTag, updatedReleaseCommit, proposedPrTitle string) (err error) {
+	defer func() {
+		err = cnaoRepo.resetRepo()
+	}()
```
qinqon, line 96: either do `defer cnaoRepo.resetRepo()` or 

```golang
defer cun() {
    err = cnaoRepo.resetRepo()
    exitWithError(errors.Wrap(err, ....
}
```

tools/bumper/cnao_repo_commands.go
```diff
@@ -67,6 +68,35 @@ func (cnaoRepoOps *gitCnaoRepo) updateComponentsConfig(relativeConfigPath string
 	return updateComponentsYaml(configPath, componentsConfig)
 }
 
+// resetRepo is a wrapper for resetRepoInAllowedList
+func (cnaoRepoOps *gitCnaoRepo) resetRepo() error {
```
qinqon, line 72: Let's rename to `reset` since Repo is part of the struct so we have `cnaoRepo.reset()`

qinqon (changes requested):
Pretty good small nits.

tools/bumper/bumper.go
RamLavi: not sure why?
this way if `cnaoRepo.resetRepo()` it'll return an err so it'll be caught in `handleBump`'s err catch:
```
			err = handleBump(cnaoRepo, component, componentName, inputArgs.componentsConfigPath, updatedReleaseTag, updatedReleaseCommit, proposedPrTitle)
			if err != nil {
				exitWithError(errors.Wrapf(err, "Failed to bump component %s", componentName))
			}
```

tools/bumper/cnao_repo_commands.go
RamLavi: Done

tools/bumper/bumper.go
qinqon: I mean with the if statement

PR #619, opened 2020-10-15
bumper, Add collectModifiedToTreeList to collect bump files

RamLavi:
Adding collectModifiedToTreeList that will collect the files
generated by the bump-<component> modified+untracked files
and add them to a githubTree List for future digest in githubApi.

depends on PR #617

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tools/bumper/cnao_repo_commands.go
+++ b/tools/bumper/cnao_repo_commands.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"fmt"
+	"io/ioutil"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -10,6 +11,8 @@ import (
 
 	"github.com/coreos/go-semver/semver"
 	"github.com/go-git/go-git/v5"
+	"github.com/gobwas/glob"
+	"github.com/google/go-github/v32/github"
 	"github.com/pkg/errors"
 )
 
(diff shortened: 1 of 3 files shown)


tools/bumper/cnao_repo_commands.go
```diff
@@ -162,6 +166,53 @@ func (cnaoRepoOps *gitCnaoRepo) isPrAlreadyOpened(proposedPrTitle string) (bool,
...
+		content, err := ioutil.ReadFile(fileNameWithPath)
+		if err != nil {
+			return nil, errors.Wrapf(err, "Failed to read local file %s", fileNameWithPath)
+		}
+
+		if fileInGlobList(localFile, allowedList) {
+			logger.Printf("File added to tree: %s", localFile)
+			entries = append(entries, &github.TreeEntry{Path: github.String(localFile), Type: github.String("blob"), Content: github.String(string(content)), Mode: github.String("100644")})
+		} else {
+			logger.Printf("Skipping file %s, not in allowed list", localFile)
```
alonSadan, line 205: question: why no error is thrown in this case?

alonSadan (commented):
Looks very good! I added one question

tools/bumper/cnao_repo_commands.go
RamLavi: I understand why you'd want to add an error in this case.. but I don't think it preferable, let me explain.

one of the reasons I added the allowList is that we may be running the bumper locally while debugging (/some other reason we would prefer  to run it like that) , and naturally in this scenario you wouldn't want to treat the WIP bumper files as part of the "bump script" changes. this is where allowList is handy, and in this scenario, you wouldn't want to return error, but only skip it..

In any case the PRs generated need to be reviewed and approved by the moderator

alonSadan: oh I understand. Make sense. Thank you for explaining.

tools/bumper/cnao_repo_commands_test.go
```diff
@@ -339,4 +340,126 @@ var _ = Describe("Testing internal git CNAO Repo", func() {
...
+		func(r collectModifiedToTreeListParams) {
+			defer os.RemoveAll(gitCnaoRepo.gitRepo.localDir)
+
+			By("Modifying files in the Repo")
+			modifyFiles(gitCnaoRepo.gitRepo.localDir, r.files)
+
+			By("Running collectModifiedToTreeList api")
+			treeList, err := gitCnaoRepo.collectModifiedToTreeList(r.allowedList)
+
+			By("Checking that tree is as expected")
```
rhrazdil, line 400: nit: I think the `By("Checking that tree is as expected")` should be before `Expect(treeList).To(ConsistOf(r.expectedTreeEntryList), "tree list should be equal to expected")`

Checking the error may deserve separate `By` step

RamLavi: you're right, `Expect(err).To(HaveOccurred(), "should fail collecting Modified files to tree list")` should be with `By("Running collectModifiedToTreeList api")`
Done

rhrazdil:
Thanks
/lgtm

tools/bumper/cnao_repo_commands_test.go
```diff
@@ -339,4 +340,125 @@ var _ = Describe("Testing internal git CNAO Repo", func() {
...
+			Expect(treeList).To(ConsistOf(r.expectedTreeEntryList), "tree list should be equal to expected")
+
+		},
+		Entry("Should return error when no file is modified, allowedList allows all files", collectModifiedToTreeListParams{
+			files:                 []string{},
+			allowedList:           []string{"*"},
+			expectedTreeEntryList: nil,
+			shouldReturnErr:       true,
+		}),
```
qinqon, line 414: Are you sure we have to return error at this scenario ?

qinqon (commented):
Just a question.

tools/bumper/cnao_repo_commands_test.go
RamLavi: I think so, otherwise the bump pr will be empty. We should at least have a change in components.yaml,

qinqon: What I mean is that instead of return an error here we can return an empty `TreeEntry` and  the caller check that if this is empty do nothing.

RamLavi: I agree it's more fluent that way. Also, it's shouldn't stop other bumps so I change that as well.
Done

PR #624, opened 2020-10-20
Create component-bumper.yml

RamLavi:
create a gitActions job that will run the bumper script.
the job will run periodically daily, and produce bump PRs when needed.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/.github/workflows/component-bumper.yml
+++ b/.github/workflows/component-bumper.yml
@@ -1,0 +1,18 @@
+name: Auto-Bump Components' Versions
+
+on:
+  schedule:
+    - cron:  '0 5 * * *'
+
+jobs:
+
+  build:
+    name: CNAO Component Bump Job
+    runs-on: ubuntu-latest
+    steps:
+
+    - name: Check out code into the Go module directory
+      uses: actions/checkout@v2
+
+    - name: Run bumper script
+      run: make ARGS="-config-path=components.yaml -token=${{ secrets.GITHUB_TOKEN }}" auto-bumper


.github/workflows/component-bumper.yml
```diff
@@ -0,0 +1,32 @@
...
+  build:
+    name: CNAO Component Bump Job
+    runs-on: ubuntu-latest
+    steps:
+
+    - name: Set up Go 1.x
+      uses: actions/setup-go@v2
+      with:
+        go-version: ^1.13
+      id: go
```
qinqon, line 18: We don't need to install golang since it get installed as part of `make auto-bumper` that will call install-go.sh

.github/workflows/component-bumper.yml
```diff
@@ -0,0 +1,32 @@
...
+    - name: Check out code into the Go module directory
+      uses: actions/checkout@v2
+
+    - name: Get dependencies
+      run: |
+        go get -v -t -d ./...
+        if [ -f Gopkg.toml ]; then
+            curl https://raw.githubusercontent.com/golang/dep/master/install.sh | sh
+            dep ensure
+        fi
```
qinqon, line 29: We have everything under ./vendor no need for this.

qinqon (changes requested):
We just need to call `make auto-bumper` the golang stuff is already in the project itself.

.github/workflows/component-bumper.yml
RamLavi: Done

.github/workflows/component-bumper.yml
RamLavi: right!

phoracek (approved):
/lgtm
/approve

:1st_place_medal:

PR #631, opened 2020-10-26
tier1, Add multus tier1 automation test

RamLavi:
This PR adds a tier1 automation of multus e2e test
The test runs multus's [test-simple-macvlan1.sh](https://github.com/intel/multus-cni/blob/master/e2e/test-simple-macvlan1.sh) script

**Release note**:

```release-note
NONE
```

### DIFF
--- a/automation/check-patch.e2e-multus-functests.sh
+++ b/automation/check-patch.e2e-multus-functests.sh
@@ -1,0 +1,98 @@
+#!/usr/bin/env bash
+
+set -xeuE
+
+# This script should be able to execute multus
+# functional tests against Kubernetes cluster with
+# CNAO built with latest changes, on any
+# environment with basic dependencies listed in
+# check-patch.packages installed and docker running.
+#
+# Usage:
+# yum -y install automation/check-patch.packages
+# automation/check-patch.e2e-multus-functests.sh
+
+teardown() {
+    rm -rf "${TMP_COMPONENT_PATH}"
+    make cluster-down
+}
+
+function __get-tools() {
+    mkdir -p ${TMP_COMPONENT_PATH}/e2e/bin
+
+    JQ_PATH=${TMP_COMPONENT_PATH}/e2e/bin/jq
+    if ! [ -f ${JQ_PATH} ]; then
+        echo "install jq"
+        curl -Lo ${JQ_PATH} https://github.com/stedolan/jq/releases/download/jq-1.6/jq-linux64
+        chmod +x ${JQ_PATH}
+    fi
+}
+
+function __prepare-test-environment() {
+    __get-tools
+
+    export CLUSTER_PATH=${TMP_PROJECT_PATH}/_kubevirtci/
+    export KUBECTL=${TMP_PROJECT_PATH}/cluster/kubectl.sh
+
+    echo "Deploy cni-plugins pod"
+    CNI_PLUGIN_YAML="cni-install.yml"
+    $KUBECTL create -f ${CNI_PLUGIN_YAML}
+    $KUBECTL -n kube-system wait --for=condition=ready -l name=cni-plugins pod --timeout=300s
+
+    TEST_YAML="*macvlan1.yml"
+    # matching the test nodes to the CNAO cluster nodes
+    sed -i 's/kind-worker$/node01/g' ${TEST_YAML}
+    sed -i 's/kind-worker2$/node02/g' ${TEST_YAML}
+
+    echo "Set eth1 NIC to up using NodeNetworkConfigurationPolicy"
+    cat <<EOF | $KUBECTL apply -f -
+apiVersion: nmstate.io/v1beta1
+kind: NodeNetworkConfigurationPolicy
+metadata:
+  name: eth1-policy
+spec:
+  desiredState:
+    interfaces:
+    - name: eth1
+      description: Configuring eth1 on all nodes
+      type: ethernet
+      state: up
+      ipv4:
+        dhcp: true
+        enabled: true
+EOF
+}
+
+function __remove-test-manifests() {
+    echo "Remove test manifests"
+    $KUBECTL delete -f ${TEST_YAML}
+    $KUBECTL delete -f ${CNI_PLUGIN_YAML}
+}
+
+main() {
+    # Setup CNAO and artifacts temp directory
+    source automation/check-patch.setup.sh
+    cd ${TMP_PROJECT_PATH}
+
+    # Spin-up ephemeral cluster with latest CNAO, 2 nodes and secondary NIC needed for test
+    export KUBEVIRT_NUM_NODES=2
+    export KUBEVIRT_NUM_SECONDARY_NICS=1
+    # this script also exports KUBECONFIG, and fetch $COMPONENT repository
+    COMPONENT="multus" source automation/components-functests.setup.sh
+
+    trap teardown EXIT SIGINT
+
+    # Run multus functional tests
+    (
+        cd ${TMP_COMPONENT_PATH}/e2e
+
+        __prepare-test-environment
+
+        echo "Run multus macvlan test"
+        ./test-simple-macvlan1.sh
+
+        __remove-test-manifests
+    )
+}
+
+[[ "${BASH_SOURCE[0]}" == "$0" ]] && main "$@"


automation/check-patch.e2e-multus-functests.sh
```diff
@@ -0,0 +1,91 @@
...
+# CNAO built with latest changes, on any
+# environment with basic dependencies listed in
+# check-patch.packages installed and docker running.
+#
+# yum -y install automation/check-patch.packages
+# automation/check-patch.e2e-multus-functests.sh
+
+teardown() {
+    make cluster-down
+    rm -rf "${TMP_COMPONENT_PATH}"
```
oshoval, line 14: in case the above command failed, this one wont be reached, so maybe u want to reorder ?

automation/check-patch.e2e-multus-functests.sh
```diff
@@ -0,0 +1,91 @@
...
+}
+
+function __get-tools() {
+    if [ ! -d bin ]; then
+        mkdir bin
+    fi
+
+    echo "install jq"
+    curl -Lo ./bin/jq https://github.com/stedolan/jq/releases/download/jq-1.6/jq-linux64
+    chmod +x ./bin/jq
```
oshoval, line 24: u can use container of jq please

automation/check-patch.e2e-multus-functests.sh
```diff
@@ -0,0 +1,91 @@
...
+function __remove-test-manifests() {
+    echo "Remove test manifests"
+    kubectl delete -f ${TEST_YAML}
+    kubectl delete -f ${CNI_PLUGIN_YAML}
+}
+
+main() {
+    # Setup CNAO and artifacts temp directory
+    source automation/check-patch.setup.sh
+    cd ${TMP_PROJECT_PATH}
```
oshoval, line 68: where TMP_PROJECT_PATH is declared?
maybe u want set -u , so it will fail if its not declared?

its ok i guess that there is TMP_COMPONENT_PATH and TMP_PROJECT_PATH right?

automation/check-patch.e2e-multus-functests.sh
```diff
@@ -0,0 +1,91 @@
...
+
+    # Spin-up ephemeral cluster with latest CNAO, 2 nodes and secondary NIC needed for test
+    export KUBEVIRT_NUM_NODES=2
+    export KUBEVIRT_NUM_SECONDARY_NICS=1
+    # this script also exports KUBECONFIG, and fetch $COMPONENT repository
+    COMPONENT="multus" source automation/components-functests.setup.sh
+
+    trap teardown EXIT
```
oshoval, line 76: aint the trap should be before line 74 ?
you might want to add SIGINT

automation/check-patch.e2e-multus-functests.sh
```diff
@@ -0,0 +1,91 @@
+#!/bin/bash -xe
```
oshoval, line 1: u might want to add -u (see comment about it below),
and -E because u use '(' subshell

automation/check-patch.e2e-multus-functests.sh
RamLavi: jq is used inside the multus source test. changing it to dockerized jq will require changing it in multus repo first and only I can remove this tool installation here, so in any case it shouldn't be in this PR context.

automation/check-patch.e2e-multus-functests.sh
RamLavi: > where TMP_PROJECT_PATH is declared?

inside automation/check-patch.setup.sh, which is used for all automation scripts.

> maybe u want set -u , so it will fail if its not declared?

will do

> 
> its ok i guess that there is TMP_COMPONENT_PATH and TMP_PROJECT_PATH right?

yes, the former refers multus/kmp/... and the latter refers to cnao itself

automation/check-patch.e2e-multus-functests.sh
RamLavi: no, only `automation/components-functests.setup.sh` creates the cluster.
IIUC only EXIT is required.

automation/check-patch.e2e-multus-functests.sh
RamLavi: Done

automation/check-patch.e2e-multus-functests.sh
RamLavi: Done

automation/check-patch.e2e-multus-functests.sh
oshoval: ctrl c, will work just if SIGINT is used, from what i remember
up to you

RamLavi:
@oshoval fixes [here](https://github.com/kubevirt/cluster-network-addons-operator/compare/2804dfdea09b006973d3b101e8c15d9948658050..803beb4cc90c95b1e5e0b275e853acbb45bb5f24)

automation/check-patch.e2e-multus-functests.sh
```diff
@@ -0,0 +1,92 @@
+#!/bin/bash -xeuE
+
+# This script should be able to execute multus
```
oshoval, line 3: if you want, add the license please, depends what is the de facto standard on this repo

automation/check-patch.e2e-multus-functests.sh
```diff
@@ -0,0 +1,92 @@
...
+
+teardown() {
+    rm -rf "${TMP_COMPONENT_PATH}"
+    make cluster-down
+
+}
+
+function __get-tools() {
+    if [ ! -d bin ]; then
+        mkdir bin
```
oshoval, line 20: u can simply mkdir -p bin please, and no if needed

automation/check-patch.e2e-multus-functests.sh
oshoval: but if 
no, only automation/components-functests.setup.sh creates the cluster.
failed, u do want to cluster-down as some containers might already exists

RamLavi: in bash EXIT works also for `ctrl c`,(implied [here](https://unix.stackexchange.com/questions/57940/trap-int-term-exit-really-necessary) and also tested locally)
unless this script runs on dash, it should work.

oshoval: ye but on prow we have dash for example, this is exactly why some scripts / PR use SIGINT as well
btw i think for Or it didnt work on bash in one of his PRs

also you can add /bin/bash to the shebang it would be better to enforce bash for it imo

RamLavi:
/retest

unrelated flaky

automation/check-patch.e2e-multus-functests.sh
RamLavi: ack, if it makes it more robust then I'm all forward
Done

automation/check-patch.e2e-multus-functests.sh
RamLavi: licence header is not used throughout the repo, if we want to add one, we should should discuss it. if decided that we want it, should be added in a separate PR for all scripts&code IMO

automation/check-patch.e2e-multus-functests.sh
RamLavi: Done

RamLavi:
@oshoval fixes2 [here](https://github.com/kubevirt/cluster-network-addons-operator/compare/803beb4cc90c95b1e5e0b275e853acbb45bb5f24..8e93219f000732b4af5fb43d0fb4183a95eed070)

oshoval:
> @oshoval fixes2 [here](https://github.com/kubevirt/cluster-network-addons-operator/compare/803beb4cc90c95b1e5e0b275e853acbb45bb5f24..8e93219f000732b4af5fb43d0fb4183a95eed070)

Thanks
can u see this please ?
https://github.com/kubevirt/cluster-network-addons-operator/pull/631#discussion_r512450188

automation/check-patch.e2e-multus-functests.sh
RamLavi: >>no, only automation/components-functests.setup.sh creates the cluster.
>
>but if that setup failed
>failed, u do want to cluster-down as some containers might already exists

automation/components-functests.setup.sh has it's own [teardown](https://github.com/kubevirt/cluster-network-addons-operator/blob/master/automation/components-functests.setup.sh#L29).

oshoval: interesting point, dont have yet an answer

wont it cluster-down when it finishes ? leaving the rest of the code without a cluster?

RamLavi: it does not, otherwise the test would have failed locally

oshoval: of course, i understand, do you know why ?
afterall it has EXIT which make cluster-down, maybe in case of a success it wont run it ?
(i dont remember)

anyway lets not block on it i guess

automation/check-patch.e2e-multus-functests.sh
```diff
@@ -0,0 +1,91 @@
...
+
+set -xeuE
+
+# This script should be able to execute multus
+# functional tests against Kubernetes cluster with
+# CNAO built with latest changes, on any
+# environment with basic dependencies listed in
+# check-patch.packages installed and docker running.
+#
+# yum -y install automation/check-patch.packages
```
oshoval, line 11: nit:
maybe add "Usage:"

automation/check-patch.e2e-multus-functests.sh
```diff
@@ -0,0 +1,91 @@
...
+teardown() {
+    rm -rf "${TMP_COMPONENT_PATH}"
+    make cluster-down
+}
+
+function __get-tools() {
+    mkdir -p bin
+
+    echo "install jq"
+    curl -Lo ./bin/jq https://github.com/stedolan/jq/releases/download/jq-1.6/jq-linux64
```
oshoval, line 23: maybe check if no jq already ? 
else it would just curl it each run

automation/check-patch.e2e-multus-functests.sh
RamLavi: not sure what you mean, the script is no different from other automation scripts: one simply runs the script

oshoval: just because you wrote lines 11 and 12
so just meant to add "title"
```
Usage:
yum -y install automation/check-patch.packages
....
```

RamLavi: ack. got it. thanks!

automation/check-patch.e2e-multus-functests.sh
RamLavi: Done

RamLavi:
@oshoval fixes3 [here](https://github.com/kubevirt/cluster-network-addons-operator/compare/8e93219f000732b4af5fb43d0fb4183a95eed070..8c1fee05710d8493a3fa6689c1dc695b375eb294)

oshoval (approved):
Thanks

oshoval:
the approve didnt add lgtm
/lgtm

RamLavi:
/hold

we should make sure this script is not broken by  #635 changes

automation/check-patch.e2e-multus-functests.sh
```diff
@@ -0,0 +1,98 @@
...
+
+teardown() {
+    rm -rf "${TMP_COMPONENT_PATH}"
+    make cluster-down
+}
+
+function __get-tools() {
+    mkdir -p bin
+
+    JQ_PATH=./bin/jq
```
phoracek, line 23: Should not we make this into a absolute path under TMP_PROJECT_PATH?. This seems that the helper function would just create a bin directory somewhere and then we need to figure out where.

Could you think about a way to share the path with the caller? Or the caller could pass the path where to install the helper.

Well, what I mind is that it creates **something somewhere**.

And who adds JQ_PATH under the PATH?

RamLavi: it's supposed to be inside the multus repo: `${TMP_COMPONENT_PATH}/e2e`

>And who adds JQ_PATH under the PATH?

[well](https://github.com/intel/multus-cni/blob/master/e2e/get_tools.sh)... :)


anyway I'll change it  to be absolute for better readability

phoracek (approved):
/lgtm
/approve

PR #634, opened 2020-10-28
tier1, Add kubernetes nmstate tier1 automation test

RamLavi:
**What this PR does / why we need it**:
This PR adds a script that runs nmstate e2e test on CNAO cluster

**Special notes for your reviewer**:

**Release note**:

```release-note
NONE
```

PR #633, opened 2020-10-28
tier1, Add bridge-marker tier1 automation test

RamLavi:
**What this PR does / why we need it**:
this PR adds a script that runs bridge-marker e2e test on CNAO cluster

**Special notes for your reviewer**:
This PR depends on bridge-marker [PR](https://github.com/kubevirt/bridge-marker/pull/20) merger.
**Release note**:

```release-note
NONE
```

PR #635, opened 2020-10-29
fix cluster scripts

RamLavi:
This PR adds a few fixes to cluster folder, as a preliminary work for future tier1 tests PRs: #633, #634, #631.

- **cluster scripts, Fix scripts to work from remote paths**
Currently, kubectl.sh, up.sh, down.sh and cli.sh scripts cannot work
from remote paths, since they depend to be working
from CNAO's repo root path.
This was not an issue until now, since script were
only operated from Makefile targets.
This currently prevents us from running these scripts
from another folder, a functionality needed for tier1
tests.
Fixed the scripts to use the script path instead of
relative path.

- **cluster scripts, Fix ambiguous between cluster and vendored _kubevirtci folders**
Currently, both _kubecirtci/ and cluster/ scripts
use the KUBEVIRTCI_PATH var. The issue is that they
point to different paths.
This was not an issue until now because they are only
used internally.
However if one tries to set this variable and calls a script
that use both scripts (for example cluster/kubectl.sh) then the
script breaks due to this mismatch.
kubevirtci.sh script namings further contibute to this ambiguity.
-- Changed cluster/ scripts to use different variable name: CLUSTER_PATH
-- Changed kubevirtci.sh script name to cluster.sh
-- Fixed script function names accordingly: cluster::<function-name>

- **cluster-scripts, Add ssh.sh script**
In order to use ssh.sh in future tier1
tests, we add the ssh.sh script.

- **cluster scripts, Fix kubevirtci.sh to remove _kubevirtci folder**
Currently, when kubevirtci commit version is changed, we get
an error when accessing kubevirtci::install.
Added a fix that will clean up the old _kubevirtci folder

**Release note**:

```release-note
NONE
```

### DIFF
--- a/automation/components-functests.setup.sh
+++ b/automation/components-functests.setup.sh
@@ -19,7 +19,7 @@ set -exu
 
 source hack/components/git-utils.sh
 source hack/components/yaml-utils.sh
-source cluster/kubevirtci.sh
+source cluster/cluster.sh
 
 teardown() {
     make cluster-down
@@ -34,7 +34,7 @@ make cluster-down cluster-up
 
 # Export .kubeconfig full path, so it will be possible
 # to use 'kubectl' directly from the component directory path
-export KUBECONFIG=$(kubevirtci::kubeconfig)
+export KUBECONFIG=$(cluster::kubeconfig)
 
 # Deploy CNAO latest changes
 make cluster-operator-push
--- a/automation/components-functests.setup.sh
+++ b/automation/components-functests.setup.sh
@@ -41,7 +41,7 @@ make cluster-operator-push
 make cluster-operator-install
 
 # Deploy all network addons components with CNAO
-    cat <<EOF | kubectl apply -f -
+    cat <<EOF | cluster/kubectl.sh apply -f -
 apiVersion: networkaddonsoperator.network.kubevirt.io/v1alpha1
 kind: NetworkAddonsConfig
 metadata:
@@ -58,7 +58,7 @@ spec:
   imagePullPolicy: Always
 EOF
 
-kubectl wait networkaddonsconfig cluster --for condition=Available --timeout=13m
+cluster/kubectl.sh wait networkaddonsconfig cluster --for condition=Available --timeout=13m
 
 # Clone component repository
 component_url=$(yaml-utils::get_component_url ${COMPONENT})
--- a/cluster/cli.sh
+++ b/cluster/cli.sh
@@ -15,7 +15,7 @@
 # limitations under the License.
 
 SCRIPTS_PATH="$(dirname "$(realpath "$0")")"
-source ${SCRIPTS_PATH}/kubevirtci.sh
-kubevirtci::install
+source ${SCRIPTS_PATH}/cluster.sh
+cluster::install
 
-$(kubevirtci::path)/cluster-up/cli.sh "$@"
+$(cluster::path)/cluster-up/cli.sh "$@"
--- a/cluster/cli.sh
+++ b/cluster/cli.sh
@@ -14,7 +14,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-source ./cluster/kubevirtci.sh
+SCRIPTS_PATH="$(dirname "$(realpath "$0")")"
+source ${SCRIPTS_PATH}/kubevirtci.sh
 kubevirtci::install
 
 $(kubevirtci::path)/cluster-up/cli.sh "$@"
--- a/cluster/cluster.sh
+++ b/cluster/cluster.sh
@@ -15,10 +15,26 @@
 export KUBEVIRT_PROVIDER=${KUBEVIRT_PROVIDER:-'k8s-1.17'}
 
 KUBEVIRTCI_VERSION='0b941ea5dc647d3aea6f6a6fff95563d6ce0445e'
+KUBEVIRTCI_REPO='https://github.com/kubevirt/kubevirtci.git'
 # The CLUSTER_PATH var is used in cluster folder and points to the _kubevirtci where the cluster is deployed from.
 CLUSTER_PATH=${CLUSTER_PATH:-"${PWD}/_kubevirtci/"}
 
+function cluster::_get_repo() {
+    git --git-dir ${CLUSTER_PATH}/.git remote get-url origin
+}
+
+function cluster::_get_version() {
+    git --git-dir ${CLUSTER_PATH}/.git log --format="%H" -n 1
+}
+
 function cluster::install() {
+    # Remove cloned kubevirtci repository if it does not match the requested one
+    if [ -d ${CLUSTER_PATH} ]; then
+        if [ $(cluster::_get_repo) != ${KUBEVIRTCI_REPO} -o $(cluster::_get_version) != ${KUBEVIRTCI_VERSION} ]; then
+            rm -rf ${CLUSTER_PATH}
+        fi
+    fi
+
     if [ ! -d ${CLUSTER_PATH} ]; then
         git clone https://github.com/kubevirt/kubevirtci.git ${CLUSTER_PATH}
         (
--- a/cluster/cluster.sh
+++ b/cluster/cluster.sh
@@ -15,22 +15,23 @@
 export KUBEVIRT_PROVIDER=${KUBEVIRT_PROVIDER:-'k8s-1.17'}
 
 KUBEVIRTCI_VERSION='0b941ea5dc647d3aea6f6a6fff95563d6ce0445e'
-KUBEVIRTCI_PATH="${PWD}/_kubevirtci"
+# The CLUSTER_PATH var is used in cluster folder and points to the _kubevirtci where the cluster is deployed from.
+CLUSTER_PATH=${CLUSTER_PATH:-"${PWD}/_kubevirtci/"}
 
-function kubevirtci::install() {
-    if [ ! -d ${KUBEVIRTCI_PATH} ]; then
-        git clone https://github.com/kubevirt/kubevirtci.git ${KUBEVIRTCI_PATH}
+function cluster::install() {
+    if [ ! -d ${CLUSTER_PATH} ]; then
+        git clone https://github.com/kubevirt/kubevirtci.git ${CLUSTER_PATH}
         (
-            cd ${KUBEVIRTCI_PATH}
+            cd ${CLUSTER_PATH}
             git checkout ${KUBEVIRTCI_VERSION}
         )
     fi
 }
 
-function kubevirtci::path() {
-    echo -n ${KUBEVIRTCI_PATH}
+function cluster::path() {
+    echo -n ${CLUSTER_PATH}
 }
 
-function kubevirtci::kubeconfig() {
-    echo -n ${KUBEVIRTCI_PATH}/_ci-configs/${KUBEVIRT_PROVIDER}/.kubeconfig
+function cluster::kubeconfig() {
+    echo -n ${CLUSTER_PATH}/_ci-configs/${KUBEVIRT_PROVIDER}/.kubeconfig
 }
--- a/cluster/down.sh
+++ b/cluster/down.sh
@@ -16,7 +16,8 @@
 
 set -ex
 
-source ./cluster/kubevirtci.sh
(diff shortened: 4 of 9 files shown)


cluster/cli.sh
```diff
@@ -14,7 +14,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-source ./cluster/kubevirtci.sh
+SCRIPTPATH="$( cd "$(dirname "$0")" >/dev/null 2>&1 ; pwd -P )"
```
phoracek, line 17: SCRIPTS_PATH

cluster/kubevirtci.sh
```diff
@@ -15,22 +15,39 @@
 export KUBEVIRT_PROVIDER=${KUBEVIRT_PROVIDER:-'k8s-1.17'}
 
 KUBEVIRTCI_VERSION='0b941ea5dc647d3aea6f6a6fff95563d6ce0445e'
-KUBEVIRTCI_PATH="${PWD}/_kubevirtci"
+KUBEVIRTCI_REPO='https://github.com/kubevirt/kubevirtci.git'
+KUBEVIRTCI_PATH=${KUBEVIRTCI_PATH:-"${PWD}/_kubevirtci/cluster-up/"}
+KUBEVIRTCI_PARENT_PATH="$(dirname "$KUBEVIRTCI_PATH")"
```
phoracek, line 20: It is weird that you setup full path on the previous line and then shorten it here. I'd expect first to setup the parent and then just extend it with the subdirectory

phoracek, line 20: This is KUBEVIRTCI_PATH and the one above is CLUSTER_UP_PATH, no?

cluster/kubevirtci.sh
```diff
@@ -15,22 +15,39 @@
...
+
+function kubevirtci::_get_version() {
+    git --git-dir ${KUBEVIRTCI_PARENT_PATH}/.git log --format="%H" -n 1
+}
 
 function kubevirtci::install() {
-    if [ ! -d ${KUBEVIRTCI_PATH} ]; then
-        git clone https://github.com/kubevirt/kubevirtci.git ${KUBEVIRTCI_PATH}
+    # Remove cloned kubevirtci repository if it does not match the requested one
+    if [ -d ${KUBEVIRTCI_PARENT_PATH} ]; then
```
phoracek, line 32: nice!

cluster/kubevirtci.sh
```diff
@@ -15,22 +15,39 @@
 export KUBEVIRT_PROVIDER=${KUBEVIRT_PROVIDER:-'k8s-1.17'}
 
 KUBEVIRTCI_VERSION='0b941ea5dc647d3aea6f6a6fff95563d6ce0445e'
-KUBEVIRTCI_PATH="${PWD}/_kubevirtci"
+KUBEVIRTCI_REPO='https://github.com/kubevirt/kubevirtci.git'
+KUBEVIRTCI_PATH=${KUBEVIRTCI_PATH:-"${PWD}/_kubevirtci/cluster-up/"}
```
phoracek, line 19: We are not using this anywhere, are we?

phoracek (changes requested):

cluster/kubevirtci.sh
RamLavi: I know what you mean,  let me explain my reasoning:
_kubevirtci is using the path with `cluster-up/`, like so:
- _kubevirtci's `KUBEVIRTCI_PATH`: `/tmp/cnao/cluster-network-addons-operator/_kubevirtci/cluster-up/`
- cluster's `KUBEVIRTCI_PATH`: `/tmp/cnao/cluster-network-addons-operator/_kubevirtci/`

it would be more reasonable to change `KUBEVIRTCI_PATH` to the one `cluster` is using, but this would mean doing this change in kubevirtci, which would potentially break every repo using it. 

So I ended up thinking it's not worth it to change it there and I should just make it here like so.
What do you think? should I fix it in kubevirtci?

RamLavi: yes, but for the same [reasoning](https://github.com/kubevirt/cluster-network-addons-operator/pull/635#discussion_r514416276)  - I thought to call it like so.

cluster/kubevirtci.sh
RamLavi: took it from @qinqon's knmstate :)

cluster/kubevirtci.sh
RamLavi: `KUBEVIRTCI_PATH` is used in `cluster`'s `kubevirtci.sh` and in `_kubevirtci` folder.

cluster/cli.sh
RamLavi: Done

cluster/kubevirtci.sh
```diff
@@ -15,22 +15,39 @@
...
-            cd ${KUBEVIRTCI_PATH}
+            cd ${KUBEVIRTCI_PARENT_PATH}
             git checkout ${KUBEVIRTCI_VERSION}
         )
     fi
 }
 
 function kubevirtci::path() {
-    echo -n ${KUBEVIRTCI_PATH}
+    echo -n ${KUBEVIRTCI_PARENT_PATH}
```
phoracek, line 48: This is part of the confusion for me. kubevirtci::path does not return KUBEVIRTCI_PATH.

To make the distinction clear, would it make sense to rename this script to cluster.sh and all kubevirtci:: as well?

Overloading of the name kubevirtci is killing me here.

cluster/kubevirtci.sh
phoracek: Please add a comment noting that this variable is used only by vendored kubevirtci scripts. And make sure that we don't use in in any of our scripts.

cluster/kubevirtci.sh
phoracek: Can we rename this to CLUSTER_PATH then?

phoracek (changes requested):

cluster/kubevirtci.sh
RamLavi: Done

cluster/kubevirtci.sh
RamLavi: I think `CLUSTER_KUBEVIRTCI_PATH` would be better, as it points to the _kubevirtci where the cluster is spun. tell me what you think

RamLavi: wait.. the vendored already iuse `KUBEVIRTCI_CLUSTER_PATH` .. 
they took all the d=good names :) 

let's stick to `CLUSTER_PATH` then

RamLavi:
/retest
flaky lot

cluster/cluster.sh
```diff
@@ -15,11 +15,27 @@
...
 function cluster::install() {
+    # Remove cloned kubevirtci repository if it does not match the requested one
+    if [ -d ${CLUSTER_PATH} ]; then
+        if [ $(cluster::_get_repo) != ${KUBEVIRTCI_REPO} -o $(cluster::_get_version) != ${KUBEVIRTCI_VERSION} ]; then
+            rm -rf ${CLUSTER_PATH}
+        fi
+    fi
+
     if [ ! -d ${CLUSTER_PATH} ]; then
         git clone https://github.com/kubevirt/kubevirtci.git ${CLUSTER_PATH}
```
ormergi, line 40: Hey we can optimize this to get only the latest commit with adding `--branch master --depth 1`

ormergi (commented):
Good catch!
You can simplify SCRIPTS_PATH with `$(dirname "$(realpath "$0")")`

cluster/cluster.sh
RamLavi: when I tried it I got an error:
```
+ git clone --branch master --depth 1 https://github.com/kubevirt/kubevirtci.git /home/ralavi/go/src/github.com/kubevirt/cluster-network-addons-operator/_kubevirtci/
Cloning into '/home/ralavi/go/src/github.com/kubevirt/cluster-network-addons-operator/_kubevirtci'...
remote: Enumerating objects: 1874, done.
remote: Counting objects: 100% (1874/1874), done.
remote: Compressing objects: 100% (1358/1358), done.
remote: Total 1874 (delta 565), reused 1543 (delta 445), pack-reused 0
Receiving objects: 100% (1874/1874), 3.57 MiB | 717.00 KiB/s, done.
Resolving deltas: 100% (565/565), done.
+ cd /home/ralavi/go/src/github.com/kubevirt/cluster-network-addons-operator/_kubevirtci/
+ git checkout 0b941ea5dc647d3aea6f6a6fff95563d6ce0445e
fatal: reference is not a tree: 0b941ea5dc647d3aea6f6a6fff95563d6ce0445e
make: *** [Makefile:123: cluster-up] Error 128

```

RamLavi:
@ormergi changes [here](https://github.com/kubevirt/cluster-network-addons-operator/compare/6d24dbfc5d8a61c5db5f23726a76992afc5d20dd..16fe247ea10df2a2a9d3df245840b154e078e7b1)

cluster/cluster.sh
ormergi: Thanks for trying!

ormergi (approved):
Thanks

automation/components-functests.setup.sh
```diff
@@ -34,14 +34,14 @@ make cluster-down cluster-up
 
 # Export .kubeconfig full path, so it will be possible
```
phoracek, line 35: KUBECONFIG is not needed since you switched from kubectl to cluster/kubectl.sh, is it?

cluster/cluster.sh
```diff
@@ -0,0 +1,54 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+export KUBEVIRT_PROVIDER=${KUBEVIRT_PROVIDER:-'k8s-1.17'}
+
+KUBEVIRTCI_VERSION='0b941ea5dc647d3aea6f6a6fff95563d6ce0445e'
+KUBEVIRTCI_REPO='https://github.com/kubevirt/kubevirtci.git'
+# KUBEVIRTCI_PATH var shoud be used only in vendored _kubecirtci folder
```
phoracek, line 19: You are talking about KUBEVIRTCI_PATH, but I don't see such thing around

phoracek (changes requested):

automation/components-functests.setup.sh
RamLavi: This is actually needed, some tests simply need KUBECONFIG (for example [knmstate](https://github.com/nmstate/kubernetes-nmstate/blob/master/hack/run-e2e-test-handler.sh#L4) and [ovs-cni](https://github.com/kubevirt/ovs-cni/blob/master/hack/functests.sh#L23))

cluster/cluster.sh
RamLavi: well yes, you asked me to remove them entirely from cluster folder. :) 
I left this comment for so that future us might avoid using it again..

automation/components-functests.setup.sh
phoracek: But it is not a general requirement for all the e2e lanes. Would it make sense to have this export in each component-specific script? Keep it close to the user and don't turn this into a junkyard of all that may be needed.

cluster/cluster.sh
phoracek: Documenting all things we should not do would get out of hand pretty fast. You see that this one got out of context for me even before we merged it... I'd prefer to drop it and repeat our conversation when we attempt to use this variable here in the future again.

RamLavi: Done

automation/components-functests.setup.sh
RamLavi: I agree, but keep in mind this script only runs on tier1 test scripts. 
it's a general requirement for tier1 tests, since they all run on external cluster (CNAO's)

phoracek: Well, I don't have a terribly strong argument against it. Let's go with your proposal then.

phoracek (approved):
/lgtm
/approve

PR #637, opened 2020-11-02
bumper, Add local git rebase action in bumper gitAction Job

RamLavi:
Currently, if we re-run a gitAction job, it is not rebased
to the most updated master.
Added a local rebase action to fix that.

This PR may break the bumper script, but it's necessary to debug this functionality on the repo
**Release note**:

```release-note
NONE
```

### DIFF
--- a/.github/workflows/component-bumper.yml
+++ b/.github/workflows/component-bumper.yml
@@ -13,6 +13,11 @@ jobs:
 
     - name: Check out code into the Go module directory
       uses: actions/checkout@v2
+      with:
+        ref: master
+
+    - name: Pull latest master
+      run:  git pull --ff-only --rebase origin master
 
     - name: Run bumper script
       run: make ARGS="-config-path=components.yaml -token=${{ secrets.GITHUB_TOKEN }}" auto-bumper


.github/workflows/component-bumper.yml
```diff
@@ -13,6 +13,11 @@ jobs:
 
     - name: Check out code into the Go module directory
       uses: actions/checkout@v2
+      with:
+        ref: master
+
+    - name: Pull latest master
+      run: git pull --rebase origin master
```
phoracek, line 20: I hate rebase commits. Would it make sense to go with --only-ff or do you see it bringing and complications?

.github/workflows/component-bumper.yml
```diff
@@ -13,6 +13,11 @@ jobs:
 
     - name: Check out code into the Go module directory
       uses: actions/checkout@v2
+      with:
+        ref: master
```
phoracek, line 17: What happens once we branch out? Do I need to change this on the branch out? Can we take the branch as a variable from actions?

phoracek (changes requested):

.github/workflows/component-bumper.yml
RamLavi: That's a great idea actually

.github/workflows/component-bumper.yml
RamLavi: do you think we would ever run bumper script from a branch that is not master?
The way I see it, the only case it might be need to do so is on the stable branch, and in that case I thought we would actually prefer to /backport it. what do you think?

.github/workflows/component-bumper.yml
qinqon: Another option is to do a range cherry-pick with all the commits on the PR.

.github/workflows/component-bumper.yml
qinqon: Bumper script on release branch is super difficult since we really don't know what components branch do we have to look at, maybe we can put that information at components.yaml.

.github/workflows/component-bumper.yml
```diff
@@ -13,6 +13,11 @@ jobs:
...
+
+    - name: Pull latest master
```
qinqon, line 19: Add a comment to remove this after everything is working fine.

qinqon (changes requested):
One question I have is what we will do when we have to fix this thing after bumper script is at production.

.github/workflows/component-bumper.yml
phoracek: components.yaml is pointing to the branch we take patches from.

On branch-out of CNAO, we have couple of options:
- Branch out a component for stable release and pin components.yaml to that branch
- Keep taking component from master and only once we are fine with the state of the stable branch, turn it "static" in components.yaml

If you expect this to be too difficult, we can limit this script to the master for now.

RamLavi: if we branch out, then the components.yaml can be different, but if we define that bumper only runs on master and only backport the component-bumps PR we consider safe for stable branch, then I'm not sure it's really needed.
Or maybe I don't catch your drift. Let's further discuss

.github/workflows/component-bumper.yml
RamLavi: oh I may have not been clear - this is not supposed to be a temporary change.
This change will allow us to re-run the same job with the latest master (it doesn't rebase automatically like prow does)

.github/workflows/component-bumper.yml
RamLavi: I think I prefer the `--only-ff` option. 
Done

.github/workflows/component-bumper.yml
phoracek: While master would follow minor releases of our sub-components, stable branch would follow patch releases. So there may be nothing to backport.

Anyway, only now I realized this PR is indeed only working with master. We can follow-up about stable branches later.

phoracek (approved):
/lgtm
/approve
/hold

Please release the hold when you and Quique are ok with this.

.github/workflows/component-bumper.yml
RamLavi: ack. I will work on it on a separate PR, after I make sure this rebase works well

PR #640, opened 2020-11-04
crd, Fix placementConfiguration validation schema

RamLavi:
**What this PR does / why we need it**:
Currently the crd lacks the validation schema of
placementConfiguration. This is because the crd
is not automatically generated at the moment.
Fixed placementConfiguration validation schema.

This PR mimics the output of the auto-generated crd, as displayed in this [PR](https://github.com/kubevirt/cluster-network-addons-operator/pull/577)

**Special notes for your reviewer**:

**Release note**:
```release-note
NONE
```

PR #641, opened 2020-11-04
Allow day2 placement config changes

RamLavi:
Currently placement-configuration should not allow
day2 changes.
Changed to allow day2 changes

This PR depends on placement-configuration PR #640 to be merged.
**Release note**:

```release-note
Allow placement-configuration day2 changes
```

### DIFF
--- a/pkg/network/placement_configuration.go
+++ b/pkg/network/placement_configuration.go
@@ -1,16 +1,13 @@
 package network
 
 import (
-	"reflect"
-
-	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
 
 	cnao "github.com/kubevirt/cluster-network-addons-operator/pkg/apis/networkaddonsoperator/shared"
 )
 
-var (
-	defaultPlacementConfiguration = cnao.PlacementConfiguration{
+func GetDefaultPlacementConfiguration() cnao.PlacementConfiguration {
+	return cnao.PlacementConfiguration{
 		Infra: &cnao.Placement{
 			NodeSelector: map[string]string{
 				"beta.kubernetes.io/arch":        "amd64",
@@ -36,32 +33,19 @@ var (
 			},
 		},
 	}
-)
-
-func fillDefaultsPlacementConfiguration(conf, previous *cnao.NetworkAddonsConfigSpec) []error {
-	if conf.PlacementConfiguration == nil || conf.PlacementConfiguration.Infra == nil || conf.PlacementConfiguration.Workloads == nil {
-		if previous != nil && previous.PlacementConfiguration != nil {
-			conf.PlacementConfiguration = previous.PlacementConfiguration
-			return []error{}
-		}
-
-		if conf.PlacementConfiguration == nil {
-			conf.PlacementConfiguration = &defaultPlacementConfiguration
-			return []error{}
-		}
-		if conf.PlacementConfiguration.Infra == nil {
-			conf.PlacementConfiguration.Infra = defaultPlacementConfiguration.Infra
-		}
-		if conf.PlacementConfiguration.Workloads == nil {
-			conf.PlacementConfiguration.Workloads = defaultPlacementConfiguration.Workloads
-		}
-	}
-	return []error{}
 }
 
-func changeSafePlacementConfiguration(prev, next *cnao.NetworkAddonsConfigSpec) []error {
-	if prev.PlacementConfiguration != nil && next.PlacementConfiguration != nil && !reflect.DeepEqual(prev.PlacementConfiguration, next.PlacementConfiguration) {
-		return []error{errors.Errorf("cannot modify PlacementConfiguration once it is deployed")}
+func fillDefaultsPlacementConfiguration(conf, previous *cnao.NetworkAddonsConfigSpec) []error {
+	defaultPlacementConfiguration := GetDefaultPlacementConfiguration()
+	if conf.PlacementConfiguration == nil {
+		conf.PlacementConfiguration = &defaultPlacementConfiguration
+		return []error{}
+	}
+	if conf.PlacementConfiguration.Infra == nil {
+		conf.PlacementConfiguration.Infra = defaultPlacementConfiguration.Infra
+	}
+	if conf.PlacementConfiguration.Workloads == nil {
+		conf.PlacementConfiguration.Workloads = defaultPlacementConfiguration.Workloads
 	}
 	return []error{}
 }
--- a/test/check/check.go
+++ b/test/check/check.go
@@ -100,6 +100,26 @@ func CheckConfigCondition(gvk schema.GroupVersionKind, conditionType ConditionTy
 	}
 }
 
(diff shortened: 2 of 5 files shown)


test/e2e/workflow/deployment_test.go
```diff
@@ -133,6 +133,57 @@ var _ = Describe("NetworkAddonsConfig", func() {
...
+
+				By("Checking PlacementConfiguration was updated")
+				config := GetConfig(gvk)
+				Expect(ConvertToConfigV1(config).Spec.PlacementConfiguration).To(Equal(currentPlacementConfiguration), "Should succeed deploying PlacementConfiguration on non pre-existing config")
+			})
+
+			It("should be able to update PlacementConfigurations", func() {
+				updatedPlacementConfiguration := &cnao.PlacementConfiguration{
+					Workloads: &cnao.Placement{ NodeSelector: map[string]string{
+						"kubernetes.io/hostname": "node02"},
```
phoracek, line 165: What if somebody runs our tests with a single node? I believe that is the default on `make cluster-up`. Are we even setting 2 nodes on our CI?

Would you please add check raising an error if we don't have enough nodes to https://github.com/kubevirt/cluster-network-addons-operator/blob/master/test/e2e/workflow/main_test.go.

phoracek (changes requested):

test/e2e/workflow/deployment_test.go
RamLavi: I agree, I can simply change it to `node01` too, the only thing that matters for this test is that the change sticks

phoracek: Sounds good enough

test/e2e/workflow/deployment_test.go
```diff
@@ -133,6 +133,57 @@ var _ = Describe("NetworkAddonsConfig", func() {
 			components = append(components, MacvtapComponent)
 			testConfigUpdate(gvk, configSpec, components)
 		})
+		Context("and PlacementConfiguration is deployed", func() {
```
phoracek, line 136: Heh, I missed something. This new test is kind of redundant since it is already covered in the unit test.

What we should check here is deployment of some components. Deploy it with the default placement and then with an alternative placement, check that all the components are still ok and we turn available.

Can we add such a test? My main worry is that some of the components would fail to adapt to the new placement OR that our conditions would not be able to track removal of components with the old placement and deployment of the new ones.

RamLavi: the motivation for adding this test was that the unit test didn't catch that PlacementConfiguration was actually broken on master branch until fixed [here](https://github.com/kubevirt/cluster-network-addons-operator/pull/640).

But we can do that also by adding it to the existing test deploying it with the other components.
Let me know what you think

rhrazdil: Do we want the change of placementConfiguration to immediately apply on components pods? I imagine that users may expect that.
Currently though, I think that updating placementConfiguration will only affect future component pods, and the existing will remain as before.

phoracek: @rhrazdil  We apply it immediately on all deployed components.  When the CR is changed, we reconcile it, render new objects for all components and apply them.

phoracek: @RamLavi why adding to an existing one? If this was indeed an issue, let's keep the test. However, I still miss a test showing that we are able to apply the changes on live components, turn "Progressing" and eventually become "Available" again.

RamLavi: ack, will add more components to the original test..

RamLavi: Done

rhrazdil (changes requested):

test/check/check.go
```diff
@@ -100,6 +100,26 @@ func CheckConfigCondition(gvk schema.GroupVersionKind, conditionType ConditionTy
 	}
 }
 
+func GetPlacementFromComponentDaemonSet(component Component) ([]cnao.Placement, error) {
```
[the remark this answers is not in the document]
RamLavi: Done, thanks!

test/e2e/workflow/deployment_test.go
```diff
@@ -133,22 +135,62 @@ var _ = Describe("NetworkAddonsConfig", func() {
...
+			}
+			configSpec := cnao.NetworkAddonsConfigSpec{
+				LinuxBridge: &cnao.LinuxBridge{},
+				Multus:      &cnao.Multus{},
+				Ovs:         &cnao.Ovs{},
+				MacvtapCni:  &cnao.MacvtapCni{},
+				PlacementConfiguration: &cnao.PlacementConfiguration{},
+			}
+			checkWorkloadPlacementOnComponents := func(component []Component, expectedWorkLoadPlacement cnao.Placement) {
+				for _, component := range components {
```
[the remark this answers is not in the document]
RamLavi: dumb mistake on my part. `components` is already accessible to the helper function
Done

pkg/network/placement_configuration.go
```diff
@@ -1,16 +1,13 @@
...
-
-	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
 
 	cnao "github.com/kubevirt/cluster-network-addons-operator/pkg/apis/networkaddonsoperator/shared"
 )
 
 var (
-	defaultPlacementConfiguration = cnao.PlacementConfiguration{
+	DefaultPlacementConfiguration = cnao.PlacementConfiguration{
```
phoracek, line 10: Why are you making a global variable public? It's dangerous, what if somebody changes the value? Global hell.

Please turn it into a function returning the default structure, to avoid mutability.

test/e2e/workflow/deployment_test.go
```diff
@@ -133,22 +135,62 @@ var _ = Describe("NetworkAddonsConfig", func() {
...
+
+				By("Checking PlacementConfiguration was set on components to updated workload PlacementConfiguration")
+				checkWorkloadPlacementOnComponents(*configSpec.PlacementConfiguration.Workloads)
+			})
+		})
 	})
 
 	Context("when all components are already deployed", func() {
 		components := []Component{
-			MultusComponent,
```
phoracek, line 140: Why are you changing this test?

phoracek (changes requested):

test/e2e/workflow/deployment_test.go
RamLavi: totally my bad, fixed :)

pkg/network/placement_configuration.go
RamLavi: [Done](https://github.com/kubevirt/cluster-network-addons-operator/compare/fae80f26414a0483f7945b258ee2d805475afad6..fa1df96d89d5c6fad893209dad371683d7f80781)

phoracek (approved):
/lgtm
/approve

Thanks

Issue #650, opened 2020-11-10
[flaky ci] NetworkAddonsConfig when a valid config is deployed and a component which does support removal is removed from the Spec should remain at Available condition


dhiller:
**What happened**:

Test `NetworkAddonsConfig when a valid config is deployed and a component which does support removal is removed from the Spec should remain at Available condition` flaked before PR #638 got merged

Offending run: https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_cluster-network-addons-operator/638/pull-e2e-cluster-network-addons-operator-workflow-k8s/1325803948359028736

PR history: https://prow.apps.ovirt.org/pr-history/?org=kubevirt&repo=cluster-network-addons-operator&pr=638

Test stacktrace:
```
/tmp/cnao/cluster-network-addons-operator/test/e2e/workflow/validation_test.go:41
Timed out after 120.201s.
Timed out waiting for the condition, current config:
object:
  apiVersion: networkaddonsoperator.network.kubevirt.io/v1
  kind: NetworkAddonsConfig
  metadata:
    creationTimestamp: "2020-11-09T14:27:20Z"
    generation: 1
    name: cluster
    resourceVersion: "1193"
    selfLink: /apis/networkaddonsoperator.network.kubevirt.io/v1/networkaddonsconfigs/cluster
    uid: a1d0dca2-ecb5-4d0c-91c6-fd94ad467dd5
  spec:
    linuxBridge: {}
    nmstate: {} 
```

Similar failure: https://github.com/kubevirt/cluster-network-addons-operator/issues/523

/triage build-watcher

**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:

qinqon:
new runs with the issue:
https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_cluster-network-addons-operator/656/pull-e2e-cluster-network-addons-operator-workflow-k8s/1326811082827763712
https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_cluster-network-addons-operator/656/pull-e2e-cluster-network-addons-operator-workflow-k8s/1326792460189306880
[...]

PR #665, opened 2020-11-15
delete objects, Fix tracking of already removed objects

RamLavi:
This PR fixes a transient situation where an object which just got deleted
is still tracked by the CNAO reconciler, thus causing a temporary failed status
This fix also stabilizes the flaky [issue](https://github.com/kubevirt/cluster-network-addons-operator/issues/650) reported in validation e2e tests. 

- **delete objects, Fix Tracking of already removed objects**
Currently, the reconciler updates the list of objects to deploy
only after any objects are removed. This causes a temporary period
where the object is declared as gone.
Changed the order of trackDeployedObjects and deleteObjects to
keep the tracked objects updated correctly.

- **expectAvaialableEvent, Change to Watch policy to since now**
In order to check Available event after object removal,
we move to use CheckAvailableEvent.
We also change the event watch policy to SinceNow,
since it seems that the change is sufficiently caught.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
+++ b/pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
@@ -267,6 +267,9 @@ func (r *ReconcileNetworkAddonsConfig) Reconcile(request reconcile.Request) (rec
 		return reconcile.Result{}, err
 	}
 
+	// Track state of all deployed pods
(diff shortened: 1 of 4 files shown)


RamLavi:
/hold
bug recreated

RamLavi:
/hold cancel
ran it multiple times, I think it's safe

qinqon:
/lgtm
/approve

Let's roll and break chings !!!

PR #672, opened 2020-11-24
validation_test, fix cross test flaky

RamLavi:
Sometimes we observe that the condition
set in the first test still exists when
starting the second test.
To avoid cross tests intrusions, removed
the config in AfterEach routine.

This PR is supposed to fix the new flaky reported in #650 ,where we fail on:
```
2020/11/19 13:29:37 Warning event received: Event(v1.ObjectReference{Kind:"NetworkAddonsConfig", Namespace:"", Name:"cluster", UID:"975c427d-0280-4448-aa07-b5508d99544c", APIVersion:"networkaddonsoperator.network.kubevirt.io/v1", ResourceVersion:"1257", FieldPath:""}): type: 'Warning' reason: 'Failed: FailedToValidate' Operator failed: failed to validate NetworkConfig.Spec: invalid configuration:
both or none of the KubeMacPool ranges needs to be configure
```

**Release note**:

```release-note
NONE
```

### DIFF
--- a/test/e2e/workflow/validation_test.go
+++ b/test/e2e/workflow/validation_test.go
@@ -13,6 +13,9 @@ import (
 
 var _ = Describe("NetworkAddonsConfig", func() {
 	gvk := GetCnaoV1GroupVersionKind()
+	AfterEach(func() {
+		DeleteConfig(gvk)
+	})
 	Context("when there is no running config", func() {
 		Context("and an invalid config is created", func() {
 			BeforeEach(func() {


test/e2e/workflow/validation_test.go
```diff
@@ -13,6 +13,9 @@ import (
 
 var _ = Describe("NetworkAddonsConfig", func() {
 	gvk := GetCnaoV1GroupVersionKind()
+	AfterEach(func() {
```
phoracek, line 16: This should be already done here https://github.com/kubevirt/cluster-network-addons-operator/blob/master/test/e2e/workflow/main_test.go#L59. If it is not, it is a bug in the suite cleanup, no?

phoracek (changes requested):

test/e2e/workflow/validation_test.go
oshoval, line 16: Since the Contexts that create it are the inner contexts, 
wouldn't it be better to add it for them ?
it will repeat twice, but the scope will be more accurate

oshoval: because the inner gvk wins ?
https://github.com/kubevirt/cluster-network-addons-operator/pull/672/files#diff-fddb4e4e7a7cebcb6649159ab67dfa1c36da1d4b193f1d2e19e93638c7a7b11eR15

RamLavi: or perhaps it's not a good fix at all
I think you're right Petr, will further investigate it. closing the PR in the meantime

PR #685, opened 2020-12-01
bumper, Change bumper git action jobs names

RamLavi:
I initially thought that having the same name, the two workflows (master & release-0.42 bumper jobs) will show up in the same [entree](https://github.com/kubevirt/cluster-network-addons-operator/actions/runs/397658238).
Now that I see it's not like that, I've added this PR to make the master & release-0.42 bumper jobs to have distinct names.

- **bumper, Change bumper git action jobs names**
Currently the job run has a distinctive name, but the workflow job
has the same name. Since the jobs are not grouped, then better to
give them a distinctive name as well.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/.github/workflows/component-bumper-master.yml
+++ b/.github/workflows/component-bumper-master.yml
@@ -1,4 +1,4 @@
-name: Auto-Bump Components' Versions
+name: master - Auto-Bump Components' Versions
 
 on:
   schedule:
--- a/.github/workflows/component-bumper-release-0.42.yml
+++ b/.github/workflows/component-bumper-release-0.42.yml
@@ -1,4 +1,4 @@
-name: Auto-Bump Components' Versions
+name: release-0.42 - Auto-Bump Components' Versions
 
 on:
   schedule:


RamLavi:
exploiting this PR to check KMP tier1 test is ok. will run several tests on it

qinqon:
@RamLavi  Is not github actions branch aware ? I mean we just need to a branch-out version of .github/workflow

RamLavi:
> @RamLavi Is not github actions branch aware ? I mean we just need to a branch-out version of .github/workflow

it is. sort of. workflow jobs run only from master. you can run the stable branch using checkout@v2 like I did in the release-0.42 workflow.

but this PR is not related to that, it is only so that you will be able to distinguish the 2 jobs from [here](https://github.com/kubevirt/cluster-network-addons-operator/actions), instead of needing to go inside each job to see what job it [ran](https://github.com/kubevirt/cluster-network-addons-operator/actions/runs/397658238).

qinqon:
@RamLavi .github/workflow at release branches get ignored by github ?

RamLavi:
> @RamLavi .github/workflow at release branches get ignored by github ?

yep. see [this](https://github.community/t/scheduled-job-on-specific-branches/144332)

RamLavi:
We are fine without this change. Dropping it

PR #698, opened 2020-12-17
bumper, Add quay login to bumper job

RamLavi:
when bumping linux-bridge component we need to be registered as a quay user
that has write permissions, in order to push them.
Added a step to login to quay using the "kubevirt+network" quay robot credentials.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/.github/workflows/component-bumper-release-0.42.yml
+++ b/.github/workflows/component-bumper-release-0.42.yml
@@ -14,6 +14,9 @@ jobs:
     - name: Set environment variables
       run: echo "BASE_BRANCH=release-0.42" >> $GITHUB_ENV
 
+    - name: Login to Quay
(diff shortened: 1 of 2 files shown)


.github/workflows/component-bumper-release-0.42.yml
```diff
@@ -14,6 +14,9 @@ jobs:
     - name: Set environment variables
       run: echo "BASE_BRANCH=release-0.42" >> $GITHUB_ENV
 
+    - name: Login to Quay
+      run: docker login -u="kubevirt+network" -p="${{ secrets.QUAY_ROBOT_TOKEN }}" quay.io
```
qinqon, line 18: Are we allowed to store those tokens here ?

RamLavi: why not? the secret is stored in the github repo secret section, and securely used here.

qinqon: fair enough

Issue #693, opened 2020-12-10
Autobumper has failed to pin new nmstate version


phoracek:
**What happened**:

knmstate 0.35 was released but (unlike the bridge-marker) it was not picked up by CNAO automation.

I see one failing Actions job in the history: https://github.com/kubevirt/cluster-network-addons-operator/runs/1528591039?check_suite_focus=true

RamLavi:
@phoracek, knmstate wasn't picked because linux-bridge bump failed, causing the bumper script to stop.
I initially thought it would be best to stop the bumper script at the first bump failure, do you think we should change that?

phoracek:
I see.

If the failure is caused by the vendored sub-component and not us, we should continue with the process and at the end mark the job as failed (but bump as many components as we can). We can get into a situation where we really need knmstate bumped, but another sub-component is broken.

If the failure is caused by CNAO, then we should fix it ASAP, but I don't think we'd need to gracefully ignore part of our (CNAO's) fails.

RamLavi:
fixed. closing issue

PR #708, opened 2020-12-20
bumper, Change bumper fail policy

RamLavi:
This PR is opened in relation to a issue #693 opened due to knmstate bump fail that caused the bumper to stop bumping.

- **bumper, Change bumper fail policy**
Currently, the bumper script will fail upon the first bump failure.
We want to be able to run the bumper on all components even if one
component bump is broken.
The new behavior is that all found bumps will occur, and in the end if at least one
component failed to bump - the job will exit with error.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tools/bumper/bumper.go
+++ b/tools/bumper/bumper.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"flag"
 	"fmt"
+	"strings"
 	"log"
 	"os"
 
@@ -47,6 +48,7 @@ func main() {
 		exitWithError(errors.Wrap(err, "Failed to get components config"))
 	}
 
+	var failedComponents []string
 	for componentName, component := range componentsConfig.Components {
 		logger.Printf("~~Checking if %s needs bumping~~", componentName)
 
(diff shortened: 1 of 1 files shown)


RamLavi:
irrelevant to PR

tools/bumper/bumper.go
```diff
@@ -82,12 +84,17 @@ func main() {
...
+				logger.Printf("Bump %s component was initiated but did not succeed. err = %s", componentName, err)
+				failedComponents = append(failedComponents, componentName)
 			}
 		} else {
 			logger.Printf("Bump not needed in component %s", componentName)
 		}
 	}
+
+	if len(failedComponents) != 0 {
+		exitWithError(fmt.Errorf("Not all component bumps were successful: %s", strings.Join(failedComponents, ",")))
```
qinqon, line 96:
```suggestion
		exitWithError(fmt.Errorf("Not all component bumps were successful, failed components: %s", strings.Join(failedComponents, ",")))
```

qinqon (changes requested):
nit

tools/bumper/bumper.go

PR #709, opened 2020-12-20
bumper, fix returning annotated tag sha on tagged update-policy

RamLavi:
Currently when using tagged update policy mode, the bumper script is returning
the annotated tag sha and not the relevant commit sha of this tag.
The sha of an annotated tag is different than the sha of it's matching commit,
whereas the sha of lightweight tag is the same as the sha of it's matching commit.
This will cause the future bump to fail, as the annotated tag is not directly connected
via the githubApi.
Moreover, the mocked listMatchingRefs routine did not catch the issue since it used git log, which
show that commit sha's and not the tags themselves in order to assemble the tag list.
Fixed to return the commit sha.
Fixed the listMatchingRefs mock to use 'show-ref --tag', and adjusted the test to catch this issue.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tools/bumper/component_commands.go
+++ b/tools/bumper/component_commands.go
@@ -218,7 +218,7 @@ func (componentOps *gitComponent) getLatestTaggedFromBranch(repo, owner, branch,
 
 		for _, commit := range branchCommits {
 			if commit.GetSHA() == commitShaOfTag {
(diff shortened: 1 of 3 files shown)


RamLavi:

irrelevant to PR

PR #717, opened 2020-12-22
validation test, Change test look for warning events

RamLavi:
Currently the validation test It("should remain at Available condition") watches for an
Available event after a component's removal. Since there is a current issue with watching
for repeating events [[1]](https://github.com/kubernetes/kubernetes/issues/97367#issuecomment-747421652), the test is flaky.
Change test to watch for the absence of warning (failure) events instead.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/test/check/events.go
+++ b/test/check/events.go
@@ -209,3 +209,16 @@ func (w *ObjectEventWatcher) WaitNotFor(stopChan chan struct{}, eventType EventT
 	}, fmt.Sprintf("not happen event type %s, reason = %s", string(eventType), reflect.ValueOf(reason).String()))
 	return
 }
+
+func (w *ObjectEventWatcher) WaitNotForType(stopChan chan struct{}, eventType EventType) (e *corev1.Event) {
+	w.dontFailOnMissingEvent = true
+	w.Watch(stopChan, func(event *corev1.Event) bool {
+		if event.Type == string(eventType) {
+			e = event
+			Fail(fmt.Sprintf("Did not expect %s. reason is %s", string(eventType), event.Reason), 1)
+			return true
+		}
+		return false
+	}, fmt.Sprintf("Event type %s did not occur", string(eventType)))
+	return
+}
--- a/test/check/check.go
+++ b/test/check/check.go
@@ -609,3 +609,13 @@ func CheckFailedEvent(gvk schema.GroupVersionKind, reason string) {
 	defer close(stopChan)
 	objectEventWatcher.WaitFor(stopChan, WarningEvent, fmt.Sprintf("%s: %s", eventemitter.FailedReason, reason))
 }
+
+func CheckNoWarningEvents(gvk schema.GroupVersionKind, rv string) {
+	By("Check absence of Warning events")
+	config := GetConfig(gvk)
+	configV1 := ConvertToConfigV1(config)
+	objectEventWatcher := NewObjectEventWatcher(configV1).SinceResourceVersion(rv).Timeout(time.Minute)
+	stopChan := make(chan struct{})
+	defer close(stopChan)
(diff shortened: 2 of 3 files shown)


RamLavi:
/hold

will want to re-run a few times to check if there's any issues

RamLavi:
/test pull-e2e-cluster-network-addons-operator-workflow-k8s

#1

alonSadan:
The code looks good to me:)
What I am worried about is that "no warnings"  could be weaker condition than "available" .
Could it be that a component will not be deployed, and no warnings will apear?
For example when it takes more time then usual to deploy? in that case we may want to increase timeouts on tests to "allow"
warnings to show up.
there could be other edgy scenarios. What do you think?

RamLavi:
> The code looks good to me:)
> What I am worried about is that "no warnings" could be weaker condition than "available" .

I agree that this is a weaker check, but unfortunately the alternative of watching the Available event is not working due to the issue I opened on k8s (see description).

> Could it be that a component will not be deployed, and no warnings will appear?
> For example when it takes more time then usual to deploy? in that case we may want to increase timeouts on tests to "allow"
> warnings to show up.
> there could be other edgy scenarios. What do you think?

I can add a check to make sure that the components are as expected, that should make the test a bit stronger. 
Do you have more suggestions?

test/check/events.go
```diff
@@ -209,3 +209,16 @@ func (w *ObjectEventWatcher) WaitNotFor(stopChan chan struct{}, eventType EventT
 	}, fmt.Sprintf("not happen event type %s, reason = %s", string(eventType), reflect.ValueOf(reason).String()))
 	return
 }
+
+func (w *ObjectEventWatcher) WaitNotForType(stopChan chan struct{}, eventType EventType) (e *corev1.Event) {
+	w.dontFailOnMissingEvent = true
+	w.Watch(stopChan, func(event *corev1.Event) bool {
+		if event.Type == string(eventType) {
+			e = event
+			Fail(fmt.Sprintf("Did not expect %s. reason is %s", string(eventType), event.Reason), 1)
```
qinqon, line 218: Do we need `defer GinkgoRecover` inside the func ? since this run at a coroutine.

qinqon (changes requested):

test/check/events.go
RamLavi: I think we already have `defer GinkgoRecover` inside the [Watch() function](https://github.com/kubevirt/cluster-network-addons-operator/blob/master/test/check/events.go#L163) where the goroutine is actually running

qinqon: Then we good.

alonSadan:
> > The code looks good to me:)
> > What I am worried about is that "no warnings" could be weaker condition than "available" .
> 
> I agree that this is a weaker check, but unfortunately the alternative of watching the Available event is not working due to the issue I opened on k8s (see description), and option of polling the status.Conditions has also proven problematic.
> 
> > Could it be that a component will not be deployed, and no warnings will appear?
> > For example when it takes more time then usual to deploy? in that case we may want to increase timeouts on tests to "allow"
> > warnings to show up.
> > there could be other edgy scenarios. What do you think?
> 
> [Added](https://github.com/kubevirt/cluster-network-addons-operator/compare/f50205873692daad7008853d4c11a7e869e5bbd3..1ccea55da0896a459d471629ad775795a6560d51) a check to make sure that the components are as expected, that should make the test a bit stronger.
> Do you have more suggestions?

nothing that I can think of.
Thanks.
/lgtm

RamLavi:
/hold cancel
I think we can safely say this flaky is resolved.

qinqon:
/approve

It has being run several times so this is safe now.

PR #730, opened 2021-01-07
unit-test, Remove copy to artifact in script

RamLavi:
We don't copy anything to artifacts folder during the unit test so
better remove it.
side note: should also fix the issue preventing prow ci [PR](https://github.com/kubevirt/project-infra/pull/811) to pass rehearsals

**Release note**:

```release-note
NONE
```

### DIFF
--- a/automation/check-patch.unit-test.sh
+++ b/automation/check-patch.unit-test.sh
@@ -4,14 +4,9 @@
 # cluster on any environment with basic dependencies listed in
 # check-patch.packages installed and docker running.
 
-teardown() {
-    cp $(find . -name "*junit*.xml") $ARTIFACTS
-}
-
 main() {
     source automation/check-patch.setup.sh
     cd ${TMP_PROJECT_PATH}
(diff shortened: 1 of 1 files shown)


phoracek (commented):
We still want to copy the junit xml output, no? Otherwise, we won't get info per test on the Prow view.

RamLavi:
> We still want to copy the junit xml output, no? Otherwise, we won't get info per test on the Prow view.

if there is info to copy then I agree, but I didn't see anywhere in the flow where such info is copied
Maybe we should keep it for future use, what do you say?

phoracek:
Indeed, can't find any collectors in our _test.go.

/lgtm
/approve

RamLavi:
/override pull-e2e-cluster-network-addons-operator-kubemacpool-functests

fails are known flakes and are not relevant to this PR

RamLavi:
- ran the unit-test manually locally to make sure this PR is safe
- will re-cherry-pick after #731 is merged on the stable branch

PR #795, opened 2021-03-08
certs, Add cert defaults to kubemacpool's bump script and manifest

RamLavi:
The PR adds cert default to kubemacpool's bump script and manifest, to prevent upgrade issues

- **lifecycle test, Add recent releases**
currently during non-release PR we only check old release
Until we insert something more synamic, manually removing
some of the old releases and introducing new ones.

- **certs, add cert default to kubemacpool's bump script and manifest**
add cert defaults in kubemacpool manifest so that during upgrade
the parameters are not reset.

The first commit **should fail** standalone until we introduce the cert defaults change in the second commit.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/automation/check-patch.e2e-lifecycle-k8s.sh
+++ b/automation/check-patch.e2e-lifecycle-k8s.sh
@@ -30,7 +30,7 @@ main() {
         export E2E_TEST_TIMEOUT=4h
     else
         # Don't run all upgrade tests in regular PRs, stick to those released under HCO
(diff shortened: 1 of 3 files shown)


RamLavi:
first commit [failed](https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_cluster-network-addons-operator/795/pull-e2e-cluster-network-addons-operator-lifecycle-k8s/1368840149152043008) as expected
With adding the second commit we should see that the problem is resolved.

PR #803, opened 2021-03-18
Remove Secret leftover checks at lifecycle test

qinqon:
Depending on the CNAO version the Secret has different annotations and
sometimes they are removed and sometimes not, this change just remove
the check since Secret is something managed by kube-admission-webhook
and not CNAO.

**Release note**:

```release-note
NONE
```


RamLavi (approved):
/lgtm
/approve

PR #804, opened 2021-03-19
Release 0.49.0

phoracek:
Signed-off-by: Petr Horáček <phoracek@redhat.com>

Features:
* bump nmstate to v0.43.0 (#800)

Bugs:
* Remove Secret leftover checks at lifecycle test (#803)

This also drops some older versions from lifecycle tests, leaving only the 3 latest releases and those released under HCO.

**Release note**:

```release-note
NONE
```


RamLavi:
@qinqon should this flaky be already fixed here? 
```
NodeNetworkConfigurationPolicy bonding default interface when there is a default interface with dynamic address should successfully move default IP address on top of the bond
```

qinqon:
> @qinqon should this flaky be already fixed here? 
> 
> ```
> 
> NodeNetworkConfigurationPolicy bonding default interface when there is a default interface with dynamic address should successfully move default IP address on top of the bond
> 
> ```

I think so, I haven't see it lately there

PR #805, opened 2021-03-21
gitActions, Restrict bumper workflows only to non-fork repo

RamLavi:
The bumper script workflow is not meant to run on forks.
To make sure this, we add a build condition to the current workflows.

**Release note**:
```release-note
NONE
```

### DIFF
--- a/.github/workflows/component-bumper-release-0.42.yml
+++ b/.github/workflows/component-bumper-release-0.42.yml
@@ -8,6 +8,7 @@ jobs:
 
   build:
     name: release-0.42 - CNAO Component Bump Job
+    if: (github.repository == 'kubevirt/cluster-network-addons-operator')
     runs-on: ubuntu-latest
     steps:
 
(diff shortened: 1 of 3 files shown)


RamLavi:
/hold
pending answer from github support that this is the best way to restrict workflows in forks

oshoval (approved):
its seems a natural solution
i dont think we should wait for support answer,
in the worst case we can change afterwards,

.github/workflows/component-bumper-release-0.42.yml
```diff
@@ -8,6 +8,7 @@ jobs:
 
   build:
     name: release-0.42 - CNAO Component Bump Job
+    if: (github.repository == 'kubevirt/cluster-network-addons-operator')
```
qinqon, line 11: Why do we have bump jobs at release branches?

qinqon (changes requested):

.github/workflows/component-bumper-release-0.42.yml
RamLavi: these are stable branches. 
The bumper also bumps stable branches, should a component gets updated. 
A good example is knmstate that got a new release on its stable branch, causing an auto bump on release-0.44 stable branch on CNAO..

qinqon (approved):

PR #808, opened 2021-03-25
Allow self sign config modification

qinqon:
In case of HCO upgrade, the default self sign configuartion values are
different from the ones at CNAO also HCO is suppose to allow
modification of cert rotation knobs, this change remove the check since
modification on those parameter just restart the knmstate webhooks and
kubemacpool and take the new values.

**Release note**:

```release-note
Alow self sign configuration modification
```


phoracek (commented):
Is there any simple way to test this e2e though? Overall, it looks fine, we did a similar thing with the placement API. I would sleep better if we had a test for it though, to prevent regressions in case our certificate mechanism changes.

qinqon:
> Is there any simple way to test this e2e though? Overall, it looks fine, we did a similar thing with the placement API. I would sleep better if we had a test for it though, to prevent regressions in case our certificate mechanism changes.

Added a e2e tests that check defaults and then modify them, similar to placement.

phoracek (approved):
/lgtm
/approve

PR #811, opened 2021-04-07
Use HCO cert rotation default values

qinqon:
CNAO uses different values than the ones used at HCO. This change align
both so we detect issues with that sooner.

**Release note**:

```release-note
Use rotate cert default configuration from HCO
```


phoracek (approved):
/lgtm
/approve

PR #812, opened 2021-04-07
Release 0.51.0

qinqon:
Test HCO default values at a release PR

**Release note**:

```release-note
NONE
```


qinqon:
/hold

Done to verify full upgrade path with HCO cert config default values.

PR #813, opened 2021-04-08
Use default at bad old self signed conf

qinqon:

If previous selfSignConfiguration is missing a field like in the case of
certOverlapInterval (it was not introduced at the beginning) then the
default values are not going to be used and the defaults from the
manifests apply so we can end up with a bad overall configuration. This
change use default values if previous ones were missing one of the
fields.

A follow up PR will be done testing that after upgrade certs are not rotated.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/network/self_sign_configuration.go
+++ b/pkg/network/self_sign_configuration.go
@@ -99,13 +99,21 @@ func DefaultSelfSignConfiguration() *cnao.SelfSignConfiguration {
 	}
 }
 func fillDefaultsSelfSignConfiguration(conf, previous *cnao.NetworkAddonsConfigSpec) []error {
-	if conf.SelfSignConfiguration == nil || conf.SelfSignConfiguration.CARotateInterval == "" || conf.SelfSignConfiguration.CAOverlapInterval == "" || conf.SelfSignConfiguration.CertRotateInterval == "" || conf.SelfSignConfiguration.CertOverlapInterval == "" {
-		if previous != nil && previous.SelfSignConfiguration != nil {
-			conf.SelfSignConfiguration = previous.SelfSignConfiguration
-			return []error{}
-		}
+	isMissingSelfSignConfiguration := func(confToCheck *cnao.NetworkAddonsConfigSpec) bool {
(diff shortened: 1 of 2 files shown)


pkg/network/self_sign_configuration.go
```diff
@@ -99,13 +99,22 @@ func DefaultSelfSignConfiguration() *cnao.SelfSignConfiguration {
 	}
 }
 func fillDefaultsSelfSignConfiguration(conf, previous *cnao.NetworkAddonsConfigSpec) []error {
-	if conf.SelfSignConfiguration == nil || conf.SelfSignConfiguration.CARotateInterval == "" || conf.SelfSignConfiguration.CAOverlapInterval == "" || conf.SelfSignConfiguration.CertRotateInterval == "" || conf.SelfSignConfiguration.CertOverlapInterval == "" {
-		if previous != nil && previous.SelfSignConfiguration != nil {
+
```
RamLavi, line 102: can you remove the extra line?

RamLavi (changes requested):
small nit

cluster/operator-push.sh
```diff
@@ -38,9 +38,9 @@ rm -rf _out/
 cp -r manifests/ _out/
 IMAGE_REGISTRY=$manifest_registry DEPLOY_DIR=_out make gen-manifests
 
-make cluster-clean
+#make cluster-clean
```
phoracek, line 41: Either delete it or keep uncommented

pkg/network/network.go
```diff
@@ -118,6 +118,10 @@ func Render(conf *cnao.NetworkAddonsConfigSpec, manifestDir string, openshiftNet
 	if err != nil {
 		return nil, err
 	}
+
+	for _, obj := range o {
+		log.Printf("kmp obj: %+v", *obj)
```
phoracek, line 123: This does not look like a proper log we'd like to have in our stderr. Either name it appropriately or delete it. Note that this may spam the log a lot.

phoracek (changes requested):

PR #815, opened 2021-04-08
Migrate NMState to use kubernetes-nmstate-operator if available

bcrochet:
This patch make CNAO use the kubernetes-nmstate-operator if it is
running and available. The only prereq for this is to have the
operator pod running. If that's detected via the
app=kubernetes-nmstate-operator label on a Deployment, then CNAO
will only render an nmstates.nmstate.io CR to kick off the handler
and webhook from the operator.

If the operator is added, it also has upgrade logic to move to using
the operator instead of installing kubernetes-nmstate directly. It
will remove the old instance.

Signed-off-by: Brad P. Crochet <brad@redhat.com>

See above.

See https://nmstate.io/kubernetes-nmstate/deployment.html for installing the operator

Depends on https://github.com/kubevirt/cluster-network-addons-operator/pull/828

**Release note**:

```release-note
Migrate Kubernetes NMState install to use kubernetes-nmstate-operator if it is already available in the cluster. This will work on upgrade as well.
```


qinqon:
@bcrochet Let's fix CI before review this

pkg/network/nmstate.go
```diff
@@ -74,13 +104,33 @@ func removeDaemonSetHandlerWorker(ctx context.Context, client k8sclient.Client)
...
 	if conf.NMState == nil {
 		return []error{}
 	}
 
 	errList := []error{}
 	errList = append(errList, removeDaemonSetHandlerWorker(ctx, client)...)
+	if clusterInfo.NmstateOperator {
+		errList = append(errList, removeStandaloneHandler(ctx, client)...)
+		errList = append(errList, removeStandaloneWebhook(ctx, client)...)
+	}
```
qinqon, line 133: We have bump CNAO to latest nmstate and now there is a new Deployment "cert-manager" it has to be added here.

qinqon (changes requested):
firt pass, let's add cert-manager too.

pkg/network/nmstate.go
bcrochet: Done.

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
```diff
@@ -529,6 +538,22 @@ func isSCCAvailable(c kubernetes.Interface) (bool, error) {
 	return isResourceAvailable(c, "securitycontextconstraints", "security.openshift.io", "v1")
 }
 
+func isRunningKubernetesNMStateOperator(c kubernetes.Interface) (bool, error) {
```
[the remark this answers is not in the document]
bcrochet: > Do we have to check if the NMState CR CNAO has is compatible with the installed CRD from the operator ?
> 

I don't think so. Versioning of the NMState API should take care of that. And the nmstate operator should handle that.

> Imaging that we are installing CNAO on openshift with old NMState operator we will apply a possible too new NMState CR and the opposite CNAO being old and openshift NMState operator newer.

So, an OpenShift cluster already has NMState operator installed.. CNAO would try to install a newer CR? That's an interesting scenario, but again, I think versioning would come into play there. I suppose we could detect in CNAO that the version we are trying to install is what we are trying to install... As for he CNAO being old and nmstate operator being newer... that should be handled by the nmstate operator.

> 
> Also in case the CR has being already applied I would do nothing to reduce possible problems.

If the CR is already applied, it should be a big no-op. But it would be interesting to find out what happens there.

qinqon:
/hold

We have to bump nmstate at CNAO before merging this since it depends on the fix for handler lock

qinqon:
/hold cancel

Fix mertged and released

qinqon:
@bcrochet we hav already a handler lock test on kubernetes-nmstate so the function is working fine, can we do at least a manual test with this PR for CNAO to everify that it's working, it would be even better a e2e tests at CNAO but don't know if it's worth it.

bcrochet:
> @bcrochet we hav already a handler lock test on kubernetes-nmstate so the function is working fine, can we do at least a manual test with this PR for CNAO to everify that it's working, it would be even better a e2e tests at CNAO but don't know if it's worth it.

Quick test shows that it does indeed keep the operator-installed instance (outside of CNAO) from taking the lock.

qinqon:
@bcrochet it's worth adding some e2e test here ?

bcrochet:
> @bcrochet it's worth adding some e2e test here ?

A couple would be a good idea.

test/e2e/workflow/deployment_test.go
```diff
@@ -336,8 +337,71 @@ var _ = Describe("NetworkAddonsConfig", func() {
...
+	result, err := kubectl.Kubectl("apply", "-f", "test/manifests/kubernetes-nmstate-operator.yaml")
+	Expect(err).ToNot(HaveOccurred(), fmt.Sprintf("Error when running kubectl: %s", result))
+}
+
+func uninstallNMStateOperator() {
+	By("Uninstalling kubernetes-nmstate-operator")
+	result, err := kubectl.Kubectl("delete", "-f", "test/manifests/kubernetes-nmstate-operator.yaml")
+	Expect(err).ToNot(HaveOccurred(), fmt.Sprintf("Error when running kubectl: %s", result))
+}
+
```
qinqon, line 404: Can we apply the manifest from the kubernetes-nmstate version at components.yaml ?

qinqon (changes requested):
Let's take the manifests from components.yaml

test/manifests/kubernetes-nmstate-operator.yaml
```diff
@@ -0,0 +1,263 @@
+---
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
```
qinqon, line 3: we don't need this anymore right ?

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
```diff
@@ -529,6 +539,21 @@ func isSCCAvailable(c kubernetes.Interface) (bool, error) {
 	return isResourceAvailable(c, "securitycontextconstraints", "security.openshift.io", "v1")
 }
 
+func isRunningKubernetesNMStateOperator(c k8sclient.Client) (bool, error) {
+	deployments := &appsv1.DeploymentList{}
+	err := c.List(context.TODO(), deployments, k8sclient.MatchingLabels{"app": "kubernetes-nmstate-operator"})
```
qinqon, line 544: Don't we have to specify the namespace ?

bcrochet: We wouldn't necessarily know what the namespace is.

PR #846, opened 2021-05-20
Add relationship labels to all directly deployed components

oshoval:
Each component deployed by CNAO should have labels on its resources.

The labels are called relationship labels (app.kubernetes.io/*),
and are used in order to create a relationship graph in the UI.

The labels are:
`app.kubernetes.io/component`
`app.kubernetes.io/managed-by`
`app.kubernetes.io/version`
`app.kubernetes.io/part-of`

This PR adapts Render function to add those labels to all the directly deployed components.
Follow PRs on each component repository will handle its own internal deployed components.

```release-note
None
```

### DIFF
--- a/pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
+++ b/pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
@@ -340,14 +340,11 @@ func (r *ReconcileNetworkAddonsConfig) renderObjectsV1(networkAddonsConfig *cnao
 	}
 	objs = append([]*unstructured.Unstructured{applied}, objs...)
 
-	// Label objects with version of the operator they were created by
-	for _, obj := range objs {
-		labels := obj.GetLabels()
-		if labels == nil {
-			labels = map[string]string{}
-		}
-		labels[cnaov1.SchemeGroupVersion.Group+"/version"] = operatorVersionLabel
-		obj.SetLabels(labels)
+	err = updateObjectsLabels(networkAddonsConfig.GetLabels(), objs)
+	if err != nil {
+		log.Printf("failed to update objects labels: %v", err)
+		err = errors.Wrapf(err, "failed to update objects labels")
+		return objs, err
 	}
 
 	return objs, nil
@@ -496,6 +493,63 @@ func (r *ReconcileNetworkAddonsConfig) stopTrackingObjects() {
 
 	// Trigger status manager to notice the change
 	r.statusManager.SetFromPods()
+}
+
+func updateObjectsLabels(crLables map[string]string, objs []*unstructured.Unstructured) error {
+	var err error
+	for _, obj := range objs {
+		labels := obj.GetLabels()
+		if labels == nil {
+			labels = map[string]string{}
+		}
+		// Label objects with version of the operator they were created by
+		labels[cnaov1.SchemeGroupVersion.Group+"/version"] = operatorVersionLabel
+
+		appLabelKeys := []string{names.COMPONENT_LABEL_KEY, names.PART_OF_LABEL_KEY, names.VERSION_LABEL_KEY, names.MANAGED_BY_LABEL_KEY}
+
+		labels = updateLabelsFromCR(labels, crLables, appLabelKeys)
+		if err != nil {
+			return err
+		}
+		obj.SetLabels(labels)
+
+		err = updateObjectTemplateLabels(obj, labels, appLabelKeys)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func updateLabelsFromCR(labels, crLables map[string]string, appLabelKeys []string) map[string]string {
+	labels[names.COMPONENT_LABEL_KEY] = names.COMPONENT_LABEL_DEFAULT_VALUE
+	labels[names.MANAGED_BY_LABEL_KEY] = names.MANAGED_BY_LABEL_DEFAULT_VALUE
+	for _, key := range appLabelKeys {
+		if value, exist := crLables[key]; exist == true {
+			labels[key] = value
+		}
+	}
+
+	return labels
+}
+
+func updateObjectTemplateLabels(obj *unstructured.Unstructured, labels map[string]string, appLabelKeys []string) error {
+	kind := obj.GetKind()
+	if kind == "DaemonSet" || kind == "ReplicaSet" || kind == "Deployment" || kind == "StatefulSet" {
+		for _, key := range appLabelKeys {
+			if value, exist := labels[key]; exist == true {
+				err := unstructured.SetNestedField(obj.Object, value, "spec", "template", "metadata", "labels", key)
+				if err != nil {
+					log.Printf("failed to add relationship label %s: %v", key, err)
+					err = errors.Wrapf(err, "failed to add relationship labels")
+					return err
+				}
+			}
+		}
+	}
+
+	return nil
 }
 
 func collectContainersInfo(parentKind string, parentName string, containers []v1.Container) []cnao.Container {
--- a/pkg/controller/networkaddonsconfig/networkaddonsconfig_controller_test.go
+++ b/pkg/controller/networkaddonsconfig/networkaddonsconfig_controller_test.go
@@ -1,0 +1,163 @@
+package networkaddonsconfig
+
+import (
+	"fmt"
+
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	"github.com/kubevirt/cluster-network-addons-operator/pkg/names"
+	"github.com/kubevirt/cluster-network-addons-operator/pkg/render"
+)
+
+type checkUnit struct {
+	key           string
+	shouldExist   bool
+	expectedValue string
+}
+
+var _ = Describe("Networkaddonsconfig", func() {
+	const testDataFile = "testdata/data.yaml"
+
+	Context("When CR is not labeled", func() {
+		var objs []*unstructured.Unstructured
+
+		BeforeEach(func() {
+			var err error
+			renderData := render.MakeRenderData()
+			objs, err = render.RenderTemplate(testDataFile, &renderData)
+			Expect(err).NotTo(HaveOccurred())
+
+			crLabels := map[string]string{}
+			err = updateObjectsLabels(crLabels, objs)
+			Expect(err).NotTo(HaveOccurred())
+		})
+
+		It("Should find only default relationship labels", func() {
+			appLabelKeys := []checkUnit{
+				{
+					key:           names.COMPONENT_LABEL_KEY,
+					shouldExist:   true,
+					expectedValue: names.COMPONENT_LABEL_DEFAULT_VALUE,
+				},
+				{
+					key:           names.MANAGED_BY_LABEL_KEY,
+					shouldExist:   true,
+					expectedValue: names.MANAGED_BY_LABEL_DEFAULT_VALUE,
+				},
+				{
+					key:           names.PART_OF_LABEL_KEY,
+					shouldExist:   false,
+					expectedValue: "Invalid",
+				},
+				{
+					key:           names.VERSION_LABEL_KEY,
+					shouldExist:   false,
+					expectedValue: "Invalid",
+				},
+			}
+
+			checkObjectsRelationshipLabels(objs, appLabelKeys)
+		})
+	})
+	Context("When CR is labeled", func() {
+		var objs []*unstructured.Unstructured
+		var crLabels map[string]string
+		const expectedComponentLabel = "component_unit_tests"
+		const expectedManagedByLabel = "managed_by_unit_tests"
+		const expectedPartOfLabel = "part_of_unit_tests"
+		const expectedVersionLabel = "version_of_unit_tests"
+
+		BeforeEach(func() {
+			var err error
+			renderData := render.MakeRenderData()
+			objs, err = render.RenderTemplate(testDataFile, &renderData)
+			Expect(err).NotTo(HaveOccurred())
+
+			crLabels = map[string]string{}
+			crLabels[names.COMPONENT_LABEL_KEY] = expectedComponentLabel
+			crLabels[names.MANAGED_BY_LABEL_KEY] = expectedManagedByLabel
+			crLabels[names.PART_OF_LABEL_KEY] = expectedPartOfLabel
+			crLabels[names.VERSION_LABEL_KEY] = expectedVersionLabel
+			err = updateObjectsLabels(crLabels, objs)
+			Expect(err).NotTo(HaveOccurred())
+		})
+
+		It("Should find all labels, overridden by CR labels", func() {
+			appLabelKeys := []checkUnit{
+				{
+					key:           names.COMPONENT_LABEL_KEY,
+					shouldExist:   true,
+					expectedValue: expectedComponentLabel,
+				},
+				{
+					key:           names.MANAGED_BY_LABEL_KEY,
+					shouldExist:   true,
+					expectedValue: expectedManagedByLabel,
+				},
+				{
+					key:           names.PART_OF_LABEL_KEY,
+					shouldExist:   true,
+					expectedValue: expectedPartOfLabel,
+				},
+				{
+					key:           names.VERSION_LABEL_KEY,
+					shouldExist:   true,
+					expectedValue: expectedVersionLabel,
+				},
+			}
+
+			checkObjectsRelationshipLabels(objs, appLabelKeys)
+		})
+	})
+})
+
+func checkObjectsRelationshipLabels(objs []*unstructured.Unstructured, appLabelKeys []checkUnit) {
+	appLabelsNotExists := []checkUnit{
+		{
+			key:           names.COMPONENT_LABEL_KEY,
+			shouldExist:   false,
+			expectedValue: "Invalid",
+		},
+		{
+			key:           names.MANAGED_BY_LABEL_KEY,
+			shouldExist:   false,
+			expectedValue: "Invalid",
+		},
+		{
+			key:           names.PART_OF_LABEL_KEY,
+			shouldExist:   false,
+			expectedValue: "Invalid",
+		},
+		{
+			key:           names.VERSION_LABEL_KEY,
+			shouldExist:   false,
+			expectedValue: "Invalid",
+		},
+	}
+
+	for _, obj := range objs {
+		labels := obj.GetLabels()
+		checkLabels(labels, appLabelKeys)
+
+		kind := obj.GetKind()
+		templateLabels, _, _ := unstructured.NestedStringMap(obj.Object, "spec", "template", "metadata", "labels")
+		if kind == "DaemonSet" || kind == "ReplicaSet" || kind == "Deployment" || kind == "StatefulSet" {
+			checkLabels(templateLabels, appLabelKeys)
+		} else {
+			checkLabels(templateLabels, appLabelsNotExists)
+		}
+	}
+}
+
+func checkLabels(labels map[string]string, units []checkUnit) {
+	for _, unit := range units {
+		value, found := labels[unit.key]
+		Expect(found).To(Equal(unit.shouldExist), fmt.Sprintf("%s exist test", unit.key))
+		if unit.shouldExist {
+			Expect(value).To(Equal(unit.expectedValue), fmt.Sprintf("%s value test", unit.key))
+		}
+	}
+}
--- a/pkg/controller/networkaddonsconfig/testdata/data.yaml
+++ b/pkg/controller/networkaddonsconfig/testdata/data.yaml
@@ -1,0 +1,70 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  labels:
+    control-plane: mac-controller-manager
+    controller-tools.k8s.io: "1.0"
+  name: kubemacpool-mac-controller-manager
+  namespace: test-ns
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      control-plane: mac-controller-manager
+      controller-tools.k8s.io: "1.0"
+  strategy:
+    type: Recreate
+  template:
+    metadata:
+      annotations:
+        description: KubeMacPool manages MAC allocation to Pods and VMs
+      labels:
+        app: kubemacpool
+        control-plane: mac-controller-manager
+        controller-tools.k8s.io: "1.0"
+    spec:
+        image: dummy
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  labels:
+    control-plane: mac-controller-manager
+    controller-tools.k8s.io: "1.0"
+  name: kubemacpool-mac-range-config
+  namespace: cluster-network-addons
+data:
+  RANGE_START: 00:00:00:00:00:04
+  RANGE_END: 00:00:00:00:00:05
+---
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  labels:
+    component: kubernetes-nmstate-handler
+    networkaddonsoperator.network.kubevirt.io/version: 99.0.0
+  name: nmstate-handler
+  namespace: cluster-network-addons
+spec:
+  selector:
+    matchLabels:
+      name: nmstate-handler
+  template:
+    metadata:
+      labels:
+        app: kubernetes-nmstate
+        component: kubernetes-nmstate-handler
+        name: nmstate-handler
+    spec:
+        image: dummy
+        imagePullPolicy: IfNotPresent
+        name: nmstate-handler
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  annotations:
+    kubernetes.io/service-account.name: cluster-network-addons-operator
+  name: cluster-network-addons-operator-token-test
+  namespace: cluster-network-addons
+type: kubernetes.io/service-account-token
--- a/test/check/check.go
+++ b/test/check/check.go
@@ -176,10 +176,10 @@ func CheckOperatorIsReady(timeout time.Duration) {
 	By("Checking that the operator is up and running")
 	if timeout != CheckImmediately {
 		Eventually(func() error {
-			return checkForDeployment(components.Name)
+			return checkForDeployment(components.Name, false)
 		}, timeout, time.Second).ShouldNot(HaveOccurred(), fmt.Sprintf("Timed out waiting for the operator to become ready"))
 	} else {
-		Expect(checkForDeployment(components.Name)).ShouldNot(HaveOccurred(), "Operator is not ready")
+		Expect(checkForDeployment(components.Name, false)).ShouldNot(HaveOccurred(), "Operator is not ready")
 	}
 }
 
@@ -187,10 +187,10 @@ func CheckNMStateOperatorIsReady(timeout time.Duration) {
 	By("Checking that the operator is up and running")
 	if timeout != CheckImmediately {
 		Eventually(func() error {
-			return checkForGenericDeployment("nmstate-operator", "nmstate", false)
+			return checkForGenericDeployment("nmstate-operator", "nmstate", false, false)
 		}, timeout, time.Second).ShouldNot(HaveOccurred(), fmt.Sprintf("Timed out waiting for the operator to become ready"))
 	} else {
-		Expect(checkForGenericDeployment("nmstate-operator", "nmstate", false)).ShouldNot(HaveOccurred(), "Operator is not ready")
+		Expect(checkForGenericDeployment("nmstate-operator", "nmstate", false, false)).ShouldNot(HaveOccurred(), "Operator is not ready")
 	}
 }
 
@@ -296,7 +296,7 @@ func checkForComponent(component *Component) error {
 	}
 
 	for _, deployment := range component.Deployments {
-		errsAppend(checkForDeployment(deployment))
+		errsAppend(checkForDeployment(deployment, true))
 	}
 
 	if component.Secret != "" {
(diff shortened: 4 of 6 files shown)


oshoval:
@RamLavi 
Hi, I need please ok to test if possible

oshoval:
Working on adding tests

oshoval:
https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_cluster-network-addons-operator/846/pull-e2e-cluster-network-addons-operator-lifecycle-k8s/1397140339046748160
happens also on 
https://prow.ci.kubevirt.io/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_cluster-network-addons-operator/829/pull-e2e-cluster-network-addons-operator-lifecycle-k8s/1387725826513965056

both have 
```
 Warning  FailedCreate      14m (x3 over 14m)  daemonset-controller  Error creating: pods "bridge-marker-" is forbidden: error looking up service account cluster-network-addons/bridge-marker: serviceaccount "bridge-marker" not found
```

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
```diff
@@ -498,6 +495,97 @@ func (r *ReconcileNetworkAddonsConfig) stopTrackingObjects() {
...
+
+		obj.SetLabels(labels)
+	}
+
+	return objs, nil
+}
+
+func updateLabels(obj *unstructured.Unstructured, labels, crLables map[string]string) (map[string]string, error) {
+	var err error
+	const partOfLabelKey = "app.kubernetes.io/part-of"
```
RamLavi, line 526: perhaps this should be declared in a more general place? it's would make it easier to be used by unit tests later on.

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
```diff
@@ -498,6 +495,97 @@ func (r *ReconcileNetworkAddonsConfig) stopTrackingObjects() {
...
+	const versionLabelKey = "app.kubernetes.io/version"
+	for key, label := range crLables {
```
RamLavi, line 528: I think it's better to not iterate over all the crLabels.
Since it's a `map[string]string`, maybe it's better to do something like
```
appLabelKeys := []string{partOfLabelKey, versionLabelKey}
	for _, key := range appLabelKeys {
		if value, exist := crLables[key]; exist == true {
			labels, err = updateLabelByKey(obj, key, value, labels)
			if err != nil {
				return labels, err
			}
		}
	}
```

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
```diff
@@ -498,6 +495,97 @@ func (r *ReconcileNetworkAddonsConfig) stopTrackingObjects() {
...
+		return labels, err
+	}
+
+	return labels, nil
+}
+
+func updateManagedByLabel(obj *unstructured.Unstructured, labels, crLables map[string]string) (map[string]string, error) {
+	var err error
+
+	const managedbyLabelKey = "app.kubernetes.io/managed-by"
```
RamLavi, line 551: ditto. should be set in a more centralized place

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
```diff
@@ -498,6 +495,97 @@ func (r *ReconcileNetworkAddonsConfig) stopTrackingObjects() {
...
+	var err error
+	for _, obj := range objs {
+		labels := obj.GetLabels()
+		if labels == nil {
+			labels = map[string]string{}
+		}
+		// Label objects with version of the operator they were created by
+		labels[cnaov1.SchemeGroupVersion.Group+"/version"] = operatorVersionLabel
+
+		labels, err = updateLabels(obj, labels, crLables)
```
RamLavi, line 508: it seems that `updateLabels/updateAppLabelsFromCR` is not only updating the `labels`, but also changing the `obj.spec.template...` via `updateTemplateLabels`. that seems kind of off to me, since the function is not doing exactly what it is declaring to do.

RamLavi (changes requested):
added some comments.

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
oshoval: So to update to function name to `updateObjLabels` and return obj instead labels ?

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
oshoval: btw unit tests does not check reconcile
the one that do check is e2e  test/check/check.go
```

checkForGenericDeployment
checkForComponent
checkConfigChange
```
Those check for example the version label that was already exists, in the same code that the PR updated

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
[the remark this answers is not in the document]
oshoval: merge with this discussion i think
https://github.com/kubevirt/cluster-network-addons-operator/pull/846#discussion_r639507663

oshoval: updated to updateObjLabels

oshoval:
Updated according resolved comments, answered some, thanks

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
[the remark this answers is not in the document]
RamLavi: IIUC `labels[cnaov1.SchemeGroupVersion.Group+"/version"] = operatorVersionLabel` is not done inside `updateLabels`.

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
RamLavi: is there a problem to create a `networkaddonsconfig_controller_test.go` and add the unit tests for updateObjectsLabels(), updateObjLabels(), updateLabelByKey(), updateManagedByLabel()?

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
RamLavi: well we can put it in names/names.go

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
oshoval: I would suggest to start with the e2e tests, the infrastructure for them is ready
on following PRs we can add the unit tests, as it seems there isn't even infra for checking the controller in unit test
is it ok ?

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
```diff
@@ -498,6 +495,97 @@ func (r *ReconcileNetworkAddonsConfig) stopTrackingObjects() {
...
+	var err error
+	for _, obj := range objs {
+		labels := obj.GetLabels()
+		if labels == nil {
+			labels = map[string]string{}
+		}
+		// Label objects with version of the operator they were created by
+		labels[cnaov1.SchemeGroupVersion.Group+"/version"] = operatorVersionLabel
+
+		labels, err = updateObjLabels(obj, labels, crLables)
```
RamLavi, line 508: I don't think it's a good name.
the whole function is called `updateObjectsLabels`, so calling the sub-function `updateObjLabels` is not specific enough..
IIUC What this function specifically does is update the app-labels that are originated from the CNAO CR, if exists.. so the function name should be specific enough to say just that.

RamLavi (changes requested):

oshoval:
Addressed comments, and the refactor as we spoke
leftovers - tests

oshoval:
Fixed typo
Added more consts

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
```diff
@@ -498,6 +495,65 @@ func (r *ReconcileNetworkAddonsConfig) stopTrackingObjects() {
...
+		if err != nil {
+			return err
+		}
+
+		err = updateTemplateLabels(obj, labels)
+		if err != nil {
+			return err
+		}
+
+		obj.SetLabels(labels)
```
RamLavi, line 521: it may be more coherent to move this before the `err = updateTemplateLabels(obj, labels)`  line

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
```diff
@@ -498,6 +495,65 @@ func (r *ReconcileNetworkAddonsConfig) stopTrackingObjects() {
...
+	var err error
+	for _, obj := range objs {
+		labels := obj.GetLabels()
+		if labels == nil {
+			labels = map[string]string{}
+		}
+		// Label objects with version of the operator they were created by
+		labels[cnaov1.SchemeGroupVersion.Group+"/version"] = operatorVersionLabel
+
+		labels[names.COMPONENT_LABEL_KEY] = names.COMPONENT_LABEL_DEFAULT_VALUE
```
RamLavi, line 508: I think we should move these 2 lines that initialize the `managed-by` and `component` labels to be inside `updateLabelsFromCR`. It would make it more visual that we're setting a default value then rewriting it, and make it harder for future us to mistakenly reorder the commands without noticing we're changing the label behavior.

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
```diff
@@ -498,6 +495,65 @@ func (r *ReconcileNetworkAddonsConfig) stopTrackingObjects() {
...
+		labels[names.MANAGED_BY_LABEL_KEY] = names.MANAGED_BY_LABEL_DEFAULT_VALUE
+
+		labels = updateLabelsFromCR(labels, crLables)
+		if err != nil {
+			return err
+		}
+
+		err = updateTemplateLabels(obj, labels)
```
RamLavi, line 516: suggestion: `updateObjectTemplateLabels`?

RamLavi (changes requested):

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
oshoval: we don't override `component` so i will just remove the `managed-by`

RamLavi: Thinking about it, I think we should also override `component`, so both should be moved to `updateLabelsFromCR`

oshoval: isnt moving it to `updateLabelsFromCR`
breaks the behavior that we are setting a default value which is not from the CR ?

oshoval:
addressed comments, left the discussion open (even that addressed it already)

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
RamLavi: perhaps then `initializeAndUpdateLabelsFromCR`

oshoval: Prepare to not over complicate the function name

oshoval:
Added e2e (unit tests next)

oshoval:
1st push - rebase
2nd - attempt to fix lifecycle (lets see first it pass before review please)

oshoval:
Added preliminary unit tests, will extend upon needs

oshoval:
Fix lifecycle tests label check selection (so it will still check version but not relationship)

oshoval:
This flake is not due to the PR, it appeared before and not related
https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_cluster-network-addons-operator/846/pull-e2e-cluster-network-addons-operator-nmstate-functests/1401449324868538368

test/check/check.go
```diff
@@ -176,21 +176,21 @@ func CheckOperatorIsReady(timeout time.Duration) {
...
+		Expect(checkForDeployment(components.Name, false)).ShouldNot(HaveOccurred(), "Operator is not ready")
 	}
 }
 
 func CheckNMStateOperatorIsReady(timeout time.Duration) {
 	By("Checking that the operator is up and running")
 	if timeout != CheckImmediately {
 		Eventually(func() error {
-			return checkForGenericDeployment("nmstate-operator", "nmstate", false)
+			return checkForGenericDeployment("nmstate-operator", "nmstate", false, false)
```
RamLavi, line 190: Can you elaborate why you set it to false?
UUIC `CheckNMStateOperatorIsReady` is running under `e2e/workflow` tests, so they should only run on latest code.
shouldn't we pass `true` then?

test/check/check.go
```diff
@@ -468,11 +524,51 @@ func checkForDaemonSet(name string) error {
...
+
+	err = checkRelationshipLabels(mutatingWebhookConfig.GetLabels(), "MutatingWebhookConfiguration", name)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func checkRelationshipLabels(labels map[string]string, kind, name string) error {
```
RamLavi, line 556: can we unify `checkWorkloadRelationLabels` and `checkRelationshipLabels` by changing the input var `labels map[string]string` to be a list of `[]map[string]string`?
```suggestion
func checkRelationshipLabels(labelMapList []map[string]string, kind, name string) error {
```

You could then just iterate over the labels:
```
for labels := range labelMapList{
... // current checkRelationshipLabels() code
}
```

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller_test.go
```diff
@@ -0,0 +1,123 @@
+package networkaddonsconfig
+
+import (
+	"fmt"
+
+	"github.com/kubevirt/cluster-network-addons-operator/pkg/names"
```
RamLavi, line 6: please move `github.com/kubevirt/cluster-network-addons-operator` related imports to a separate import group

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller_test.go
```diff
@@ -0,0 +1,123 @@
...
+	"github.com/kubevirt/cluster-network-addons-operator/pkg/render"
+	. "github.com/onsi/ginkgo"
```
RamLavi, line 8: ditto for test related imports.

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller_test.go
```diff
@@ -0,0 +1,123 @@
...
+
+type checkUnit struct {
+	key           string
+	shouldExist   bool
+	expectedValue string
+}
+
+var _ = Describe("Networkaddonsconfig", func() {
+	Context("When CR is not labeled", func() {
+		It("Should find only default labels", func() {
```
RamLavi, line 21:
```suggestion
		It("Should find only default Relation labels", func() {
```

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller_test.go
```diff
@@ -0,0 +1,123 @@
...
+			renderData := render.MakeRenderData()
+			objs, err := render.RenderTemplate("testdata/data.yaml", &renderData)
```
RamLavi, line 23: please move `testdata/data.yaml` to const

RamLavi, line 23: could you rename `data.yaml` to be meaningful? suggestion ``deployment-without-relation-labels.yaml"

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller_test.go
```diff
@@ -0,0 +1,123 @@
...
+type checkUnit struct {
+	key           string
+	shouldExist   bool
+	expectedValue string
+}
+
+var _ = Describe("Networkaddonsconfig", func() {
+	Context("When CR is not labeled", func() {
+		It("Should find only default labels", func() {
+			renderData := render.MakeRenderData()
```
RamLavi, line 22: MakeRenderData+RenderTemplate should be in BeforeEach() IMO, as it is more related to `Context("When CR is not labeled"` and not to the `It("Should find only default labels"` part

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller_test.go
```diff
@@ -0,0 +1,123 @@
...
+			Expect(err).NotTo(HaveOccurred())
+
+			templateLabels, _, _ := unstructured.NestedStringMap(objs[0].Object, "spec", "template", "metadata", "labels")
+			err = checkLabels(templateLabels, appLabelKeys)
+			Expect(err).NotTo(HaveOccurred())
+		})
+	})
+	Context("When CR is labeled", func() {
+		It("Should find all labels, overrided by CR labels", func() {
+			renderData := render.MakeRenderData()
```
RamLavi, line 64: ditto

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller_test.go
```diff
@@ -0,0 +1,123 @@
...
+			err = checkLabels(labels, appLabelKeys)
+			Expect(err).NotTo(HaveOccurred())
+
+			templateLabels, _, _ := unstructured.NestedStringMap(objs[0].Object, "spec", "template", "metadata", "labels")
+			err = checkLabels(templateLabels, appLabelKeys)
+			Expect(err).NotTo(HaveOccurred())
+		})
+	})
+	Context("When CR is labeled", func() {
+		It("Should find all labels, overrided by CR labels", func() {
```
RamLavi, line 63: typo
```suggestion
		It("Should find all labels, overridden by CR labels", func() {
```

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller_test.go
```diff
@@ -0,0 +1,123 @@
...
+			renderData := render.MakeRenderData()
+			objs, err := render.RenderTemplate("testdata/data.yaml", &renderData)
+			Expect(err).NotTo(HaveOccurred())
+
+			const expectedManagedByLabel = "managed_by_unit_tests"
```
RamLavi, line 68: the whole part configuring the crLables should also be in `BeforeEach()`

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller_test.go
```diff
@@ -0,0 +1,123 @@
...
+			const expectedPartOfLabel = "part_of_unit_tests"
+			const expectedVersionLabel = "version_of_unit_tests"
+
+			crLables := map[string]string{}
+			crLables[names.MANAGED_BY_LABEL_KEY] = expectedManagedByLabel
```
RamLavi, line 73: we should be able to override the componet label. can you add that?

pkg/controller/networkaddonsconfig/testdata/data.yaml
```diff
@@ -0,0 +1,26 @@
+apiVersion: apps/v1
```
RamLavi, line 1: can we add a few more examples like daemonset, secret? then he test will just render all and check them all

RamLavi (changes requested):
first pass

RamLavi:
> This flake is not due to the PR, it appeared before and not related
> https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_cluster-network-addons-operator/846/pull-e2e-cluster-network-addons-operator-nmstate-functests/1401449324868538368

your right

test/check/check.go
```diff
@@ -468,11 +520,51 @@ func checkForDaemonSet(name string) error {
...
+	if labels == nil {
+		return nil
```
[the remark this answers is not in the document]
oshoval: Kept the same behavior as on master, there are cases where there is no labels at all,
see here when i tried to fix it
https://github.com/kubevirt/cluster-network-addons-operator/pull/859

i can try to understand better if it just on lifecycle and fix it, since checkRelationshipLabels
is not run on lifecycle

test/check/check.go
oshoval: It run installNMStateOperator which install nmstate standalone
so it doesnt need to have labels right ?

test/check/check.go
oshoval: I think it will be less readable,
i prefer the logic separation if possible, as one is the labels from the spec and one is the labels from the metadata
and this way its explicit instead of array of arrays

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller_test.go
oshoval: But its different per each `It`

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller_test.go
oshoval: overriden one of them (`expectedManagedByLabel`) and left the other intact, to show it knows also to keep a value if its not on crLabels
but i can if thats what you prefer

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller_test.go
oshoval: I think not , especially if we add data as this one suggests
https://github.com/kubevirt/cluster-network-addons-operator/pull/846#discussion_r646128959

oshoval:
Marked as resolved the changes, answered some, will continue the unanswered
thanks

test/check/check.go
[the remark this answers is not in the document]
RamLavi: I see. can you open an issue on this so we remember to further investigate?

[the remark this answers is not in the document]
oshoval: ok
note that In this PR i already fixed it for the relationships labels

test/check/check.go
RamLavi: I see what you mean, but imo the benefit would outcast the issues.
unifying it would make unit tests simpler, and would reduce redundant code.. 
at the end we are checking sets of labels of type `map[string]string`.

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller_test.go
RamLavi: then could be "objects-without-relation-labels.yaml" or something the like

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller_test.go
RamLavi: >to show it knows also to keep a value if its not on crLabels

we are already checking that in the other test though.
I think unit tests should document the desired functionality, sort of documentation of what we expect to happen.
so IMO we should also override the componet label as well.

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller_test.go
oshoval: The objects dont need to have relation labels, that data is raw, so i don't think we should emphasize it
its like testdata/template.yaml that already appears on other folders

i would prefer to keep it simple and raw if possible please

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller_test.go
oshoval: note - I already done that in the previous push, this is why its resolved

test/check/check.go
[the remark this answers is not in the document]
oshoval: https://github.com/kubevirt/cluster-network-addons-operator/issues/875

oshoval:
Addressed https://github.com/kubevirt/cluster-network-addons-operator/pull/846#discussion_r646128959
with few cosmetics

test/check/check.go
oshoval: as we spoke its not unit test, but e2e
anyhow, i did partially that, let me know if its good enough, (on next push),
i still kept the original `checkRelationshipLabels`, so it will be easier to call it when there is just one label to check, without creating an array

oshoval:
addressed https://github.com/kubevirt/cluster-network-addons-operator/pull/846#discussion_r646125545

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller_test.go
```diff
@@ -0,0 +1,166 @@
...
+		const expectedManagedByLabel = "managed_by_unit_tests"
+		const expectedPartOfLabel = "part_of_unit_tests"
+		const expectedVersionLabel = "version_of_unit_tests"
+
+		BeforeEach(func() {
+			var err error
+			renderData := render.MakeRenderData()
+			objs, err = render.RenderTemplate(testDataFile, &renderData)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(len(objs)).To(Equal(numOfObjectsInDataFile))
```
RamLavi, line 80:
```suggestion
			Expect(objs).To(HaveLen(numOfObjectsInDataFile))
```
also, why do we need this? if RenderTemplate fails to render some objects, won't it return an error?

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller_test.go
```diff
@@ -0,0 +1,166 @@
...
+
+			crLabels := map[string]string{}
+			err = updateObjectsLabels(crLabels, objs)
+			Expect(err).NotTo(HaveOccurred())
+		})
+
+		It("Should find only default relation labels", func() {
+			appLabelKeys := []checkUnit{
```
RamLavi, line 41: we should stick to either appLables/relationshipLabels/RelationLabels terminology (not only here, but throughout the PR).
```suggestion
			expectedRelationshipLabels := []checkUnit{
```

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller_test.go
```diff
@@ -0,0 +1,166 @@
...
+			crLabels[names.COMPONENT_LABEL_KEY] = expectedComponentLabel
+			crLabels[names.MANAGED_BY_LABEL_KEY] = expectedManagedByLabel
+			crLabels[names.PART_OF_LABEL_KEY] = expectedPartOfLabel
+			crLabels[names.VERSION_LABEL_KEY] = expectedVersionLabel
+			err = updateObjectsLabels(crLabels, objs)
+			Expect(err).NotTo(HaveOccurred())
+		})
+
+		It("Should find all labels, overridden by CR labels", func() {
+			appLabelKeys := []checkUnit{
```
RamLavi, line 92: ditto

test/check/check.go
RamLavi: ack

test/check/check.go
```diff
@@ -400,6 +439,16 @@ func checkForGenericDeployment(name, namespace string, checkLabels bool) error {
 	return nil
 }
 
+func checkWorkloadRelationLabels(labelMapList []map[string]string, kind, name string) error {
```
RamLavi, line 442: question, what do you mean by `Workload`?

RamLavi (changes requested):
second pass
looks good, some syntactic comments.

test/check/check.go
oshoval: https://cloud.google.com/kubernetes-engine/docs/how-to/deploying-workloads-overview
those that have template

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller_test.go
oshoval: yep it will, it just so it will be predictable when we add stuff to the template
i can remove it if needed, it was added just an extra validation

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller_test.go
oshoval: so i will stick to appLabel which is the one that already done on almost all the files already
its also shorter, and the one we started with

function names do have the word relationship here and there, imo its fine, wdyt ?

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller_test.go
```diff
@@ -0,0 +1,166 @@
...
+	key           string
+	shouldExist   bool
+	expectedValue string
+}
+
+var _ = Describe("Networkaddonsconfig", func() {
+	const testDataFile = "testdata/data.yaml"
+	const numOfObjectsInDataFile = 4
+
+	Context("When CR is not labeled", func() {
```
RamLavi, line 25: Can we change the unit test to a table, where `crLabels` and the expected `[]checkUnit` are the input?

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller_test.go
RamLavi: I don't think it's needed, provided that `RenderTemplate` fails if not all objects are rendered. 
I don't want to update this param every time I add an object to the data.yaml if it's not necessary

test/check/check.go
RamLavi: ah. understood. then it's a very good naming :) thanks!

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller_test.go
oshoval: I think it will be problmematic, because as we can see current implementation each of the tests / BeforeEach creates the specific crLabels and checkUnit it needs, wrapped by the right context 

creating a table means there will be a mixed section that will create them all

RamLavi: > I think it will be problmematic, because as we can see current implementation each of the tests / BeforeEach creates the specific crLabels and checkUnit it needs, wrapped by the right context

if you move the `updateObjectsLabels()` to the test body, then you can use the input crLabels&checkUnit in it with no issue

>creating a table means there will be a mixed section that will create them all,
context will be lost, as it wraps nicely the current context imo atm
for example when crLables is not tagged, and when crLables is tagged is separated

I see what you mean, I usually convey the context inside each table entry, like so
```
Entry("When CR is not labeled, Should find only default relationship labels", 
updateObjectsLabelsParams{			
expectedCheckUnit: {...},
crLabels: {...}
}),
```
This way the context is not lost. another example [here](https://github.com/kubevirt/cluster-network-addons-operator/blob/master/tools/bumper/cnao_repo_commands_test.go#L131)

oshoval: I understand, thanks, but I find it much more readable as its now,
I would prefer to keep it this way please if possible,
I think the suggested method will be too dense for those tests, as they have parameters and logic that is more readable at the current state.

Moreover as suggested by you before and done, `updateObjectsLabels` more natural place is the current one in the `BeforeEach`

RamLavi: I find it neater, but I won't insist on it.

oshoval:
Thanks Ram

RamLavi:
> Thanks Ram

thank you!

/approve
/retest

PR #895, opened 2021-06-16
linux-bridge, Add Priority class

oshoval:
Add `system-cluster-critical` to `linux-bridge-plugin` DaemonSet.
Since linux bridge's pod isn't bound to a specific node,
(it mostly copies some binaries to the node)
yet its an important pod, assign `system-cluster-critical`
pc to it.
This will make the control plane less sensitive to preemption
than user workloads.

```release-note
None
```


phoracek (approved):
/lgtm
/approve

PR #912, opened 2021-06-24
release-0.53, Update k8s-nmstate to use tagged policy

oshoval:
Since k8s-nmstate has a stable branch,
we can use tagged policy instead of static.
This way, each new release of k8s-nmstate release-0.47
will be auto consumed by CNAO bumper.

```release-note
None
```

### DIFF
--- a/components.yaml
+++ b/components.yaml
@@ -33,7 +33,7 @@ components:
     url: https://github.com/nmstate/kubernetes-nmstate
     commit: 27eef45a9bb2552d05fab4dd37c3c2ff892bdc66
     branch: release-0.47
-    update-policy: static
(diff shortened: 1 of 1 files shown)


oshoval:
Saw this error already, not due to this PR
`Error creating: pods "bridge-marker-" is forbidden: error looking up service account cluster-network-addons/bridge-marker: serviceaccount "bridge-marker" not found`

https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_cluster-network-addons-operator/912/pull-e2e-cnao-lifecycle-k8s-release-0.53/1408035668948422656

/retest

PR #915, opened 2021-06-27
bridge-marker, release-0.53, Set source as tagged

oshoval:
A stable branch was created on bridge-marker for
CNAO release-0.53.
Update manifest to use it.

```release-note
None
```


oshoval:
not due to this PR, which affects cnao-bumper only after it will be merged

/retest

PR #923, opened 2021-07-06
update default provider to k8s 1.21

RamLavi:
**What this PR does / why we need it**:
This PR updates the KBUEVIRTCI_TAG and updated the default provider to k8s-1.21

**Special notes for your reviewer**:

**Release note**:

```release-note
NONE
```

PR #928, opened 2021-07-12
Add alerts infra and basic alert CnaoDown

RamLavi:
This PR add infra support and basic alert `CnaoDown` to CNAO 

manually tested with CRC:
metric:
monitoring-graph
you can see that CNAO being down for >5 minutes is casuing the alert below
alert:
alarm fired

**Release note**:

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -81,7 +81,7 @@ whitespace: $(all_sources)
 	./hack/whitespace.sh --fix
 	touch $@
 
-check: whitespace-check vet goimports-check gen-k8s-check test/unit
+check: whitespace-check vet goimports-check gen-k8s-check test/unit prom-rules-verify
 	./hack/check.sh
 
 whitespace-check: $(all_sources)
@@ -120,6 +120,11 @@ docker-push-operator:
 
 docker-push-registry:
 	docker push $(IMAGE_REGISTRY)/$(REGISTRY_IMAGE):$(IMAGE_TAG)
+
+prom-rules-verify:
+	hack/prom-rule-ci/verify-rules.sh \
+	data/monitoring/prom-rule.yaml \
+	hack/prom-rule-ci/prom-rules-tests.yaml
 
 cluster-up:
 	./cluster/up.sh
@@ -221,4 +226,5 @@ bump-all: bump-nmstate bump-kubemacpool bump-macvtap-cni bump-linux-bridge bump-
 	bump-% \
 	bump-all \
 	gen-k8s \
+	prom-rules-verify \
 	release
--- a/cluster/up.sh
+++ b/cluster/up.sh
@@ -18,6 +18,9 @@ set -ex
 
 SCRIPTS_PATH="$(dirname "$(realpath "$0")")"
 source ${SCRIPTS_PATH}/cluster.sh
+
+export KUBEVIRT_DEPLOY_PROMETHEUS=true
+
 cluster::install
 
 $(cluster::path)/cluster-up/up.sh
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -3,8 +3,6 @@ package main
 import (
 	"context"
 	"flag"
-	"fmt"
-
 	"log"
 	"os"
 	"runtime"
@@ -12,11 +10,8 @@ import (
 	osv1 "github.com/openshift/api/operator/v1"
 	"github.com/operator-framework/operator-sdk/pkg/k8sutil"
 	"github.com/operator-framework/operator-sdk/pkg/leader"
-	"github.com/operator-framework/operator-sdk/pkg/metrics"
 	sdkVersion "github.com/operator-framework/operator-sdk/version"
 	"github.com/spf13/pflag"
-	v1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/util/intstr"
 	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
@@ -24,13 +19,8 @@ import (
 
 	"github.com/kubevirt/cluster-network-addons-operator/pkg/apis"
 	"github.com/kubevirt/cluster-network-addons-operator/pkg/controller"
+	"github.com/kubevirt/cluster-network-addons-operator/pkg/monitoring"
 	"github.com/kubevirt/cluster-network-addons-operator/pkg/util/k8s"
-)
-
-var (
-	metricsHost               = "0.0.0.0"
-	metricsPort         int32 = 8383
-	operatorMetricsPort int32 = 8686
 )
 
 func printVersion() {
@@ -72,7 +62,7 @@ func main() {
 	// Create a new Cmd to provide shared dependencies and start components
 	mgr, err := manager.New(cfg, manager.Options{
 		Namespace:          namespace,
-		MetricsBindAddress: fmt.Sprintf("%s:%d", metricsHost, metricsPort),
+		MetricsBindAddress: monitoring.GetMetricsAddress(),
 		MapperProvider:     k8s.NewDynamicRESTMapper,
 	})
 	if err != nil {
@@ -95,17 +85,6 @@ func main() {
 	// Setup all Controllers
 	if err := controller.AddToManager(mgr); err != nil {
 		log.Printf("failed setting up operator controllers: %v", err)
-		os.Exit(1)
-	}
-
-	servicePorts := []v1.ServicePort{
-		{Port: metricsPort, Name: metrics.OperatorPortName, Protocol: v1.ProtocolTCP, TargetPort: intstr.IntOrString{Type: intstr.Int, IntVal: metricsPort}},
-		{Port: operatorMetricsPort, Name: metrics.CRPortName, Protocol: v1.ProtocolTCP, TargetPort: intstr.IntOrString{Type: intstr.Int, IntVal: operatorMetricsPort}},
-	}
-
-	// Create Service object to expose the metrics port.
-	if _, err = metrics.CreateMetricsService(ctx, cfg, servicePorts); err != nil {
-		log.Printf("failed to create metrics server: %v", err)
 		os.Exit(1)
 	}
 
--- a/data/monitoring/prom-rule.yaml
+++ b/data/monitoring/prom-rule.yaml
@@ -1,0 +1,9 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  labels:
+    prometheus.cnao.io: ""
+  name: prometheus-rules-cluster-network-addons-operator
+  namespace: {{ .Namespace }}
+spec:
+  groups:
--- a/data/monitoring/prom-rule.yaml
+++ b/data/monitoring/prom-rule.yaml
@@ -7,3 +7,15 @@ metadata:
   namespace: {{ .Namespace }}
 spec:
   groups:
+    - name: kubevirt.cnao.rules
+      rules:
+        - expr: sum(up{namespace='{{ .Namespace }}', pod=~'cluster-network-addons-operator-.*'} or vector(0))
+          record: kubevirt_cnao_num_up_operators
+        - alert: CnaoDown
+          annotations:
+            summary: CNAO pod is down.
+            runbook_url: http://kubevirt.io/monitoring/runbooks/CnaoDown
+          expr: kubevirt_cnao_num_up_operators == 0
+          for: 5m
+          labels:
+            severity: warning
--- a/data/monitoring/service.yaml
+++ b/data/monitoring/service.yaml
@@ -1,0 +1,17 @@
+apiVersion: v1
+kind: Service
+metadata:
+  labels:
+    prometheus.cnao.io: ""
+  name: cluster-network-addons-operator-prometheus-metrics
+  namespace: {{ .Namespace }}
+spec:
+  ports:
+    - name: metrics
+      port: 8080
+      protocol: TCP
+      targetPort: metrics
+  selector:
+    prometheus.cnao.io: ""
+  sessionAffinity: None
+  type: ClusterIP
--- a/hack/prom-rule-ci/prom-rules-tests.yaml
+++ b/hack/prom-rule-ci/prom-rules-tests.yaml
@@ -1,0 +1,37 @@
+---
+rule_files:
+  - /tmp/rules.verify.yaml
+
+group_eval_order:
+  - kubevirt.cnao.rules
+
+tests:
+# CnaoDown positive tests
+  - interval: 1m
+    input_series:
+      - series: "kubevirt_cnao_num_up_operators"
+        values: "0 0 0 0 0 0"
+
+    alert_rule_test:
+      - eval_time: 5m
+        alertname: CnaoDown
+        exp_alerts:
+          - exp_annotations:
+              summary: "CNAO pod is down."
+              runbook_url: "http://kubevirt.io/monitoring/runbooks/CnaoDown"
+            exp_labels:
+              severity: "warning"
+# CnaoDown negative tests
+  - interval: 1m
+    input_series:
+      - series: "kubevirt_cnao_num_up_operators"
+        values: "1 1 1 0 1 1"
+      - series: "kubevirt_cnao_num_up_operators"
+        values: "1 1 1 1 1 1"
+      - series: "kubevirt_cnao_num_up_operators"
+        values: "0 0 0 0 0 1"
+
+    alert_rule_test:
+      - eval_time: 5m
+        alertname: CnaoDown
+        exp_alerts:
--- a/hack/prom-rule-ci/verify-rules.sh
+++ b/hack/prom-rule-ci/verify-rules.sh
@@ -1,0 +1,68 @@
+#!/usr/bin/env bash
+
+set -xe
+
+source hack/components/yaml-utils.sh
+
+TMP_PATH=$(mktemp -d -p /tmp -t alerts-test-XXXXXX)
+PROMETHERUS_IMAGE=quay.io/prometheus/prometheus:v2.15.2
+
+teardown() {
+  rm -rf "${TMP_PATH}"
+}
+
+# get_prometheus_rule_spec extracts the spec from the prometheus rule instance
+get_prometheus_rule_spec() {
+  local prom_yaml=$1
+  local output_prom_spec=$2
+
+  prom_rule_spec=$(yaml-utils::get_param ${prom_yaml} spec)
+  echo "${prom_rule_spec}" >${output_prom_spec}
+}
+
+# test_setup prepares all the needed files in the tmp folder
+test_setup() {
+  local prom_yaml=$1
+  local tests_yaml=$2
+  local tmp_prom_spec=$3
+  local tmp_tests_file=$4
+
+  get_prometheus_rule_spec ${prom_yaml} ${tmp_prom_spec}
+  cp ${tests_yaml} ${tmp_tests_file}
+}
+
+verify_prom_spec() {
+  local tmp_prom_spec=$1
+
+  docker run --rm --entrypoint=/bin/promtool \
+  -v ${tmp_prom_spec}:/tmp/rules.verify.yaml:ro \
+  ${PROMETHERUS_IMAGE} \
+  check rules /tmp/rules.verify.yaml
+}
+
+check_rules() {
+  local tmp_prom_spec=$1
+  local tmp_tests_file=$2
+
+  docker run --rm --entrypoint=/bin/promtool \
+  -v ${tmp_prom_spec}:/tmp/rules.verify.yaml:ro \
+  -v ${tmp_tests_file}:/tmp/rules.test.yaml:ro \
+  ${PROMETHERUS_IMAGE} \
+  test rules /tmp/rules.test.yaml
+}
+
+main() {
+  local prom_yaml="${1:?}"
+  local tests_file="${2:?}"
+  local tmp_prom_spec=${TMP_PATH}/prom-spec
+  local tmp_tests_file=${TMP_PATH}/prom-rules-tests.yaml
+
+  trap teardown EXIT SIGINT
+
+  test_setup ${prom_yaml} ${tests_file} ${tmp_prom_spec} ${tmp_tests_file}
+
+  verify_prom_spec ${tmp_prom_spec}
+  check_rules ${tmp_prom_spec} ${tmp_tests_file}
+}
+
+main "$@"
--- a/pkg/components/components.go
+++ b/pkg/components/components.go
@@ -260,6 +260,14 @@ func GetDeployment(version string, operatorVersion string, namespace string, rep
 									Name:  "WATCH_NAMESPACE",
 									Value: "",
 								},
+								{
+									Name:  "MONITORING_NAMESPACE",
+									Value: "openshift-monitoring",
+								},
+								{
+									Name:  "MONITORING_SERVICE_ACCOUNT",
+									Value: "prometheus-k8s",
+								},
 							},
 						},
 					},
--- a/pkg/components/components.go
+++ b/pkg/components/components.go
@@ -13,6 +13,7 @@ import (
 
 	cnao "github.com/kubevirt/cluster-network-addons-operator/pkg/apis/networkaddonsoperator/shared"
 	cnaov1 "github.com/kubevirt/cluster-network-addons-operator/pkg/apis/networkaddonsoperator/v1"
+	"github.com/kubevirt/cluster-network-addons-operator/pkg/names"
 	"github.com/kubevirt/cluster-network-addons-operator/pkg/util/k8s"
 )
 
@@ -137,6 +138,9 @@ func GetDeployment(version string, operatorVersion string, namespace string, rep
 			Annotations: map[string]string{
 				cnaov1.SchemeGroupVersion.Group + "/version": k8s.StringToLabel(operatorVersion),
 			},
+			Labels: map[string]string{
+				names.PROMETHEUS_LABEL_KEY: "",
+			},
 		},
 		Spec: appsv1.DeploymentSpec{
 			Replicas: int32Ptr(1),
@@ -151,7 +155,8 @@ func GetDeployment(version string, operatorVersion string, namespace string, rep
 			Template: corev1.PodTemplateSpec{
 				ObjectMeta: metav1.ObjectMeta{
 					Labels: map[string]string{
-						"name": Name,
+						"name":                     Name,
+						names.PROMETHEUS_LABEL_KEY: "",
 					},
 					Annotations: map[string]string{
 						"description": "cluster-network-addons-operator manages the lifecycle of different Kubernetes network components on top of Kubernetes cluster",
--- a/pkg/components/components.go
+++ b/pkg/components/components.go
@@ -13,6 +13,7 @@ import (
 
 	cnao "github.com/kubevirt/cluster-network-addons-operator/pkg/apis/networkaddonsoperator/shared"
 	cnaov1 "github.com/kubevirt/cluster-network-addons-operator/pkg/apis/networkaddonsoperator/v1"
+	"github.com/kubevirt/cluster-network-addons-operator/pkg/monitoring"
 	"github.com/kubevirt/cluster-network-addons-operator/pkg/names"
 	"github.com/kubevirt/cluster-network-addons-operator/pkg/util/k8s"
 )
@@ -177,6 +178,13 @@ func GetDeployment(version string, operatorVersion string, namespace string, rep
 								Requests: corev1.ResourceList{
 									corev1.ResourceCPU:    resource.MustParse("50m"),
 									corev1.ResourceMemory: resource.MustParse("30Mi"),
+								},
+							},
+							Ports: []corev1.ContainerPort{
+								corev1.ContainerPort{
+									Name:          "metrics",
+									Protocol:      "TCP",
+									ContainerPort: monitoring.GetMetricsPort(),
 								},
 							},
 							Env: []corev1.EnvVar{
--- a/pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
+++ b/pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
@@ -504,6 +504,11 @@ func updateObjectsLabels(crLables map[string]string, objs []*unstructured.Unstru
 		}
 		// Label objects with version of the operator they were created by
 		labels[cnaov1.SchemeGroupVersion.Group+"/version"] = operatorVersionLabel
+		labels[names.PROMETHEUS_LABEL_KEY] = ""
+		err = updateObjectTemplateLabels(obj, labels, []string{names.PROMETHEUS_LABEL_KEY})
+		if err != nil {
+			return err
+		}
 
 		appLabelKeys := []string{names.COMPONENT_LABEL_KEY, names.PART_OF_LABEL_KEY, names.VERSION_LABEL_KEY, names.MANAGED_BY_LABEL_KEY}
 
--- a/pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
+++ b/pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
@@ -92,7 +92,7 @@ func Add(mgr manager.Manager) error {
 	}
 	clusterInfo.SCCAvailable = sccAvailable
 
-	addMonitorServiceResources, err := isMonitoringAvailable(clientset)
+	addMonitorServiceResources, err := IsMonitoringAvailable(clientset)
 	if err != nil {
 		// we don't want CNAO to fail only if monitoring cannot be activated.
 		addMonitorServiceResources = false
@@ -608,7 +608,7 @@ func isSCCAvailable(c kubernetes.Interface) (bool, error) {
 }
 
 // isMonitoringAvailable checks if we can deploy the monitoring component
-func isMonitoringAvailable(c kubernetes.Interface) (bool, error) {
+func IsMonitoringAvailable(c kubernetes.Interface) (bool, error) {
 	prometheusRuleResourceAvailable, err := isResourceAvailable(c, "customresourcedefinitions/prometheusrules.monitoring.coreos.com", "apiextensions.k8s.io", "v1")
 	if err != nil {
 		return false, errors.Wrap(err, "failed to check if prometheusRule resource is available")
--- a/pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
+++ b/pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
@@ -9,6 +9,7 @@ import (
 	"strings"
 	"time"
 
+	monitoringv1 "github.com/coreos/prometheus-operator/pkg/apis/monitoring/v1"
 	osv1 "github.com/openshift/api/operator/v1"
 	osnetnames "github.com/openshift/cluster-network-operator/pkg/names"
 	"github.com/pkg/errors"
@@ -90,6 +91,14 @@ func Add(mgr manager.Manager) error {
 		return fmt.Errorf("failed to check for availability of SCC: %v", err)
 	}
 	clusterInfo.SCCAvailable = sccAvailable
+
+	addMonitorServiceResources, err := isMonitoringAvailable(clientset)
+	if err != nil {
+		// we don't want CNAO to fail only if monitoring cannot be activated.
+		addMonitorServiceResources = false
+		log.Printf("failed to check for availability of Monitoring namespace: %v", err)
+	}
+	clusterInfo.MonitoringAvailable = addMonitorServiceResources
 
 	return add(mgr, newReconciler(mgr, namespace, clusterInfo))
 }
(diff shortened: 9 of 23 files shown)


RamLavi:
/hold
this PR depends on PR #923 to be merged.

cmd/manager/main.go
```diff
@@ -72,7 +72,7 @@ func main() {
 	// Create a new Cmd to provide shared dependencies and start components
 	mgr, err := manager.New(cfg, manager.Options{
 		Namespace:          namespace,
-		MetricsBindAddress: fmt.Sprintf("%s:%d", metricsHost, metricsPort),
+		MetricsBindAddress: fmt.Sprintf("%s:%d", operatorSdkMetricsHost, operatorSdkHttpMetricsPort),
```
RamLavi, line 76: I'm not certain as to if I can/should use the same port as operatorSDK (8383), 
when I tried it failed on "address already set".
if so then I'm not sure if the which port value does `MetricsBindAddress` should contain

RamLavi:
/hold cancel 
#923 is merged, current PR rebased.

will still keep the PR under WIP until I check this under crc with prometheus deployed

RamLavi:
/hold
relying on #942 bug fix.

cmd/manager/main.go
RamLavi: I ended up removing the operator-sdk metric altogether..  the way I see it, if we don't use them,  we shouldn't keep them

pkg/monitoring/metrics.go
```diff
@@ -0,0 +1,30 @@
...
+func startPrometheusEndpoint() error {
+	go func() {
+		mux := http.NewServeMux()
+		mux.Handle("/metrics", promhttp.Handler())
+		server := http.Server{
+			Addr:    "0.0.0.0:8443",
+			Handler: mux,
+		}
+		log.Printf("Starting Prometheus metrics endpoint server")
+		err := server.ListenAndServe()
```
RamLavi, line 24: @phoracek  This is still an open issue whether we must use TLS..
pending getting a response from OCP monitoring slack,

RamLavi:
/retest
ci hiccup

RamLavi:
/hold
we still need to get an answer egarding the TLS communication

pkg/monitoring/metrics.go
RamLavi: According to OCP monitoring slack, we're good.

qinqon:
/hold 

Since we are using controller-runtime here and it has already a metrics capacity I think it would be better to ingrate it 

https://book.kubebuilder.io/reference/metrics.html

RamLavi:
[here](https://github.com/kubevirt/cluster-network-addons-operator/compare/14effb9256144e426ae25b3e4bd7b7d8cbbb9390..a62d445754c22b0b21c749827e0b9d1350ac2e4b) I changed the condition to deploy monitoring manifests to only checking relevant CRDs, and also added another env-var for the serviceAccount name

RamLavi:
> /hold
> 
> Since we are using controller-runtime here and it has already a metrics capacity I think it would be better to ingrate it
> 
> https://book.kubebuilder.io/reference/metrics.html

fixed [here](https://github.com/kubevirt/cluster-network-addons-operator/compare/a62d445754c22b0b21c749827e0b9d1350ac2e4b..6bf3b770b245c610dd81b4da47003d185c7c7e50)

cluster/up.sh
```diff
@@ -18,6 +18,13 @@ set -ex
 
 SCRIPTS_PATH="$(dirname "$(realpath "$0")")"
 source ${SCRIPTS_PATH}/cluster.sh
+
+export KUBEVIRT_DEPLOY_PROMETHEUS=true
+if [[ ($KUBEVIRT_PROVIDER =~ k8s-1\.1.*) || ($KUBEVIRT_PROVIDER =~ k8s-1.20) ]]; then
+	  echo "Prometheus is only supported for providers >= k8s-1.21\n Disabling Prometheus"
```
oshoval, line 24: nit: ident

cmd/manager/main.go
RamLavi: update - brought it back :)

cluster/up.sh
RamLavi: thanks! [Done](https://github.com/kubevirt/cluster-network-addons-operator/compare/7a61ef72f51fcc207386bac49d2e59627e0349b2..cacead5ccc2b09349d5ce311de4f84314d3e6339)

cluster/up.sh
```diff
@@ -18,6 +18,13 @@ set -ex
 
 SCRIPTS_PATH="$(dirname "$(realpath "$0")")"
 source ${SCRIPTS_PATH}/cluster.sh
+
+export KUBEVIRT_DEPLOY_PROMETHEUS=true
+if [[ ($KUBEVIRT_PROVIDER =~ k8s-1\.1.*) || ($KUBEVIRT_PROVIDER =~ k8s-1.20) ]]; then
+    echo "Prometheus is only supported for providers >= k8s-1.21\n Disabling Prometheus"
+    export KUBEVIRT_DEPLOY_PROMETHEUS=false
+fi
+
```
qinqon, line 27: We should do this at kubevirtci right ?

cmd/manager/main.go
```diff
@@ -72,7 +62,7 @@ func main() {
 	// Create a new Cmd to provide shared dependencies and start components
 	mgr, err := manager.New(cfg, manager.Options{
 		Namespace:          namespace,
-		MetricsBindAddress: fmt.Sprintf("%s:%d", metricsHost, metricsPort),
+		MetricsBindAddress: monitoring.GetMetricsAddress(),
```
qinqon, line 65: We are not adding an argument to change metrics bind address from the container and default values https://github.com/kubernetes-sigs/controller-runtime/blob/b704f447ea7c8f7059c6665143a4aa1f6da28328/pkg/metrics/listener.go#L30 are good enough, let's just not set this so we use the default metrics bind address.

data/monitoring/service.yaml
```diff
@@ -0,0 +1,17 @@
...
+kind: Service
+metadata:
+  labels:
+    prometheus.cnao.io: ""
+  name: cluster-network-addons-operator-prometheus-metrics
+  namespace: {{ .Namespace }}
+spec:
+  ports:
+    - name: metrics
+      port: 443
```
qinqon, line 11: If we are using the default 8080 would be good also since we are not using TLS 443 does not make sense here.

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
```diff
@@ -598,6 +605,25 @@ func isSCCAvailable(c kubernetes.Interface) (bool, error) {
...
+	}
+
+	if prometheusRuleResourceAvailable && serviceMonitorResourceAvailable {
+		return true, nil
+	}
+
+	log.Printf("will not deploy monitoring manifests: not all monitoring resources are available: %s: %v, %s, %v", monitoringv1.PrometheusRuleKind, prometheusRuleResourceAvailable, monitoringv1.ServiceMonitorsKind, serviceMonitorResourceAvailable)
+	return false, nil
+}
+
```
qinqon, line 626: Return false if there is an error but print the error we don't want CNAO to fail only if monitoring cannot be activated.

pkg/monitoring/metrics.go
```diff
@@ -18,3 +21,21 @@ func init() {
...
+}
+
+func GetMonitoringServiceAccount() string {
+	monitoringServiceAccountFromEnv := os.Getenv("MONITORING_SERVICE_ACCOUNT")
+
+	if monitoringServiceAccountFromEnv != "" {
+		return monitoringServiceAccountFromEnv
+	}
+	return defaultServiceAccountName
+}
```
qinqon, line 41: We are already at "monitoring" package we can do `monitoring.GetNamespace` `monitorig.GetServiceAccount`

pkg/network/monitoring.go
```diff
@@ -0,0 +1,31 @@
...
+	data.Data["MonitoringNamespace"] = monitoring.GetMonitoringNamespace()
+	data.Data["MonitoringServiceAccount"] = monitoring.GetMonitoringServiceAccount()
+
+	objs, err := render.RenderDir(filepath.Join(manifestDir, "monitoring"), &data)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to render monitoring manifests")
+	}
+
+	return objs, nil
+}
```
qinqon, line 31: This has nothing to do with "network" package, it belongs to "monitoring" package so we do `monitoring.Render`

test/check/check.go
```diff
@@ -120,6 +122,36 @@ func PlacementListFromComponentDaemonSets(component Component) ([]cnao.Placement
 	return placementList, nil
 }
 
+func getMonitoringNamespace() (string, error) {
```
qinqon, line 125: Why don't we use the production fruncion ?

test/check/check.go
```diff
@@ -120,6 +122,36 @@ func PlacementListFromComponentDaemonSets(component Component) ([]cnao.Placement
...
+
+	ns := corev1.Namespace{}
+	err = framework.Global.Client.Get(context.Background(), types.NamespacedName{Name: monitoringNamespace}, &ns)
+	if err == nil {
+		return true, nil
+	} else if apierrors.IsNotFound(err) {
+		return false, nil
+	}
+
+	return false, err
```
qinqon, line 152: We have to make this more idiomatic doing some kind of fail fast.

```golang
if err != nil {
  if apierrors.IsNotFound(err) {
    return false, nil
  }
  return false, err
}
return false, nil
```

test/check/check.go
```diff
@@ -245,6 +277,21 @@ func CheckForLeftoverObjects(currentVersion string) {
 	err = framework.Global.Client.List(context.Background(), &serviceAccounts, &listOptions)
 	Expect(err).NotTo(HaveOccurred())
 	Expect(serviceAccounts.Items).To(BeEmpty(), "Found leftover objects from the previous operator version")
+
+	services := corev1.ServiceList{}
+	err = framework.Global.Client.List(context.Background(), &services, &listOptions)
+	Expect(err).NotTo(HaveOccurred())
```
qinqon, line 283: Expect(framework.Global.Client.List(context.Background(), &services, &listOptions)).To(Succees())

test/check/check.go
```diff
@@ -245,6 +277,21 @@ func CheckForLeftoverObjects(currentVersion string) {
...
+	Expect(services.Items).To(BeEmpty(), "Found leftover objects from the previous operator version")
+
+	serviceMonitors := monitoringv1.ServiceMonitorList{}
+	err = framework.Global.Client.List(context.Background(), &serviceMonitors, &listOptions)
```
qinqon, line 287: ditto

test/check/check.go
```diff
@@ -245,6 +277,21 @@ func CheckForLeftoverObjects(currentVersion string) {
...
+	Expect(err).NotTo(HaveOccurred())
+	Expect(serviceMonitors.Items).To(BeEmpty(), "Found leftover objects from the previous operator version")
+
+	prometheusRules := monitoringv1.PrometheusRuleList{}
+	err = framework.Global.Client.List(context.Background(), &prometheusRules, &listOptions)
```
qinqon, line 292: ditto

test/e2e/lifecycle/main_test.go
```diff
@@ -39,6 +40,11 @@ var _ = BeforeSuite(func() {
 	By("Adding custom resource scheme to framework")
 	err := framework.AddToFrameworkScheme(apis.AddToScheme, &cnaov1.NetworkAddonsConfigList{})
 	Expect(err).ToNot(HaveOccurred())
+
+	err = framework.AddToFrameworkScheme(monitoringv1.AddToScheme, &monitoringv1.ServiceMonitorList{})
+	Expect(err).ToNot(HaveOccurred())
+	err = framework.AddToFrameworkScheme(monitoringv1.AddToScheme, &monitoringv1.PrometheusRuleList{})
+	Expect(err).ToNot(HaveOccurred())
```
qinqon, line 47: Use one line per check with `Exec(framework.AddToFrameworkSchema...).To(Succeed())`

Makefile
```diff
@@ -121,6 +121,11 @@ docker-push-operator:
 docker-push-registry:
 	docker push $(IMAGE_REGISTRY)/$(REGISTRY_IMAGE):$(IMAGE_TAG)
 
+prom-rules-verify:
```
qinqon, line 124: This are kind of integration tests and belong to ./test dir I think and have to be run as part of the integration tests

qinqon (changes requested):
Good work, it's a huge task, added some comments to make world a little better.

cmd/manager/main.go
```diff
@@ -72,7 +61,7 @@ func main() {
 	// Create a new Cmd to provide shared dependencies and start components
 	mgr, err := manager.New(cfg, manager.Options{
 		Namespace:          namespace,
-		MetricsBindAddress: fmt.Sprintf("%s:%d", metricsHost, metricsPort),
+		MetricsBindAddress: "0",
```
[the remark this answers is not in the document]
RamLavi: I eventually moved it back in order to use conrtoller-runtime's prometheus endpint.

cluster/up.sh
```diff
@@ -18,6 +18,13 @@ set -ex
 
 SCRIPTS_PATH="$(dirname "$(realpath "$0")")"
 source ${SCRIPTS_PATH}/cluster.sh
+
+export KUBEVIRT_DEPLOY_PROMETHEUS=true
+if [[ ($KUBEVIRT_PROVIDER =~ k8s-1\.1.*) || ($KUBEVIRT_PROVIDER =~ k8s-1.20) ]]; then
+	  echo "Prometheus is only supported for providers >= k8s-1.21\n Disabling Prometheus"
+    export KUBEVIRT_DEPLOY_PROMETHEUS=false
+fi
```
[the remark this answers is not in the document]
RamLavi: I tend to agree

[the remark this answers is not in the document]
oshoval: I think kubevirtci doesn't want to deploy Prometheus by default ?

cmd/manager/main.go
RamLavi: not sure I got you. why do we not want to use a specific port, that we can also reference to on components.go?

cluster/up.sh
RamLavi: removed this condition. 
To be clear - this means the from this commit onward `make cluster-up` with provider<1.21 will fail on kubevirtci (which is ok, the error message there is quite clear)

pkg/components/components.go
```diff
@@ -179,6 +179,13 @@ func GetDeployment(version string, operatorVersion string, namespace string, rep
 									corev1.ResourceMemory: resource.MustParse("30Mi"),
 								},
 							},
+							Ports: []corev1.ContainerPort{
+								corev1.ContainerPort{
+									Name:          "metrics",
+									Protocol:      "TCP",
+									ContainerPort: 8443,
```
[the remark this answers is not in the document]
RamLavi: didn't find const for TCP on http/net libraries.. do you know of any other places to look?

cmd/manager/main.go
RamLavi: ok I change `monitoring.GetMetricsAddress()` to use `metrics.DefaultBindAddress` , is that alright?

data/monitoring/prom-rule.yaml
```diff
@@ -0,0 +1,11 @@
...
+kind: PrometheusRule
+metadata:
+  labels:
+    prometheus.cnao.io: ""
+  name: prometheus-rules-cluster-network-addons-operator
+  namespace: {{ .Namespace }}
+spec:
+  groups:
+    - name: kubevirt.cnao.rules
+      rules:
```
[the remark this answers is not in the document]
RamLavi: no, fixed that.

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
RamLavi: I will change in the function calling this on the controller so that it will not fail and just log it.

pkg/monitoring/metrics.go
RamLavi: Done

pkg/network/monitoring.go
RamLavi: well the entire Render function is in `network.go`
I moved it to `pkg/monitoring/metrics.go`, but not sure it is comfy here.. 
what do you say?

test/check/check.go
RamLavi: I see no reason why not :)

test/check/check.go
RamLavi: removed the function entirely in favor on the production function

test/check/check.go
RamLavi: done

test/check/check.go
RamLavi: done

test/check/check.go
RamLavi: done

test/e2e/lifecycle/main_test.go
RamLavi: done

Makefile
RamLavi: well it's technically unit test.. I think it should be part of unit tests.
it's similar to what they do on [kubevirt](https://github.com/kubevirt/kubevirt/blob/main/Makefile#L167)

RamLavi:
@qinqon fixes [here](https://github.com/kubevirt/cluster-network-addons-operator/compare/cacead5ccc2b09349d5ce311de4f84314d3e6339..80b49dd132735ebedd3986b521017f46ca871581)

cmd/manager/main.go
qinqon: We don't need to pass anything and we don't need "GetMetricsAddress" since we are using controller-runtime's defaults.

data/monitoring/prom-rule.yaml
```diff
@@ -0,0 +1,20 @@
...
+        - expr: sum(up{namespace='{{ .Namespace }}', pod=~'cluster-network-addons-operator-.*'} or vector(0))
+          record: kubevirt_cnao_num_up_operators
+        - alert: CnaoDown
+          annotations:
+            summary: CNAO pod is down.
+          expr: kubevirt_cnao_num_up_operators == 0
+          for: 5m
```
oshoval, line 18: aren't 5 minutes is too much ?
one might want to know about the alert faster

hack/prom-rule-ci/prom-rules-tests.yaml
```diff
@@ -0,0 +1,21 @@
+---
+rule_files:
+  - /tmp/rules.verify.yaml
+
+group_eval_order:
+  - kubevirt.cnao.rules
+
+tests:
+  - interval: 1m
+    input_series:
```
oshoval, line 10: might need also a non zero vector for the tests?

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
```diff
@@ -593,6 +607,26 @@ func isSCCAvailable(c kubernetes.Interface) (bool, error) {
...
+	serviceMonitorResourceAvailable, err := isResourceAvailable(c, "customresourcedefinitions/servicemonitors.monitoring.coreos.com", "apiextensions.k8s.io", "v1")
+	if err != nil {
+		return false, errors.Wrap(err, "failed to check if serviceMonitor resource is available")
+	}
+
+	if prometheusRuleResourceAvailable && serviceMonitorResourceAvailable {
+		return true, nil
+	}
+
+	log.Printf("will not deploy monitoring manifests: not all monitoring resources are available: %s: %v, %s, %v", monitoringv1.PrometheusRuleKind, prometheusRuleResourceAvailable, monitoringv1.ServiceMonitorsKind, serviceMonitorResourceAvailable)
```
oshoval, line 626: maybe worth to return this error string as err in the next line ?
this way the caller will print it as err, as it does for the other errors above

pkg/monitoring/metrics.go
```diff
@@ -0,0 +1,76 @@
...
+
+func getServiceAccount() string {
+	monitoringServiceAccountFromEnv := os.Getenv("MONITORING_SERVICE_ACCOUNT")
+
+	if monitoringServiceAccountFromEnv != "" {
+		return monitoringServiceAccountFromEnv
+	}
+	return defaultServiceAccountName
+}
```
oshoval, line 76: maybe can be combined to one function which get an env var name + default value ?

pkg/network/network.go
```diff
@@ -141,6 +143,13 @@ func Render(conf *cnao.NetworkAddonsConfigSpec, manifestDir string, openshiftNet
 	}
 	objs = append(objs, o...)
 
+	// render Monitoring Service
+	o, err = monitoring.RenderMonitoring(manifestDir, clusterInfo.MonitoringAvailable)
+	if err != nil {
+		return nil, err
+	}
+	objs = append(objs, o...)
```
oshoval, line 151: maybe worth to do as an else, if there is err, the `o` might be partial theoretically, 
(not good practice to use a value when there is error)

test/check/check.go
```diff
@@ -553,6 +591,51 @@ func checkForMutatingWebhookConfiguration(name string) error {
 	return nil
 }
 
+func checkForService(name string) error {
```
oshoval, line 594: in the following functions, it makes the monitoring mandatory isn't it ?
opposed to the fact that you allow CNAO to be deployed even if the monitoring has an error

data/monitoring/prom-rule.yaml
RamLavi: I don't think there is cause to get it faster.
This is aligned with kubevirt's similar alert [VirtAPIDown](https://github.com/kubevirt/kubevirt/blob/main/pkg/virt-operator/resource/generate/components/crds.go#L564-L572)

hack/prom-rule-ci/prom-rules-tests.yaml
RamLavi: Do you mean a test checking that the alert doesn't fire on (for example) `"1 1 1 1 0 1"`?
@sradco is this possible on this unit test infra?

hack/prom-rule-ci/verify-rules.sh
```diff
@@ -0,0 +1,68 @@
+#!/usr/bin/env bash
+
+set -xe
+
+source hack/components/yaml-utils.sh
+
+TMP_PATH=$(mktemp -d -p /tmp -t alerts-test-XXXXXX)
+PROMETHERUS_IMAGE=docker.io/prom/prometheus:v2.15.2
```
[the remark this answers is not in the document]
RamLavi: @erkanerol is there a quay mirror for this?

hack/prom-rule-ci/verify-rules.sh
```diff
@@ -0,0 +1,68 @@
...
+  test rules /tmp/rules.test.yaml
+}
+
+main() {
+  local prom_yaml="${1:?}"
+  local tests_file="${2:?}"
+  local tmp_prom_spec=${TMP_PATH}/prom-spec
+  local tmp_tests_file=${TMP_PATH}/prom-rules-tests.yaml
+
+  trap teardown EXIT SIGINT SIGTERM
```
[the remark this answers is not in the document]
RamLavi: I don't see it running twice (from line [8677](https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_cluster-network-addons-operator/928/pull-cluster-network-addons-operator-unit-test/1420404095063166976#1:build-log.txt%3A8677) to line [8801](https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_cluster-network-addons-operator/928/pull-cluster-network-addons-operator-unit-test/1420404095063166976#1:build-log.txt%3A8801))

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
```diff
@@ -91,6 +92,14 @@ func Add(mgr manager.Manager) error {
 	}
 	clusterInfo.SCCAvailable = sccAvailable
 
+	addMonitorServiceResources, err := IsMonitoringAvailable(clientset)
+	if err != nil {
+		// we don't want CNAO to fail only if monitoring cannot be activated.
+		addMonitorServiceResources = false
+		fmt.Printf("failed to check for availability of Monitoring namespace: %v", err)
```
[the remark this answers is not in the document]
RamLavi: it could be nice to have, but since statusManager (and correspondingly the eventEmitter) are only initialized when the reconciler start to work and this code runs before that, then I'm not sure it's worth the trouble.

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
```diff
@@ -504,6 +513,11 @@ func updateObjectsLabels(crLables map[string]string, objs []*unstructured.Unstru
 		}
 		// Label objects with version of the operator they were created by
 		labels[cnaov1.SchemeGroupVersion.Group+"/version"] = operatorVersionLabel
+		labels[names.PROMETHEUS_LABEL_KEY] = ""
+		err = updateObjectTemplateLabels(obj, labels, []string{names.PROMETHEUS_LABEL_KEY})
+		if err != nil {
+			return err
+		}
```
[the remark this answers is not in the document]
RamLavi: `appLabelKeys` serves both `updateLabelsFromCR` and `updateObjectTemplateLabels` and `PROMETHEUS_LABEL_KEY` doesn't relate to the former function (the value is not fetched from the CR - but rather is always `""`) So I deemed them separate cases and used it on its own merit

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
RamLavi: It's kind of semantic, since I'm not really failing CNAO process if `IsMonitoringAvailable()` returns an error.
The difference between this printf and the other errors, is that in this case - nothing really failed. The cluster just doesn't have the right CRDs to run the monitoring instances, so we can't deploy monitoring. 
So instead of returning an error, I'm adding this printf that also sets it on the logs so it will be heard..

hack/prom-rule-ci/prom-rules-tests.yaml
oshoval: or does fire on `1 1 1 1 1` if needed

RamLavi: it should only fire if it is 0 for more than 5 consecutive polls
Talked to shirly and there is a negative test scenario, will change, thanks!

hack/prom-rule-ci/verify-rules.sh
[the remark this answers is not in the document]
oshoval: It depends on the signal you received,
if you will get SIGTERM from prow, for example when you force push, this what what it does
it will run both the SIGTERM and the EXIT, we saw it in the past on kubevirtci, 
anyhow imo SIGTERM is redundant

pkg/monitoring/metrics.go
RamLavi: the fact that there is a value from env var or from default value is implementation detail, which we want to hide from the user using this API. it should just return one of them, preferably the one from the env var.

pkg/network/network.go
RamLavi: if an error s returned from a function, the value of the artifacts is not defined ( could be partial, could be nil, depends on implementation). it is up to the function calling to it to check if `err!= nil` and only otherwise use the returned variables.

oshoval: Yes of course, this what i meant, just missed for a moment the `return`, because had in mind the monitoring is not mandatory

test/check/check.go
RamLavi: The tests checking monitoring components should be skipped if the CRDs are not deployed..
I do skip the test checking it specifically if the CRDs are not deployed, but you might be right in the case of [AllComponents](https://github.com/kubevirt/cluster-network-addons-operator/pull/928/files#diff-dd3d0d54317e754fd95de914de01a73b5578626b42e755fdad4998186d552600R89)... hmmm.. let me see if I can dynamically remove it

hack/prom-rule-ci/prom-rules-tests.yaml
RamLavi: Done

hack/prom-rule-ci/verify-rules.sh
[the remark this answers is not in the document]
RamLavi: Done

test/check/check.go
oshoval: Best lets inject an error that won't deploy monitoring and see what fails ? (unless you did that already)
it will what must be changed first

test/check/components.go
```diff
@@ -76,13 +80,20 @@ var (
 			"macvtap-cni",
 		},
 	}
+	MonitoringComponent = Component{
+		ComponentName:  "Monitoring",
+		Service:        "cluster-network-addons-operator-prometheus-metrics",
+		ServiceMonitor: "prometheus-rules-cluster-network-addons-operator",
```
oshoval, line 86: those two lines should be same value?
ServiceMonitor, PrometheusRule

test/e2e/workflow/deployment_test.go
```diff
@@ -89,6 +90,14 @@ var _ = Describe("NetworkAddonsConfig", func() {
 				[]Component{MacvtapComponent},
 			),
 		)
+		It("should deploy prometheus if NetworkAddonsConfigSpec is not empty", func() {
+			isMonitoringDeployed, err := networkaddonsconfig.IsMonitoringAvailable(framework.Global.KubeClient)
+			Expect(err).ToNot(HaveOccurred())
+			if !isMonitoringDeployed {
```
oshoval, line 96: why allowing in the tests to be not deployed and skip ?

RamLavi:
/retest
seems unrelated

test/check/components.go
RamLavi: I mainly did it cause kubevirt's [serviceMonitor](https://github.com/kubevirt/kubevirt/blob/main/pkg/virt-operator/resource/generate/components/crds.go#L496) and [prometheusRule](https://github.com/kubevirt/kubevirt/blob/main/pkg/virt-operator/resource/generate/components/crds.go#L539) have the same name, but It's not really mandatory and I haven't seen it done in other projects...
so.. do you think I should change it?

oshoval: maybe call them `prometheus-cnao-service-monitor`
and
`prometheus-cnao-rule` ?

test/e2e/workflow/deployment_test.go
RamLavi: well it depends.. This logic is simply mimicking what we do in the production code.
But what it means in the context of this tests is that if we don't skip this test then k8s providers <1.21 will just fail this test/e2e/workflow lane.
Do you think we should not care of old providers?

oshoval: ye lets not care about them
if and only if we will need it in the future we can reconsider, but i bet we won't need it
so better to be strict + less to maintain

test/check/check.go
RamLavi: I was thinking of wrapping `AllComponents` with `getAllComponents()` that will check and exclude monitoring if it is not deployed.. what do you think?

cmd/manager/main.go
RamLavi: I can  remove it, but then how do you feel about the fact that we're already parsing the port on `GetMetricsPort` using  `DefaultBindAddress` ([here](https://github.com/kubevirt/cluster-network-addons-operator/pull/928/files#diff-abbf397a1a9176a1fd8d85bc268d3acf4d55fb789feabd1c2ae818395e26083eR69))?

My thinking was that if we're already using DefaultBindAddress exclicitly there then I should be consistent.. waiting to see what you say

hack/prom-rule-ci/verify-rules.sh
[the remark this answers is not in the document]
erkanerol: I think it is here https://quay.io/repository/prometheus/prometheus

data/monitoring/prom-rule.yaml
erkanerol: Please check https://github.com/kubevirt/kubevirt/issues/5383 and ensure that this alert works.

erkanerol: @sradco Do we have a naming convention for alerts? Should it start with `Kubevirt`? See https://github.com/kubevirt/monitoring/tree/main/runbooks

sradco: @RamLavi No naming convention, but you are missing runbook_url and a PR for the runbook in kubevirt/monitoring

data/monitoring/prom-rule.yaml
```diff
@@ -0,0 +1,20 @@
...
+          labels:
+            severity: critical
```
sradco, line 20: @RamLavi  Does this alert severity fit the description?
Critical - When the service is down and you loss critical functionality, an action is required immediately

RamLavi: if CNAO operator is down then there's no one to make sure that CNAO's components are running, one of them is Kubemacpool. if Kubemacpool is down, then VMs can't be created. 
so considering your definition, maybe warning is more fitting?

data/monitoring/prom-rule.yaml
RamLavi: > Please check [kubevirt/kubevirt#5383](https://github.com/kubevirt/kubevirt/issues/5383) and ensure that this alert works.

I'm not using `labeldrop` so it shouldn't occur in this case. Also, I checked this on the GUI using CRC (see images in the PR description)

>@RamLavi No naming convention, but you are missing runbook_url and a PR for the runbook in kubevirt/monitoring

You are right. I will issue the relevant PRs for runbooks soon.

data/monitoring/prom-rule.yaml
RamLavi: changed to Warning. Done

test/check/check.go
RamLavi: So.. if we only keep the most updated provider, then I think we can keep `AllComponents`.
Since we always deploy prometheus, this the CRDs should always be deployed, and we should make sure that it acts as expected in our environment.
However, on production we don't want to force the user to deploy monitoring or CNAO will fail.

test/check/components.go
RamLavi: I think we should keep it consistent, we don't use cnao abbrev on CNAO instances, 
so I think we should only change the `ServiceMonitor` to `service-monitor-cluster-network-addons-operator`

test/e2e/workflow/deployment_test.go
RamLavi: I agree. Done

test/e2e/workflow/deployment_test.go
```diff
@@ -89,6 +89,9 @@ var _ = Describe("NetworkAddonsConfig", func() {
 				[]Component{MacvtapComponent},
 			),
 		)
+		It("should deploy prometheus if NetworkAddonsConfigSpec is not empty", func() {
+			testConfigCreate(gvk, cnao.NetworkAddonsConfigSpec{MacvtapCni: &cnao.MacvtapCni{}}, []Component{MacvtapComponent, MonitoringComponent})
```
oshoval, line 93: why actually not to deploy prometheus even if spec is empty ?
some of the gauges might still need to be available aren't they ?

data/monitoring/service.yaml
```diff
@@ -0,0 +1,17 @@
...
+kind: Service
+metadata:
+  labels:
+    prometheus.cnao.io: ""
+  name: cluster-network-addons-operator-prometheus-metrics
+  namespace: {{ .Namespace }}
+spec:
+  ports:
+    - name: metrics
+      port: 8080
```
oshoval, line 11: should this be a variable, that you set via the Render function in case the port is not the default one ?

RamLavi: it does not.. I'm exposing port internal 8080 to external port 8080. the internal is already defined as port `metrics` (in the next line). this 8080 is actually a free number that I chose to be the same to avoid future confusion..

test/e2e/workflow/deployment_test.go
RamLavi: it's simply the way CNAO deploys components, it only does it if the CR is there and not empty.

oshoval:
lgtm from me, thanks,
just letting others review as well

data/monitoring/prom-rule.yaml
sradco: @RamLavi The runbook_url should be part of this PR.

sradco: Example of runbook_url https://github.com/kubevirt/hyperconverged-cluster-operator/pull/1448/files

RamLavi:
> /hold
> 
> Some answers missing.

yes let's also wait on @sradco lgtm, I'm currently waiting for the new runbook url format.

RamLavi:
@sradco Added runbook URL [here](https://github.com/kubevirt/cluster-network-addons-operator/compare/b035e4aac16df185243d35005a596cf1cb9086f9..0bdee788e60f5aabeef99004d1423b4a5707dc6c)

RamLavi:
removed the prefix [here](https://github.com/kubevirt/cluster-network-addons-operator/compare/0bdee788e60f5aabeef99004d1423b4a5707dc6c..22b60f25bda44fe9260c98a0df62e87d0f2026cd), as it messes with the prometheus unit-test module
/retest

oshoval:
To remove the hold?

RamLavi:
> To remove the hold?

yeah

/hold cancel

PR #944, opened 2021-07-26
Add kubevirt_cnao_cr_ready metric and NetworkAddonsConfigNotReady alert

RamLavi:
This PR adds the Add kubevirt_cnao_cr_ready metric and NetworkAddonsConfigNotReady alert.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/data/monitoring/prom-rule.yaml
+++ b/data/monitoring/prom-rule.yaml
@@ -19,3 +19,11 @@ spec:
           for: 5m
           labels:
             severity: warning
+        - alert: NetworkAddonsConfigNotReady
+          annotations:
+            summary: CNAO CR NetworkAddonsConfig is not ready.
+            runbook_url: http://kubevirt.io/monitoring/runbooks/NetworkAddonsConfigNotReady
+          expr: sum(kubevirt_cnao_cr_ready{namespace='{{ .Namespace }}'} or vector(0)) == 0
+          for: 5m
+          labels:
+            severity: warning
--- a/pkg/controller/statusmanager/status_manager.go
+++ b/pkg/controller/statusmanager/status_manager.go
@@ -22,6 +22,7 @@ import (
 	cnao "github.com/kubevirt/cluster-network-addons-operator/pkg/apis/networkaddonsoperator/shared"
 	cnaov1 "github.com/kubevirt/cluster-network-addons-operator/pkg/apis/networkaddonsoperator/v1"
 	eventemitter "github.com/kubevirt/cluster-network-addons-operator/pkg/eventemitter"
+	"github.com/kubevirt/cluster-network-addons-operator/pkg/monitoring"
 	"github.com/kubevirt/cluster-network-addons-operator/pkg/names"
 )
 
@@ -83,6 +84,7 @@ func (status *StatusManager) Set(reachedAvailableLevel bool, conditions ...condi
 	for i := 0; i < conditionsUpdateRetries; i++ {
 		err := status.set(reachedAvailableLevel, conditions...)
 		if err == nil {
+			monitoring.SetReadyGauge(reachedAvailableLevel)
 			log.Print("Successfully updated status conditions")
 			return
 		}
--- a/pkg/monitoring/metrics.go
+++ b/pkg/monitoring/metrics.go
@@ -7,6 +7,7 @@ import (
 	"strconv"
 
 	"github.com/pkg/errors"
+	"github.com/prometheus/client_golang/prometheus"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"sigs.k8s.io/controller-runtime/pkg/metrics"
 
@@ -19,8 +20,24 @@ const (
 	defaultServiceAccountName  = "prometheus-k8s"
 )
 
+var (
+	readyGauge = prometheus.NewGauge(
+		prometheus.GaugeOpts{
+			Name: "kubevirt_cnao_cr_ready",
+			Help: "Cnao CR Ready",
+		})
+)
+
 func init() {
-	metrics.Registry.MustRegister()
+	metrics.Registry.MustRegister(readyGauge)
+}
+
+func SetReadyGauge(isReady bool) {
+	if isReady {
+		readyGauge.Set(1)
+	} else {
+		readyGauge.Set(0)
+	}
 }
 
 func RenderMonitoring(manifestDir string, monitoringAvailable bool) ([]*unstructured.Unstructured, error) {
--- a/test/check/check.go
+++ b/test/check/check.go
@@ -695,6 +695,50 @@ func checkForPrometheusRuleRemoval(name string) error {
 	return isNotFound("PrometheusRule", name, err)
 }
 
+func getMonitoringEndpoint() (*corev1.Endpoints, error) {
+	By("Finding CNAO prometheus endpoint")
+	endpoint := &corev1.Endpoints{}
+	err := framework.Global.Client.Get(context.Background(), types.NamespacedName{Name: MonitoringComponent.Service, Namespace: components.Namespace}, endpoint)
+	if err != nil {
+		return nil, err
+	}
+	return endpoint, nil
+}
+
+func scrapeEndpointAddress(epAddress corev1.EndpointAddress, epPort int32) (string, error) {
+	stdout, err := Kubectl("exec", "-n", epAddress.TargetRef.Namespace, epAddress.TargetRef.Name, "--", "curl", "-L", "-s", "-k", fmt.Sprintf("http://%s:%d/metrics", epAddress.IP, epPort))
+	if err != nil {
+		return "", err
+	}
+	return stdout, nil
+}
+
+func GetScrapedDataFromMonitoringEndpoint() (string, error) {
+	endpoint, err := getMonitoringEndpoint()
+	if err != nil {
+		return "", err
+	}
+
+	By("scraping the metrics endpoint on CNAO pod")
+	epPort := endpoint.Subsets[0].Ports[0].Port
+	for _, epAddress := range endpoint.Subsets[0].Addresses {
+		if !strings.HasPrefix(epAddress.TargetRef.Name, components.Name) {
+			continue
+		}
+		return scrapeEndpointAddress(epAddress, epPort)
+	}
+	return "", errors.New("no endpoint target ref name matches CNAO component")
+}
+
+func FindMetric(metrics string, expectedMetric string) string {
+	for _, line := range strings.Split(metrics, "\n") {
+		if strings.HasPrefix(line, expectedMetric+" ") {
+			return line
+		}
+	}
+	return ""
+}
+
 func isNotFound(componentType string, componentName string, clientGetOutput error) error {
 	if clientGetOutput != nil {
 		if apierrors.IsNotFound(clientGetOutput) {
--- a/test/e2e/workflow/deployment_test.go
+++ b/test/e2e/workflow/deployment_test.go
@@ -281,6 +281,25 @@ var _ = Describe("NetworkAddonsConfig", func() {
 			CreateConfig(gvk, configSpec)
 			CheckConfigCondition(gvk, ConditionAvailable, ConditionTrue, 15*time.Minute, 30*time.Second)
 		})
+		Context("and checking CNAO prometheus endpoint", func() {
+			var err error
+			var scrapedData string
+			BeforeEach(func() {
+				scrapedData, err = GetScrapedDataFromMonitoringEndpoint()
+				Expect(err).ToNot(HaveOccurred())
+			})
+			It("Should report the expected metrics", func() {
+				expectedMetricValueMap := map[string]string{
+					"kubevirt_cnao_cr_ready": "1",
+				}
+
+				for metricName, expectedValue := range expectedMetricValueMap {
+					metricEntry := FindMetric(scrapedData, metricName)
+					Expect(metricEntry).ToNot(BeEmpty(), fmt.Sprintf("metric %s does not appear in endpoint scrape", metricName))
+					Expect(metricEntry).To(Equal(fmt.Sprintf("%s %s", metricName, expectedValue)), fmt.Sprintf("metric %s does not have the expected value %s", metricName, expectedValue))
+				}
+			})
+		})
 	})
 	//2178
 	Context("when kubeMacPool is deployed", func() {
--- a/go.mod
+++ b/go.mod
@@ -22,6 +22,7 @@ require (
 	github.com/openshift/origin v4.1.0+incompatible
 	github.com/operator-framework/operator-sdk v0.18.0
 	github.com/pkg/errors v0.9.1
+	github.com/prometheus/client_golang v1.5.1
 	github.com/spf13/pflag v1.0.5
 	github.com/thanhpk/randstr v1.0.4
 	github.com/tomnomnom/linkheader v0.0.0-20180905144013-02ca5825eb80 // indirect
--- a/hack/prom-rule-ci/prom-rules-tests.yaml
+++ b/hack/prom-rule-ci/prom-rules-tests.yaml
@@ -35,3 +35,36 @@ tests:
       - eval_time: 5m
         alertname: CnaoDown
         exp_alerts:
+
+# NetworkAddonsConfigNotReady positive tests
+  - interval: 1m
+    input_series:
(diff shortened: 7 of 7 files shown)


RamLavi:
/hold
this PR depends on the monitoring infra PR #928 
relevant commits start at `add ready gauge and alert`

pkg/controller/statusmanager/status_manager.go
```diff
@@ -83,6 +84,7 @@ func (status *StatusManager) Set(reachedAvailableLevel bool, conditions ...condi
 	for i := 0; i < conditionsUpdateRetries; i++ {
 		err := status.set(reachedAvailableLevel, conditions...)
 		if err == nil {
+			monitoring.SetReadyGauge(reachedAvailableLevel)
```
oshoval, line 87: is there somewhere in the code where we need to set `SetReadyGauge(false)` (for example after it turned Ready, or if its going to be undeployed gracefully if there is such scenario) ?

pkg/monitoring/metrics.go
```diff
@@ -1,18 +1,29 @@
...
 	"net/http"
 
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 )
 
+var (
+    readyGauge = prometheus.NewGauge(
+    	prometheus.GaugeOpts{
+    		Name: "kubevirt_cnao_ready",
+    		Help: "Operator Components (Deployed and)? Ready",
```
oshoval, line 15: `?` is intentional ?
little weird

pkg/controller/statusmanager/status_manager.go
RamLavi: not sure what you mean.. My premise is that if statusManager works well to describe the availability of CNAO, then we can just follow that.

data/monitoring/prom-rule.yaml
```diff
@@ -18,3 +18,10 @@ spec:
           for: 5m
           labels:
             severity: critical
+        - alert: CnaoNotReady
+          annotations:
+            summary: one or more of CNAO deployed components is not ready.
+          expr: kubevirt_cnao_num_up_operators == 0
```
oshoval, line 24: same expr as above, maybe you mean kubevirt_cnao_ready ?

pkg/controller/statusmanager/status_manager.go
oshoval: just in the case of err != nil, do we need to update the Ready to 0 ? or to some value representing unknown ?
or just keep the old value

pkg/monitoring/metrics.go
RamLavi: nah, totally missed that. fixed. :)

pkg/controller/statusmanager/status_manager.go
RamLavi: if we failed to set the status, then the last status remained, so I think we should also keep the metric the same.. 

>(are there other cases of tear down where we wont reach here but do want to set it to 0 ?)

not that I can think of.

data/monitoring/prom-rule.yaml
RamLavi: the alert should fire when CNAO is not ready, so the name is fitting imo
but you are right that I used the wrong `expr` there. fixed. thanks!

test/e2e/workflow/deployment_test.go
```diff
@@ -277,6 +281,24 @@ var _ = Describe("NetworkAddonsConfig", func() {
...
+				scrapedData, err = GetScrapedDataFromMonitoringEndpoint()
+				Expect(err).ToNot(HaveOccurred())
+			})
+			It("Should report the expected metrics", func() {
+				expectedmetricValueMap := map[string]string{
+					"kubevirt_cnao_ready": "1",
+				}
+
+				for metricName, expectedValue := range expectedmetricValueMap {
+					Expect(scrapedData).To(ContainSubstring(metricName), fmt.Sprintf("metric %s does not appear in endpoint scrape", metricName))
```
oshoval, line 297: isnt it redundant, as the next line will include `metricName` ?

RamLavi: I agree, changed the test, tell me what you think

oshoval: I think that 2nd line was enough, no need to add the parsing
as the only line that match VAR_NAME VALUE is the one we need (beside *)

`Expect(scrapedData).To(ContainSubstring(fmt.Sprintf("%s %s", metricName, expectedValue)), fmt.Sprintf("metric %s does not have the correct value %s", metricName, expectedValue))`

maybe nice to have is something like
var re = regexp.MustCompile(`^VAR_NAME VALUE$`)
so its the exact line as a whole, this way it won't match a prefix of other variable (*)

wdyt?

RamLavi: scrapedData is huge, I want to avoid `Expect`ing the entire scrapedData string, as the output on the test failure will be too tiresome to debug, and also might be trancated by gomega..

test/e2e/workflow/deployment_test.go
```diff
@@ -281,6 +282,34 @@ var _ = Describe("NetworkAddonsConfig", func() {
...
+
+				for metricName, expectedValue := range expectedmetricValueMap {
+					metricFound := false
+					for _, line := range strings.Split(scrapedData, "\n") {
+						if strings.HasPrefix(line, "#") {
+							continue
+						}
+						if strings.Contains(line, metricName) {
+							metricFound = true
+							Expect(line).To(Equal(fmt.Sprintf("%s %s", metricName, expectedValue)), fmt.Sprintf("metric %s does not have the expected value %s", metricName, expectedValue))
```
oshoval, line 306: you can break here after the Expect

test/e2e/workflow/deployment_test.go
```diff
@@ -281,6 +282,34 @@ var _ = Describe("NetworkAddonsConfig", func() {
 			CreateConfig(gvk, configSpec)
 			CheckConfigCondition(gvk, ConditionAvailable, ConditionTrue, 15*time.Minute, 30*time.Second)
 		})
+		Context("and checking CNAO prometheus endpoint", func() {
+			var err error
+			var scrapedData string
+			BeforeEach(func() {
+				scrapedData, err = GetScrapedDataFromMonitoringEndpoint()
```
oshoval, line 289: Why to scrap it on the before each and not in the test body itself ?

test/e2e/workflow/deployment_test.go
```diff
@@ -281,6 +282,34 @@ var _ = Describe("NetworkAddonsConfig", func() {
...
+				Expect(err).ToNot(HaveOccurred())
+				fmt.Print(scrapedData)
```
oshoval, line 291: leftover?
don't think we want to print it

test/e2e/workflow/deployment_test.go
RamLavi: totally. Done

test/e2e/workflow/deployment_test.go
RamLavi: Because the test is about checking that the metric exists with the expected value on the scrape.
it is not about the scraping itself, and so I dubbed it as something needed to be done BeforeEach test..

RamLavi:
@oshoval fixed [here](https://github.com/kubevirt/cluster-network-addons-operator/compare/174799be91e38f23a025d6f015006b2ed47659b7..e689d1602b9e0b32df21a17bff628a7e6fe65867)

test/e2e/workflow/deployment_test.go
```diff
@@ -281,6 +282,34 @@ var _ = Describe("NetworkAddonsConfig", func() {
...
+					"kubevirt_cnao_ready": "1",
+				}
+
+				for metricName, expectedValue := range expectedmetricValueMap {
+					metricFound := false
+					for _, line := range strings.Split(scrapedData, "\n") {
+						if strings.HasPrefix(line, "#") {
+							continue
+						}
+						if strings.Contains(line, metricName) {
```
oshoval, line 303: its problematic in the rare case where there is metric name which is part of other metric name

test/check/check.go
```diff
@@ -695,6 +695,41 @@ func checkForPrometheusRuleRemoval(name string) error {
...
+	endpoint := &corev1.Endpoints{}
+	err := framework.Global.Client.Get(context.Background(), types.NamespacedName{Name: MonitoringComponent.Service, Namespace: components.Namespace}, endpoint)
+	if err != nil {
+		return nil, err
+	}
+	return endpoint, nil
+}
+
+func scrapeEndpointAddress(epAddress corev1.EndpointAddress, epPort int32) (string, error) {
+	stdout, err := Kubectl("exec", "-n", epAddress.TargetRef.Namespace, epAddress.TargetRef.Name, "--", "curl", "-L", "-k", fmt.Sprintf("http://%s:%d/metrics", epAddress.IP, epPort))
```
oshoval, line 709: worth to add -s to the curl please

data/monitoring/prom-rule.yaml
```diff
@@ -19,3 +19,11 @@ spec:
...
           labels:
             severity: warning
+        - alert: CnaoNotReady
+          annotations:
+            summary: One or more of CNAO deployed components is not ready.
+            runbook_url: http://kubevirt.io/monitoring/runbooks/CnaoNotReady
+          expr: sum(kubevirt_cnao_ready{namespace='{{ .Namespace }}'} or vector(0)) == 0
+          for: 5m
+          labels:
+            severity: warning
```
oshoval, line 29: runbook needed?

sradco: Yes. Its needed.

RamLavi: we have one https://github.com/kubevirt/monitoring/pull/19

RamLavi: it will fail the expect otherwise..

oshoval: This answer belong to https://github.com/kubevirt/cluster-network-addons-operator/pull/944#discussion_r683205618
from what i understand.

In case its allowed that a metric name will be part of other metric name, then it should not Assert, and the code currently need to be updated to support this case.
In case its not allowed, then the code is fine, because it won't happen / find it out in case it happens, (depends on the other the metrics appear in the output)

data/monitoring/prom-rule.yaml
```diff
@@ -19,3 +19,11 @@ spec:
           for: 5m
           labels:
             severity: warning
+        - alert: CnaoNotReady
+          annotations:
+            summary: One or more of CNAO deployed components is not ready.
```
qinqon, line 24: CNAO CR NetworkAddonsConfig not ready

data/monitoring/prom-rule.yaml
```diff
@@ -19,3 +19,11 @@ spec:
           for: 5m
           labels:
             severity: warning
+        - alert: CnaoNotReady
```
qinqon, line 22: Maybe I am late to the discussion but this feals like the CNAO operator is not ready more than the CNAO CR, I think this have ot be NetworkAddonsConfigNotReady

qinqon (changes requested):
Alarm's name CnaoNotReady feels like CNAO operator is not ready not that the NetworkAddonsConfig is not ready.

data/monitoring/prom-rule.yaml
RamLavi: I agree. Done

data/monitoring/prom-rule.yaml
RamLavi: Done

test/e2e/workflow/deployment_test.go
RamLavi: I agree. took your suggestion with slight changes.

data/monitoring/prom-rule.yaml
RamLavi: it could happen, so I changed it accordingly. thanks!

RamLavi:
@oshoval @qinqon fixes [here](https://github.com/kubevirt/cluster-network-addons-operator/compare/e689d1602b9e0b32df21a17bff628a7e6fe65867..d5176f847b2b62ab9f9e0ebad2f6d7679965cf1b)

test/e2e/workflow/deployment_test.go
```diff
@@ -281,6 +281,25 @@ var _ = Describe("NetworkAddonsConfig", func() {
...
+				Expect(err).ToNot(HaveOccurred())
+			})
+			It("Should report the expected metrics", func() {
+				expectedMetricValueMap := map[string]string{
+					"kubevirt_cnao_cr_ready": "1",
+				}
+
+				for metricName, expectedValue := range expectedMetricValueMap {
+					metricEntry := FindMetric(scrapedData, metricName)
+					Expect(metricEntry).ToNot(BeEmpty(), fmt.Sprintf("metric %s does not appear in endpoint scrape", metricName))
```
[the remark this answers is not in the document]
RamLavi: IMO it is more important to make the assertion clear.
1 assertion hitting both fail scenarios is not enough. The test should help the developer debugging why this test fail understand what happened in the best and most unambiguous way. 
Better to split into 2 assertions: 1 for the existence of the metric, and the second is it's value.

test/check/check.go
```diff
@@ -695,6 +695,53 @@ func checkForPrometheusRuleRemoval(name string) error {
...
+			continue
+		}
+		return scrapeEndpointAddress(epAddress, epPort)
+	}
+	return "", errors.New("no endpoint target ref name matches CNAO component")
+}
+
+func FindMetric(metrics string, expectedMetric string) string {
+	for _, line := range strings.Split(metrics, "\n") {
+		if strings.HasPrefix(line, "#") {
```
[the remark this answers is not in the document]
RamLavi: I agree. Done

data/monitoring/prom-rule.yaml
```diff
@@ -19,3 +19,11 @@ spec:
           for: 5m
           labels:
             severity: warning
+        - alert: NetworkAddonsConfigNotReady
+          annotations:
+            summary: CNAO CR NetworkAddonsConfig is not ready.
+            runbook_url: http://kubevirt.io/monitoring/runbooks/NetworkAddonsConfigNotReady
+          expr: sum(kubevirt_cnao_cr_ready{namespace='{{ .Namespace }}'} or vector(0)) == 0
```
oshoval, line 26: is the `or vector(0)` is a must ?

oshoval (approved):
Thanks

test/check/check.go
```diff
@@ -695,6 +695,50 @@ func checkForPrometheusRuleRemoval(name string) error {
...
+	endpoint := &corev1.Endpoints{}
+	err := framework.Global.Client.Get(context.Background(), types.NamespacedName{Name: MonitoringComponent.Service, Namespace: components.Namespace}, endpoint)
+	if err != nil {
+		return nil, err
+	}
+	return endpoint, nil
+}
+
+func scrapeEndpointAddress(epAddress corev1.EndpointAddress, epPort int32) (string, error) {
+	stdout, err := Kubectl("exec", "-n", epAddress.TargetRef.Namespace, epAddress.TargetRef.Name, "--", "curl", "-L", "-s", "-k", fmt.Sprintf("http://%s:%d/metrics", epAddress.IP, epPort))
```
oshoval, line 709: can you check please if curl 127.0.0.1:port works?
if so just the port will need to be fetched, and then maybe ep wont be need to be found
just use the port using the port geter from #928 if its suitable

data/monitoring/prom-rule.yaml
sradco: Yes. It handles cases were the metric is missing (no data) which could happen due to the cr not being deployed.

test/check/check.go
RamLavi: We already use the pod name in `epAddress.TargetRef.Name`, so why not use `epAddress.IP`?
Is there a reason we should't?

oshoval: I mean you totally actually don't need the ep at all
you can just use 127.0.0.1:<PORT> once you `kubectl exec` to the container.

Unless you do want to include the end point as part of the e2e test and then its ok,
basically the ep is not needed here just in order to fetch the metrics, since you are running from within the container.

RamLavi: >Unless you do want to include the end point as part of the e2e test and then its ok

yep that's it.

PR #948, opened 2021-07-28
Fix installation of bridge and tuning CNIs

RamLavi:
This PR is based on #595, and fixed a bug where the binaries are not deployed on the node by linux-bridge pod

- **data/linux-bridge, Change copy command to be fail-able**
Currently if one of the copy commands fail then the pod is
considered not failing.
Fix that by making sure the script fails upon copy failure.
- **bump-linux-bridge, refactor env vars**
Removed unused LINUX_BRIDGE_TAR_FILE
Used currently unused LINUX_BRIDGE_TAR_CONTAINER_DIR
- **data/linux-bridge, Fix installation of bridge and tuning CNIs**
There is a bug where the linux-bridge pods never installed the bridge
and tuning CNI, since the source address of the copy was not matching
directories used in the CNI plugins container image.
The only reason why this worked so far is that these plugins are
deployed by other CNI intallations used by kubevirtci cluster-up.

The desired behavior is:
1. CNAO's linux-bridge pod deploys it's own binaries with the `cnv` prefix
2. if there are no old `bridge`/`tuning`binaries, it will create a symlink to the cnv-prefixed binaries

**Release note**:

```release-note
changed CNAO's linux-bridge deployed bridge&tuning binaries name to `cnv-{bridge,tuning}` to avoid race with other CNIs
```

### DIFF
--- a/data/linux-bridge/002-linux-bridge.yaml
+++ b/data/linux-bridge/002-linux-bridge.yaml
@@ -35,8 +35,9 @@ spec:
             - /bin/bash
             - -c
             - |
-              cp -rf /usr/src/containernetworking/plugins/bin/*bridge /opt/cni/bin/
-              cp -rf /usr/src/containernetworking/plugins/bin/*tuning /opt/cni/bin/
+              echo 'Installing bridge and tuning CNIs'
+              cp -f /usr/src/containernetworking/plugins/bin/*bridge /opt/cni/bin/ || exit 1
+              cp -f /usr/src/containernetworking/plugins/bin/*tuning /opt/cni/bin/ || exit 1
               # Some projects (e.g. openshift/console) use cnv- prefix to distinguish between
               # binaries shipped by OpenShift and those shipped by KubeVirt (D/S matters).
               # Following two lines make sure we will provide both names when needed.
--- a/data/linux-bridge/002-linux-bridge.yaml
+++ b/data/linux-bridge/002-linux-bridge.yaml
@@ -36,14 +36,15 @@ spec:
             - -c
             - |
               echo 'Installing bridge and tuning CNIs'
-              cp -f /usr/src/containernetworking/plugins/bin/*bridge /opt/cni/bin/ || exit 1
-              cp -f /usr/src/containernetworking/plugins/bin/*tuning /opt/cni/bin/ || exit 1
+              cni_mount_dir=/opt/cni/bin
+              cp -f /usr/src/github.com/containernetworking/plugins/bin/bridge ${cni_mount_dir}/cnv-bridge || exit 1
+              cp -f /usr/src/github.com/containernetworking/plugins/bin/tuning ${cni_mount_dir}/cnv-tuning || exit 1
               # Some projects (e.g. openshift/console) use cnv- prefix to distinguish between
               # binaries shipped by OpenShift and those shipped by KubeVirt (D/S matters).
               # Following two lines make sure we will provide both names when needed.
-              find /opt/cni/bin/cnv-bridge || ln -s /opt/cni/bin/bridge /opt/cni/bin/cnv-bridge
-              find /opt/cni/bin/cnv-tuning || ln -s /opt/cni/bin/tuning /opt/cni/bin/cnv-tuning
-              echo "Entering sleep... (success)"
+              find ${cni_mount_dir}/bridge &>/dev/null || ln -s ${cni_mount_dir}/cnv-bridge ${cni_mount_dir}/bridge
+              find ${cni_mount_dir}/tuning &>/dev/null || ln -s ${cni_mount_dir}/cnv-tuning ${cni_mount_dir}/tuning
+              echo 'Entering sleep... (success)'
               sleep infinity
           resources:
             requests:
--- a/hack/components/bump-linux-bridge.sh
+++ b/hack/components/bump-linux-bridge.sh
@@ -23,7 +23,6 @@ LINUX_BRIDGE_TAG=$(git-utils::get_component_tag ${LINUX_BRIDGE_PATH})
 
 echo 'Build container image with linux-bridge binaries'
 LINUX_BRIDGE_TAR_CONTAINER_DIR=/usr/src/github.com/containernetworking/plugins/bin
-LINUX_BRIDGE_TAR_FILE=cni-plugins-linux-amd64-${LINUX_BRIDGE_TAG}.tgz
 LINUX_BRIDGE_IMAGE=quay.io/kubevirt/cni-default-plugins
 LINUX_BRIDGE_IMAGE_TAGGED=${LINUX_BRIDGE_IMAGE}:${LINUX_BRIDGE_TAG}
 (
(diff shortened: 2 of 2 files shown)


RamLavi:
/hold

I still don't understand how the `bridge` and `tuning` binaries got to the node's `/opt/cni/bin/` directory.. I can see it even after `make cluster-up`.
Let's further investigate and fix the commit description accordingly

data/linux-bridge/002-linux-bridge.yaml
```diff
@@ -41,9 +41,9 @@ spec:
               # Some projects (e.g. openshift/console) use cnv- prefix to distinguish between
               # binaries shipped by OpenShift and those shipped by KubeVirt (D/S matters).
               # Following two lines make sure we will provide both names when needed.
-              find /opt/cni/bin/cnv-bridge || ln -s /opt/cni/bin/bridge /opt/cni/bin/cnv-bridge
-              find /opt/cni/bin/cnv-tuning || ln -s /opt/cni/bin/tuning /opt/cni/bin/cnv-tuning
-              echo "Entering sleep... (success)"
+              find /opt/cni/bin/cnv-bridge &>/dev/null || ln -s /opt/cni/bin/bridge {{ .CNIBinDir }}/cnv-bridge
```
EdDev, line 44: I do not understand this one. You left what you look for as static to the k8s default place, same for the target of the soft-link.

RamLavi:
> /hold
> 
> I still don't understand how the `bridge` and `tuning` binaries got to the node's `/opt/cni/bin/` directory.. I can see it even after `make cluster-up`.
> Let's further investigate and fix the commit description accordingly

I suspect the binaries are copied by istio [here](https://github.com/kubevirt/kubevirtci/blob/main/cluster-provision/k8s/1.21/manifests/cni.do-not-change.yaml#L3562)
To avoid a race, added another commit to change our binaries

the big Con is that we loose backwards compatibility on U/S..

rhrazdil:
> I suspect the binaries are copied by istio here

Ah you mean calico right? I got worried for a second here :)

RamLavi:
> > I suspect the binaries are copied by istio here
> 
> Ah you mean calico right? I got worried for a second here :)

ah yeah sorry got mixed up

hack/components/bump-linux-bridge.sh
```diff
@@ -22,7 +22,7 @@ echo 'Get linux-bridge tag'
 LINUX_BRIDGE_TAG=$(git-utils::get_component_tag ${LINUX_BRIDGE_PATH})
 
 echo 'Build container image with linux-bridge binaries'
-LINUX_BRIDGE_TAR_CONTAINER_DIR=/usr/src/github.com/containernetworking/plugins/bin
```
RamLavi, line 25: note that this change will make the PR fail until @phoracek runs `PUSH_IMAGES=true make bump-linux-bridge`

data/linux-bridge/002-linux-bridge.yaml
```diff
@@ -35,14 +35,15 @@ spec:
             - /bin/bash
             - -c
             - |
-              cp -rf /usr/src/containernetworking/plugins/bin/*bridge /opt/cni/bin/
-              cp -rf /usr/src/containernetworking/plugins/bin/*tuning /opt/cni/bin/
+              echo 'Installing bridge and tuning CNIs'
+              cp -f /usr/src/containernetworking/plugins/bin/bridge {{ .CNIBinDir }}/cnv-bridge || exit 1
```
phoracek, line 39: What is the point of the exit? Would not it happen anyway?

data/linux-bridge/002-linux-bridge.yaml
```diff
@@ -52,7 +53,7 @@ spec:
             privileged: true
           volumeMounts:
             - name: cnibin
-              mountPath: /opt/cni/bin
+              mountPath: {{ .CNIBinDir }}
       volumes:
```
phoracek, line 57: this is not needed, the volume serves as an abstraction. all scripts can be static the only place for tempting is from where we mount this directory

data/linux-bridge/002-linux-bridge.yaml
EdDev: No, it will not. This is not in a bash script with -e set.

data/linux-bridge/002-linux-bridge.yaml
```diff
@@ -36,14 +36,14 @@ spec:
...
-              cp -f /usr/src/containernetworking/plugins/bin/*tuning /opt/cni/bin/ || exit 1
+              cp -f /usr/src/containernetworking/plugins/bin/bridge {{ .CNIBinDir }}/cnv-bridge || exit 1
+              cp -f /usr/src/containernetworking/plugins/bin/tuning {{ .CNIBinDir }}/cnv-tuning || exit 1
               # Some projects (e.g. openshift/console) use cnv- prefix to distinguish between
               # binaries shipped by OpenShift and those shipped by KubeVirt (D/S matters).
               # Following two lines make sure we will provide both names when needed.
-              find /opt/cni/bin/cnv-bridge || ln -s /opt/cni/bin/bridge /opt/cni/bin/cnv-bridge
-              find /opt/cni/bin/cnv-tuning || ln -s /opt/cni/bin/tuning /opt/cni/bin/cnv-tuning
-              echo "Entering sleep... (success)"
+              find {{ .CNIBinDir }}/bridge &>/dev/null || ln -s {{ .CNIBinDir }}/cnv-bridge {{ .CNIBinDir }}/bridge
```
EdDev, line 44: Using `find` instead of testing for a file to exist is really odd. But as it was used before, I guess it is not a must to change it now.

EdDev (commented):
commit: data/linux-bridge, Fix symlink and avoid race

Please elaborate on the naming changes and soft-links. We need to be clear on the behavior and expected input from the users now that this change is taking place.

This is not just an internal fix, it has user facing implications.

data/linux-bridge/002-linux-bridge.yaml
phoracek:
```
The shell does not exit if ... part of any command executed in a && or ||  list  except  the  command
                      following  the final && or ||
```

Today I learned. Ack

hack/components/bump-linux-bridge.sh
phoracek: I don't follow, why do we need to change this?

RamLavi: we don't really :) removed

data/linux-bridge/002-linux-bridge.yaml
RamLavi: yeah, we can always make improvements in another PR

data/linux-bridge/002-linux-bridge.yaml
RamLavi: I think I fixed it. tell me if not :)

RamLavi:
desired behavior reached: after `make cluster-sync+apply CR`
```
[vagrant@node01 ~]$ ls -ltr /opt/cni/bin/
-rwxr-xr-x. 1 root root  3581192 Sep  9  2020 bridge
-rwxr-xr-x. 1 root root  3356587 Jul 29 08:37 tuning
-rwxr-xr-x. 1 root root  4891616 Jul 29 13:52 cnv-bridge
-rwxr-xr-x. 1 root root  3836648 Jul 29 13:52 cnv-tuning
```
in this case, pre-deployed `bridge` and `tuning` binaries are untouched, `cnv-bridge` `cnv-tuning` and are created

when I delete them and re-`make cluster-sync+apply CR` then I get :
```
[vagrant@node01 ~]$ ls -ltr /opt/cni/bin/
-rwxr-xr-x. 1 root root  4891616 Jul 29 14:03 cnv-bridge
-rwxr-xr-x. 1 root root  3836648 Jul 29 14:03 cnv-tuning
lrwxrwxrwx. 1 root root       23 Jul 29 14:03 bridge -> /opt/cni/bin/cnv-bridge
lrwxrwxrwx. 1 root root       23 Jul 29 14:03 tuning -> /opt/cni/bin/cnv-tuning
```

data/linux-bridge/002-linux-bridge.yaml
EdDev: I do not understand @phoracek here.
It sounded to me like he was talking about having the scripts as static and the source of the mount to be templated. But I am unsure really. To me it sounds like we need all to be templates or use a variable, in order not to mess things up.

RamLavi: I think I made a mess, but fixed it now..

EdDev (approved):
Thanks!

There is still a potential future breakage because the script that provision the binaries in the container and the copy of them to the node have no shared data. This may be ok as long as we protect it using tests (validating the right file reached the target).

phoracek (approved):
/lgtm
/approve
/hold

Very nice, thanks! Feel free to cancel the hold whenever it's ready from your side.

PR #950, opened 2021-08-03
data/linux-bridge, Fix copying over symlink

RamLavi:
When we upgrade there is already the cnv-{bridge,tuning}
files in the node. In this case, 'cp -f' does not copy
the new binaries over the old symlinks.
To Fix that, changed the flag to '--remove-destination'
```
      --remove-destination     remove each existing destination file before
                                 attempting to open it (contrast with --force)
```

**Release note**:

```release-note
fix symlink in linux-bridge during upgrade
```


EdDev (approved):
Verified locally and it works.

PR #955, opened 2021-08-09
monitoring, Adding a new alert KubemacpoolDown

RamLavi:
This PR Adds a new alert KubemacpoolDown
This alert will fire only when KMP is configured and for some reason the kubemacool-manager pod is not running.
The alert priority is set to critical, as if KMP is down, vms cannot be created/updated from the cluster.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/data/monitoring/prom-rule.yaml
+++ b/data/monitoring/prom-rule.yaml
@@ -37,3 +37,15 @@ spec:
           for: 5m
           labels:
             severity: warning
+        - expr: sum(up{namespace='{{ .Namespace }}', pod=~'kubemacpool-mac-controller-manager-.*'} or vector(0))
+          record: kubevirt_cnao_kubemacpool_manager_num_up_pods_total
+        - expr: sum(kubevirt_cnao_cr_kubemacpool_deployed{namespace='{{ .Namespace }}'} or vector(0))
+          record: kubevirt_cnao_cr_kubemacpool_deployed_total
+        - alert: KubemacpoolDown
+          annotations:
+            summary: Kubemacpool is deployed by CNAO CR but kubemacpool pod is down.
+            runbook_url: http://kubevirt.io/monitoring/runbooks/KubemacpoolDown
+          expr: kubevirt_cnao_cr_kubemacpool_deployed_total == 1 and kubevirt_cnao_kubemacpool_manager_num_up_pods_total == 0
+          for: 5m
+          labels:
+            severity: critical
--- a/pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
+++ b/pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
@@ -39,6 +39,7 @@ import (
 	"github.com/kubevirt/cluster-network-addons-operator/pkg/apply"
 	"github.com/kubevirt/cluster-network-addons-operator/pkg/controller/statusmanager"
 	"github.com/kubevirt/cluster-network-addons-operator/pkg/eventemitter"
+	"github.com/kubevirt/cluster-network-addons-operator/pkg/monitoring"
 	"github.com/kubevirt/cluster-network-addons-operator/pkg/names"
 	"github.com/kubevirt/cluster-network-addons-operator/pkg/network"
 	"github.com/kubevirt/cluster-network-addons-operator/pkg/util/k8s"
@@ -220,6 +221,10 @@ func (r *ReconcileNetworkAddonsConfig) Reconcile(request reconcile.Request) (rec
 			// TODO: This can be dropped once we implement a finalizer waiting for all components to be removed
 			r.stopTrackingObjects()
 
+			if r.clusterInfo.MonitoringAvailable {
+				monitoring.ResetMonitoredComponents()
+			}
+
 			// Owned objects are automatically garbage collected. Return and don't requeue
 			return reconcile.Result{}, nil
 		}
@@ -295,6 +300,10 @@ func (r *ReconcileNetworkAddonsConfig) Reconcile(request reconcile.Request) (rec
 		// If failed, set NetworkAddonsConfig to failing and requeue
 		r.statusManager.SetFailing(statusmanager.OperatorConfig, "FailedToDeleteObjects", err.Error())
 		return reconcile.Result{}, err
+	}
+
+	if r.clusterInfo.MonitoringAvailable {
+		monitoring.TrackMonitoredComponents(&networkAddonsConfig.Spec)
 	}
 
 	// Everything went smooth, remove failures from NetworkAddonsConfig if there are any from
--- a/pkg/monitoring/metrics.go
+++ b/pkg/monitoring/metrics.go
@@ -11,6 +11,7 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"sigs.k8s.io/controller-runtime/pkg/metrics"
 
+	cnao "github.com/kubevirt/cluster-network-addons-operator/pkg/apis/networkaddonsoperator/shared"
 	"github.com/kubevirt/cluster-network-addons-operator/pkg/render"
 )
 
@@ -26,10 +27,15 @@ var (
 			Name: "kubevirt_cnao_cr_ready",
 			Help: "Cnao CR Ready",
 		})
+	kmpDeployedGauge = prometheus.NewGauge(
+		prometheus.GaugeOpts{
+			Name: "kubevirt_cnao_cr_kubemacpool_deployed",
+			Help: "Kubemacpool is deployed by Cnao CR",
+		})
 )
 
 func init() {
-	metrics.Registry.MustRegister(readyGauge)
+	metrics.Registry.MustRegister(readyGauge, kmpDeployedGauge)
 }
 
 func SetReadyGauge(isReady bool) {
@@ -38,6 +44,24 @@ func SetReadyGauge(isReady bool) {
 	} else {
 		readyGauge.Set(0)
 	}
+}
+
+func setKubemacpoolDeployedGauge(isDeployed bool) {
+	if isDeployed {
+		kmpDeployedGauge.Set(1)
+	} else {
+		kmpDeployedGauge.Set(0)
+	}
+}
+
+func ResetMonitoredComponents() {
+	SetReadyGauge(false)
+	setKubemacpoolDeployedGauge(false)
+}
+
+func TrackMonitoredComponents(conf *cnao.NetworkAddonsConfigSpec) {
+	isKubemacpoolDeployed := conf.KubeMacPool != nil
+	setKubemacpoolDeployedGauge(isKubemacpoolDeployed)
 }
 
 func RenderMonitoring(manifestDir string, monitoringAvailable bool) ([]*unstructured.Unstructured, error) {
--- a/test/e2e/workflow/deployment_test.go
+++ b/test/e2e/workflow/deployment_test.go
@@ -281,25 +281,57 @@ var _ = Describe("NetworkAddonsConfig", func() {
 			CreateConfig(gvk, configSpec)
 			CheckConfigCondition(gvk, ConditionAvailable, ConditionTrue, 15*time.Minute, 30*time.Second)
 		})
-		Context("and checking CNAO prometheus endpoint", func() {
-			var err error
-			var scrapedData string
-			BeforeEach(func() {
-				scrapedData, err = GetScrapedDataFromMonitoringEndpoint()
-				Expect(err).ToNot(HaveOccurred())
-			})
-			It("Should report the expected metrics", func() {
-				expectedMetricValueMap := map[string]string{
-					"kubevirt_cnao_cr_ready": "1",
-				}
+	})
+	Context("when deploying components and checking CNAO prometheus endpoint", func() {
+		type prometheusScrapeParams struct {
+			configSpec             cnao.NetworkAddonsConfigSpec
+			expectedMetricValueMap map[string]string
+		}
+		DescribeTable("and checking scraped data",
+			func(p prometheusScrapeParams) {
+				By("deploying the configured NetworkAddonsConfigSpec")
+				CreateConfig(gvk, p.configSpec)
+				CheckConfigCondition(gvk, ConditionAvailable, ConditionTrue, 15*time.Minute, CheckDoNotRepeat)
 
-				for metricName, expectedValue := range expectedMetricValueMap {
+				By("scraping the monitoring endpoint")
+				scrapedData, err := GetScrapedDataFromMonitoringEndpoint()
+				Expect(err).ToNot(HaveOccurred())
+
+				By("comparing the scraped Data to the expected metrics' values")
+				for metricName, expectedValue := range p.expectedMetricValueMap {
 					metricEntry := FindMetric(scrapedData, metricName)
 					Expect(metricEntry).ToNot(BeEmpty(), fmt.Sprintf("metric %s does not appear in endpoint scrape", metricName))
 					Expect(metricEntry).To(Equal(fmt.Sprintf("%s %s", metricName, expectedValue)), fmt.Sprintf("metric %s does not have the expected value %s", metricName, expectedValue))
 				}
-			})
-		})
+			},
+			Entry("should report the expected metrics when deploying all components", prometheusScrapeParams{
+				configSpec: cnao.NetworkAddonsConfigSpec{
+					LinuxBridge: &cnao.LinuxBridge{},
+					Multus:      &cnao.Multus{},
+					NMState:     &cnao.NMState{},
+					KubeMacPool: &cnao.KubeMacPool{},
+					Ovs:         &cnao.Ovs{},
+					MacvtapCni:  &cnao.MacvtapCni{},
+				},
+				expectedMetricValueMap: map[string]string{
+					"kubevirt_cnao_cr_ready":                "1",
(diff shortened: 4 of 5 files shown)


RamLavi:
open issues:
- seems to be missing a service in order for the endpoint to be seen by the GUI
- unit tests
- e2e negative cases
- when removing CR the value stays 1. need to figure out how to zero this.

RamLavi:
manually checked on CRC

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
```diff
@@ -286,6 +287,8 @@ func (r *ReconcileNetworkAddonsConfig) Reconcile(request reconcile.Request) (rec
 		return reconcile.Result{}, err
 	}
 
+	monitoring.TrackMonitoringComponents(&networkAddonsConfig.Spec, r.clusterInfo.MonitoringAvailable)
```
oshoval, line 290: maybe should be after `deleteObjects` ? (actually same question is about `trackDeployedObjects` but not part of this PR)

https://github.com/kubevirt/cluster-network-addons-operator/pull/955/files#diff-463a5613afbb95588d747500e304c2d348c7fce8ec35117067322e7aac36a8dcR296

pkg/monitoring/metrics.go
```diff
@@ -26,10 +27,15 @@ var (
 			Name: "kubevirt_cnao_cr_ready",
 			Help: "Cnao CR Ready",
 		})
+	kmpDeployedGauge = prometheus.NewGauge(
```
oshoval, line 30: kmp prefix is needed ? not sure

pkg/monitoring/metrics.go
```diff
@@ -26,10 +27,15 @@ var (
...
+		prometheus.GaugeOpts{
+			Name: "kubevirt_cnao_cr_kubemacpool_deployed",
+			Help: "Kubemacpool is deployed by Cnao CR",
+		})
 )
 
 func init() {
-	metrics.Registry.MustRegister(readyGauge)
+	metrics.Registry.MustRegister(readyGauge, kmpDeployedGauge)
```
oshoval, line 38: maybe worth that all the gauges will be part of some slice and here we can register with `...` ?
just nice to have

pkg/monitoring/metrics.go
```diff
@@ -40,6 +46,36 @@ func SetReadyGauge(isReady bool) {
...
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func trackKubemacpoolMonitoring(conf *cnao.NetworkAddonsConfigSpec) error {
+	isKubemacpoolDeployed := false
+	if conf.KubeMacPool != nil {
```
oshoval, line 72: conf.KubeMacPool means its actually deployed ?
or just that it appeared on the manifest, and might be broken / just now reconcile started to deploy it ?

pkg/monitoring/metrics.go
```diff
@@ -40,6 +46,36 @@ func SetReadyGauge(isReady bool) {
...
+	if !monitoringAvailable {
+		return nil
+	}
+
+	err := trackKubemacpoolMonitoring(conf)
+	if err != nil {
+		return err
+	}
+
+	return nil
```
oshoval, line 67: return trackKubemacpoolMonitoring(conf)

oshoval (commented):
First pass, thanks

Can we add some info please about the alert on the PR description ?

pkg/monitoring/metrics.go
RamLavi: what do you mean?

test/e2e/workflow/deployment_test.go
```diff
@@ -281,25 +281,57 @@ var _ = Describe("NetworkAddonsConfig", func() {
...
-			BeforeEach(func() {
-				scrapedData, err = GetScrapedDataFromMonitoringEndpoint()
+	})
+	Context("when deploying components and checking CNAO prometheus endpoint", func() {
+		type prometheusScrapeParams struct {
+			configSpec             cnao.NetworkAddonsConfigSpec
+			expectedMetricValueMap map[string]string
+		}
+		DescribeTable("and checking scraped data",
+			func(p prometheusScrapeParams) {
```
oshoval, line 291: can the func get 2 parameters instead of the struct ?
i don't see a reason to encapsulate those parameters into a struct

pkg/monitoring/metrics.go
oshoval: i confused something, nevermind, its fine

pkg/monitoring/metrics.go
RamLavi: This is intentional, I want this function to be a placeholder for future alerts.. similar to [Validate()](https://github.com/kubevirt/cluster-network-addons-operator/blob/main/pkg/network/network.go#L27)

perhaps I should do the `errs = append(...)` as well, what do you think?

test/e2e/workflow/deployment_test.go
RamLavi: I prefer it like this. 
helps keep parameter names understood , and makes thing neater imo.

data/monitoring/prom-rule.yaml
```diff
@@ -37,3 +37,15 @@ spec:
...
             severity: warning
+        - expr: sum(up{namespace='{{ .Namespace }}', pod=~'kubemacpool-mac-controller-manager-.*'} or vector(0))
+          record: kubevirt_cnao_kubemacpool_manager_num_up_pods_total
+        - expr: sum(kubevirt_cnao_cr_kubemacpool_deployed{namespace='{{ .Namespace }}'} or vector(0))
+          record: kubevirt_cnao_cr_kubemacpool_deployed_total
+        - alert: KubemacpoolDown
+          annotations:
+            summary: Kubemacpool is deployed by CNAO CR but kubemacpool pod is down.
+            runbook_url: http://kubevirt.io/monitoring/runbooks/KubemacpoolDown
+          expr: kubevirt_cnao_cr_kubemacpool_deployed_total == 1 and kubevirt_cnao_kubemacpool_manager_num_up_pods_total == 0
```
oshoval, line 48: maybe due to a bug on CNAO, or just missing kmp on the CR
kubevirt_cnao_cr_kubemacpool_deployed / kubevirt_cnao_cr_kubemacpool_deployed_total will be zero, and then the alert won't raise, it might be fine, but just raising it to be sure

pkg/monitoring/metrics.go
```diff
@@ -40,6 +46,36 @@ func SetReadyGauge(isReady bool) {
...
+}
+
+func trackKubemacpoolMonitoring(conf *cnao.NetworkAddonsConfigSpec) error {
+	isKubemacpoolDeployed := false
+	if conf.KubeMacPool != nil {
+		isKubemacpoolDeployed = true
+	}
+	setKubemacpoolDeployedGauge(isKubemacpoolDeployed)
+	return nil
```
oshoval, line 76: not needed

pkg/monitoring/metrics.go
```diff
@@ -40,6 +46,36 @@ func SetReadyGauge(isReady bool) {
...
+
+	return nil
+}
+
+func trackKubemacpoolMonitoring(conf *cnao.NetworkAddonsConfigSpec) error {
+	isKubemacpoolDeployed := false
+	if conf.KubeMacPool != nil {
+		isKubemacpoolDeployed = true
+	}
+	setKubemacpoolDeployedGauge(isKubemacpoolDeployed)
```
oshoval, line 75: if you like the style, the whole function can be one line 
setKubemacpoolDeployedGauge(conf.KubeMacPool != nil)

pkg/monitoring/metrics.go
oshoval: ye maybe then worth to do the append, else its pretty creating infra for air-conditioner :)

data/monitoring/prom-rule.yaml
RamLavi: the logic is that if KMP is not in the CR then KMP should *not* be deployed at all, hence the alert shouldn't fire (i.e. KMP is down but it's ok, and not something the sysAdmin should wake up in the middle of the night for).

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
RamLavi: I agree it's a more fitting place.. done

pkg/monitoring/metrics.go
RamLavi: I prefer it this way. it's more explicit.

pkg/monitoring/metrics.go
RamLavi: Done

pkg/monitoring/metrics.go
RamLavi: I prefer the variable with the long name `isKubemacpoolDeployed` to make the develper's life easier to understand..

pkg/monitoring/metrics.go
RamLavi: not sure what you mean here.. the function will return []error in general just so we have the option to return errors in the future.

RamLavi:
@oshoval [fixed](https://github.com/kubevirt/cluster-network-addons-operator/compare/ad9259124901fbb570b3ad6d0ad227269e78b04e..85a2353e3eb7a322a82e1071024c3df33c50602f) and addressed comments

pkg/monitoring/metrics.go
oshoval: i don't think we need to create infra for something that is not used right now, its a simple change that if someone need he can add a return value
atm, there is no other returns beside this one

RamLavi:
> First pass, thanks
> 
> Can we add some info please about the alert on the PR description ?

added description in the PR

pkg/monitoring/metrics.go
RamLavi: I disagree. Returning `[]error/error` from functions is generally a good practice. why not here?

oshoval: The function has only one return statement and its always return nil
there is no good practice at all of returning the same thing no matter what the function is doing
i bet there are even static analyse tools that will warn / fail / optimize it

imo there should not be redundant code

oshoval:
> This alert will fire only when KMP is configured and from some reason the kubemacool-manager pod is not running.

s/from/for/

oshoval (approved):
Thanks

pkg/monitoring/metrics.go
RamLavi: I tend to prefer adding the infra beforehand.
`TrackMonitoringComponents()` is the house for handling monitoring for all deployed components.
Each can potentially return an error, so `TrackMonitoringComponents()` collects the errors from them and flattens them to one error line if needed. 
Granted, currently there is only one: `trackKubemacpoolMonitoring()`, but I prefer that we lay out how we want to function to look like from the beginning, so that next time we want to add another monitoring (for knmstate for example) then that developer would only need to add the component's function.

We did the same when we added the `SpecialCleanUp()` [here](https://github.com/kubevirt/cluster-network-addons-operator/pull/270), and I think it's what we should do here as well.

RamLavi:
Added a function `ResetMonitoringComponents()` to reset the gouges when the CR is removed [here](https://github.com/kubevirt/cluster-network-addons-operator/compare/85a2353e3eb7a322a82e1071024c3df33c50602f..3516a4c9151adc5698758b7bfefa2c319483dad7)

pkg/monitoring/metrics.go
```diff
@@ -40,6 +47,42 @@ func SetReadyGauge(isReady bool) {
...
+		kmpDeployedGauge.Set(1)
+	} else {
+		kmpDeployedGauge.Set(0)
+	}
+}
+
+func ResetMonitoringComponents() error {
+	SetReadyGauge(false)
+	setKubemacpoolDeployedGauge(false)
+	return nil
```
oshoval, line 61: same deal here, i think return nil for a function that doesn't do anything else is totally unneeded,
its ok that a function won't return a value

maybe worth that the SetReadyGauge will be on other PR / commit, but i won't insist

oshoval (approved):

pkg/monitoring/metrics.go
RamLavi: in this case I tend to agree there is no need for error in the function. will remove

oshoval (approved):

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
```diff
@@ -297,6 +300,8 @@ func (r *ReconcileNetworkAddonsConfig) Reconcile(request reconcile.Request) (rec
...
```
phoracek, line 303: The condition MonitoringAvailable should be checked here. If it is not available, we should not invoke the module at all.  That way we encapsulate the helper module and isolate it from the dependencies. Would that make sense?

data/monitoring/prom-rule.yaml
```diff
@@ -37,3 +37,15 @@ spec:
           for: 5m
           labels:
             severity: warning
+        - expr: sum(up{namespace='{{ .Namespace }}', pod=~'kubemacpool-mac-controller-manager-.*'} or vector(0))
+          record: kubevirt_cnao_kubemacpool_manager_num_up_pods_total
+        - expr: sum(kubevirt_cnao_cr_kubemacpool_deployed{namespace='{{ .Namespace }}'} or vector(0))
```
phoracek, line 42: Why do we need the sum here? Can't we just remove true or false from kubevirt_cnao_cr_kubemacpool_deployed ? We do not care about the number of deployed objects here, do we? Only about whether it is enabled or not

pkg/monitoring/metrics.go
```diff
@@ -40,6 +47,41 @@ func SetReadyGauge(isReady bool) {
...
 
+func setKubemacpoolDeployedGauge(isDeployed bool) {
+	if isDeployed {
+		kmpDeployedGauge.Set(1)
+	} else {
+		kmpDeployedGauge.Set(0)
+	}
+}
+
+func ResetMonitoringComponents() {
```
phoracek, line 58: Are not these monitored components?

pkg/monitoring/metrics.go
```diff
@@ -40,6 +47,41 @@ func SetReadyGauge(isReady bool) {
...
+	errs = append(errs, trackKubemacpoolMonitoring(conf)...)
+
+	if len(errs) > 0 {
+		return errors.Errorf("Error in tracking monitoring components:\n%s", errorListToMultiLineString(errs))
+	}
+	return nil
+}
+
+func trackKubemacpoolMonitoring(conf *cnao.NetworkAddonsConfigSpec) []error {
+	isKubemacpoolDeployed := false
```
phoracek, line 77: Would this work?

```go
isKubemacpoolDeployed := conf.KubeMacPool != nil && conf.KubeMacPool
```

pkg/monitoring/metrics.go
phoracek: I am concerned about this hidden side effect. Could we either somehow lift this to an upper-level function or to rename this function to clearly show that it is changing the global? Maybe just use the setKubemacpoolDeployedGauge name directly? And if you could pass reference to the global to the function, it would get rid of the side effect completely

phoracek (changes requested):

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
RamLavi: I don't mind, but note we do it differently on render functions (see [here](https://github.com/kubevirt/cluster-network-addons-operator/blob/main/pkg/network/kubemacpool.go#L62-L64))
Do you think we should change it there as well? (in a different PR of course)

data/monitoring/prom-rule.yaml
RamLavi: the alert is using these 2 records `if(x==1 && y==0) for >5min then fire!` 
In order for the alert to work, the records needs to point to the same entry (x counts apples (arbitrary gauge) and y counts bananas (num of up kmp pods) )
After some playing with this with @sradco, we saw that `sum()` normalizes the records to be of same in "type" and so it works.

pkg/monitoring/metrics.go
RamLavi: They are! Thank you!!! 
finding the right wording brings me joy =)

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
RamLavi: Done

pkg/monitoring/metrics.go
RamLavi: Done

pkg/monitoring/metrics.go
RamLavi: like! Done

pkg/monitoring/metrics.go
RamLavi: I see, I'll move it upward, but not sure why I should pass the global. can you elaborate?

phoracek: If you pass link to the global, you can unit test it. It would also become clear what is being "set". With it being taken from the global context, we cannot easily unit test it. It is unclear what this function works with unless you peek into it.

phoracek: Sounds good

phoracek (approved):
/lgtm
/approve

PR #973, opened 2021-08-22
Determine webhook replicas based on infra

alonSadan:
When running on a singleReplica openshift cluster, CNAO should run
a single replica of each of It's operands. This PR  Bumps the necessary API to check 
the infrastructure topology of the cluster, applies that for nmstate webhook which currently runs two replicas,
and test the expected number of replicas for each of the component's deployments.

**Release note**:

```release-note

```


data/nmstate/operand/operator.yaml
```diff
@@ -9,7 +9,7 @@ metadata:
     app: kubernetes-nmstate
     component: kubernetes-nmstate-webhook
 spec:
-  replicas: 2
+  replicas: {{ .WebhookReplicas }}
```
phoracek, line 12: This file is generated through `make bump-nmstate`, I believe this change would be overwritten by that.

We fetch manifest templates from https://github.com/nmstate/kubernetes-nmstate/blob/main/deploy/handler/operator.yaml and replace variables set there with ours when needed (when the original name is not already fine).

Would you please contribute a PR to kubernetes-nmstate extending the template there? It does not need to be dynamic there (that will come in a follow-up PR) - adding the variable there and setting it always to 2 would be good enough.

phoracek (changes requested):

data/nmstate/operand/operator.yaml
alonSadan: ahh 
Thank you 
I will open a PR there.

alonSadan: [Done.](https://github.com/nmstate/kubernetes-nmstate/pull/810)

RamLavi (changes requested):
I think we shouldn't pass the client all the way to nmstate

singleReplica is a property of the cluster (right?), so you can sample it when CNAO and save it to `clusterInfo *ClusterInfo` when the CR is created around [here](https://github.com/kubevirt/cluster-network-addons-operator/blob/main/pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go#L78)..
In the future, should we need this for other components we will have it ready.

alonSadan:
Now we don't fail on error to get the infrastructure, and instead just log the failure and defaults to false. 
https://github.com/kubevirt/cluster-network-addons-operator/compare/050ed28e3e684d4755240efcd0c40a3cde1167f9..cb24b90aec2835b5ec0761707870e9c34590302d#diff-463a5613afbb95588d747500e304c2d348c7fce8ec35117067322e7aac36a8dcR217

also I rebased on master
https://github.com/kubevirt/cluster-network-addons-operator/compare/050ed28e3e684d4755240efcd0c40a3cde1167f9..cb24b90aec2835b5ec0761707870e9c34590302d

alonSadan:
[rebased on master](https://github.com/kubevirt/cluster-network-addons-operator/compare/cb24b90aec2835b5ec0761707870e9c34590302d..3cd904ec95baa290751ca57999a0c26404a37738)

alonSadan:
the unit tests failed on: 

```
 spec:
-  minAvailable: {{ .WebhookMinReplicas }}
+  minAvailable: {{  .WebhookMinReplicas  }}
   selector:
```
[I fixed that.](https://github.com/kubevirt/cluster-network-addons-operator/compare/39d261034f99c0950a93f9b625a7d9323471c0e6..e5b34b2389a0c57656d883877c2ee7b89e35d2be)

alonSadan:
I removed the last commit which adds a test for this feature, since CI never runs on OKD cluster, we will not reach this test ever so there is no point of adding it.
https://github.com/kubevirt/cluster-network-addons-operator/compare/e5b34b2389a0c57656d883877c2ee7b89e35d2be..146b47d768de2fdba8721a90beabd7ad27aca546

RamLavi (approved):
/lgtm

alonSadan:
@phoracek  Can you take a look as well?

phoracek:
@alonSadan I won't get to it this week, feel free to move forward

PR #978, opened 2021-08-23
reconcile, Filter daemon-sets and deployments to controlled namespace

oshoval:
In order to save compute/network resources, filter the watch
CNAO does on the deployments and daemonset to watch only
CNAO operator namespace.

```release-note
None
```

### DIFF
--- a/pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
+++ b/pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
@@ -131,7 +131,7 @@ func add(mgr manager.Manager, r *ReconcileNetworkAddonsConfig) error {
 	// Create custom predicate for NetworkAddonsConfig watcher. This makes sure that Status field
 	// updates will not trigger reconciling of the object. Reconciliation is trigger only if
 	// Spec fields differ.
-	pred := predicate.Funcs{
+	crPred := predicate.Funcs{
 		UpdateFunc: func(e event.UpdateEvent) bool {
 			oldConfig, err := runtimeObjectToNetworkAddonsConfig(e.ObjectOld)
 			if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
```diff
@@ -131,7 +131,7 @@ func add(mgr manager.Manager, r *ReconcileNetworkAddonsConfig) error {
 	// Create custom predicate for NetworkAddonsConfig watcher. This makes sure that Status field
 	// updates will not trigger reconciling of the object. Reconciliation is trigger only if
 	// Spec fields differ.
-	pred := predicate.Funcs{
+	crPred := predicate.Funcs{
```
phoracek, line 134: Predicates are filtering on the client level, we are still getting all the Pods streamed from the API server.

We should either register the cache to only watch a specific namespace using https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/manager/manager.go#L188. But with this I'm not sure whether we can limit only part of the operator.

If the option above is not possible, we may want to use new Quique's option to filter on cache: https://github.com/kubernetes-sigs/controller-runtime/pull/1435

phoracek (changes requested):

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
oshoval: btw we are not getting pods i think, we watch only for deployments and daemonsets
but i see that its done on client side

oshoval: Tried the first option https://github.com/kubevirt/cluster-network-addons-operator/pull/981
the only tests that are failing are the ones that deploy nmstate on different namespace because it can't find them,
its a no go then ?
or its safe to require that no other component is installed ?
https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_cluster-network-addons-operator/981/pull-e2e-cluster-network-addons-operator-workflow-k8s/1429820891172179968

there is also this failure but i don't think its related
https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_cluster-network-addons-operator/981/pull-e2e-cluster-network-addons-operator-nmstate-functests/1429820891654524928

oshoval:
closing as this is client side, and we want server side filtering of just cnao components, even on external namespaces (using labels most likley with Quique method, but didn't try yet)

first controller runtime is needed

PR #981, opened 2021-08-23
WIP, try to watch only operator namespace

oshoval:

```release-note
None
```


oshoval:
we see that nmstate which is deployed on other namespace than the operator namespace, is not found, and this is a problem because we need it for smooth upgrade where the nmstate is on different namespace

https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_cluster-network-addons-operator/981/pull-e2e-cluster-network-addons-operator-workflow-k8s/1429820891172179968


closing

PR #988, opened 2021-08-25
kmp functests, Fix getMetrics e2e test

oshoval:
Since getMetrics uses kmp, and kmp e2e tests
need to run from the directory where the cluster was created,
export CLUSTER_ROOT_DIRECTORY, which will be used by kmp e2e tests
in order to use the cluster created by CNAO.

```release-note
None
```


oshoval:
/cc @RamLavi 

we will need this so the new bump will work
maybe we don't even need new version, because just the tests files are going to be changed ?
https://github.com/k8snetworkplumbingwg/kubemacpool/pull/330

oshoval:
weird failure 
https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_cluster-network-addons-operator/988/pull-cluster-network-addons-operator-unit-test/1430516712909836288

lets try

/retest

PR #1031, opened 2021-09-29
status-manager, Fix bad compare of status

RamLavi:
When comparing statuses to see if the status has changed, there is a bug
since oldStatus is a pointer and config.Status is not, hence the
DeepEqual() will always return true

**Note** that this fix will not reduce status issues, since the heartbeat of statuses `Degraded` and `Available` are updated every reconcile.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/controller/statusmanager/status_manager.go
+++ b/pkg/controller/statusmanager/status_manager.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"log"
 	"os"
-	"reflect"
 	"strings"
 	"sync"
 	"time"
@@ -178,7 +177,7 @@ func (status *StatusManager) set(reachedAvailableLevel bool, conditions ...condi
 	// Failing condition had been replaced by Degraded in 0.12.0, drop it from CR if needed
 	conditionsv1.RemoveStatusCondition(&config.Status.Conditions, conditionsv1.ConditionType("Failing"))
 
-	if reflect.DeepEqual(oldStatus, config.Status) {
+	if (*oldStatus).DeepEqual(config.Status) {
 		return nil
 	}
 
--- a/pkg/apis/networkaddonsoperator/shared/networkaddonsconfig_operators.go
+++ b/pkg/apis/networkaddonsoperator/shared/networkaddonsconfig_operators.go
@@ -1,0 +1,7 @@
+package shared
+
+import "reflect"
+
+func (s NetworkAddonsConfigStatus) DeepEqual(statusToCompare NetworkAddonsConfigStatus) bool {
+	return reflect.DeepEqual(s, statusToCompare)
+}


RamLavi:
/cherry-pick release-0.58

@phoracek do we want to cherry-pick to other stable branches?

oshoval:
> DeepEqual() will always return true

you mean false maybe ?

pkg/controller/statusmanager/status_manager.go
```diff
@@ -178,7 +178,7 @@ func (status *StatusManager) set(reachedAvailableLevel bool, conditions ...condi
 	// Failing condition had been replaced by Degraded in 0.12.0, drop it from CR if needed
 	conditionsv1.RemoveStatusCondition(&config.Status.Conditions, conditionsv1.ConditionType("Failing"))
 
-	if reflect.DeepEqual(oldStatus, config.Status) {
+	if reflect.DeepEqual(*oldStatus, config.Status) {
```
qinqon, line 181: What do you think about using what is proposed here ? https://github.com/kubernetes-sigs/kubebuilder/issues/592#issuecomment-625738183

qinqon (changes requested):
Looks like there are better ways to do this.

pkg/controller/statusmanager/status_manager.go
RamLavi: interesting! will try

RamLavi: ok, so I compared the two, and I don't think it's suitable for this case. 
the reason is that `DeepDerivative` ignores unset parameters (see example), which is good for comparing configuration specs, but not good for comparing statuses. For example, if a condition is removed/added from/to the new status, we want to update on that.

This is an example of how it works (this unit tests passes)
```
Context("LALALA", func() {
		status1 := shared.NetworkAddonsConfigStatus{}
		status2 := shared.NetworkAddonsConfigStatus{}
		It("RI!!", func() {
			Expect(equality.Semantic.DeepDerivative(status1, status2)).To(BeTrue(), "1.1")
			Expect(equality.Semantic.DeepDerivative(status2, status1)).To(BeTrue(), "1.2")

			status1.Conditions = []conditionsv1.Condition{}
			Expect(equality.Semantic.DeepDerivative(status1, status2)).To(BeTrue(), "2.1")
			Expect(equality.Semantic.DeepDerivative(status2, status1)).To(BeTrue(), "2.2")

			status1.Conditions = []conditionsv1.Condition{
				conditionsv1.Condition{Reason: "simple diff"},
			}
			Expect(equality.Semantic.DeepDerivative(status1, status2)).To(BeFalse(), "3.1")
			Expect(equality.Semantic.DeepDerivative(status2, status1)).To(BeTrue(), "3.2")
		})
	})
```

so IMO better keep to DeepEqual  in this case

qinqon: Can we encapsulate this into a function that expect proper types? Also check if we are doing the same at other parts of the code and use that function, or similar. This is the kind of things that are super difficult to find if something is failing, this kind of unsafe code has to be encapsulated.

RamLavi: >Can we encapsulate this into a function that expect proper types?

can you explain? I'm not sure I understand your concern.

btw we don't use `reflect.DeepEqual()` anywhere else on CNAO.

qinqon: I mean creating a `func (s shared.NetworkAddonsConfigStatus) Equal(statusToCompare shared.NetworkAddonsConfigStatus) bool` so we encapsulate the reflect call.

RamLavi: ohhh now I get what you mean and why you asked this.. I was on another plane hehe, sure will do

RamLavi: since `shared.NetworkAddonsConfigStatus` is defined on `pkg/apis/shared/networkaddonsconfig_types.go`, do you think it's ok to add the function there? I worry that this file might break operator-sdk scripts there

Is it ok if I add a function on statusManager? `(status *StatusManager) Equal(oldStatus, newStatus cnao.NetworkAddonsConfigStatus) bool`

qinqon: We can add at "shared" package whatever we want, we don't have to add the functions to _types.go we can add a _operators.go

RamLavi: I'll open a `networkaddonsconfig_operators.go` on `cluster-network-addons-operator/pkg/apis/networkaddonsoperator/shared/`

pff.. well let deal with it one dish at a time :) I'll open an [issue](https://github.com/kubevirt/cluster-network-addons-operator/issues/1033) to align with operator-sdk/kubebuilder

RamLavi: fixed [here](https://github.com/kubevirt/cluster-network-addons-operator/compare/526fa751f785167f56341f7af86d09e1272e72c5..be1322531b382b42a88351ee401f9ca48302705b)

qinqon (changes requested):
/lgtm
/approve

PR #1032, opened 2021-09-29
ReconcileNetworkAddonsConfig, Reduce unneeded status updates

RamLavi:
Currently SetFromPods() is called in the following scenarios:
1. from ReconcilePods when a change in one of the objects are made
2. from ReconcileNetworkAddonsConfig on each Reconcile to tackle a
corner case where there are no pods deployed.
3. from ReconcileNetworkAddonsConfig on trackDeployedObjects()

Calls 2 and 3 are redundant, and 3 Should be called only when the corner
case happens.
Merged the calls in order to reduce unneeded status updates.

**Release note**:

```release-note
NONE
```


RamLavi:
/cherry-pick release-0.58

@phoracek do we want to cherry-pick to other stable branches?

phoracek:
@RamLavi we should get it at least to 0.53. Radim is now backporting an nmstate fix that should make it to 0.53 too, so we could release them together

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
```diff
@@ -507,8 +501,13 @@ func (r *ReconcileNetworkAddonsConfig) trackDeployedObjects(objs []*unstructured
 
 	r.podReconciler.SetResources(allResources)
 
-	// Trigger status manager to notice the change
-	r.statusManager.SetFromPods()
+	// From now on, r.podReconciler takes over NetworkAddonsConfig handling, it will track deployed
+	// objects if needed and set NetworkAddonsConfig.Status accordingly. However, if no resources were
+	// deployed, there is nothing that would trigger initial reconciliation. Therefore, let's
+	// perform the first check manually.
+	if len(allResources) == 0 {
```
oshoval, line 508: shouldn't it be != 0 ? (according comment "However, if no resources were deployed, there is nothing that would trigger initial reconciliation")

RamLavi: ah.. I don't think so..
`if no resources were deployed..` i.e. len(allResources) == 0, `..there is nothing that would trigger initial reconciliation.`
`Therefore, let's perform the first check manually.`, then run r.statusManager.SetFromPods() manually, and not from the podController.

alonSadan (approved):
Nice!
Thanks

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
```diff
@@ -313,7 +313,7 @@ func (r *ReconcileNetworkAddonsConfig) Reconcile(request reconcile.Request) (rec
 
 	// Everything went smooth, remove failures from NetworkAddonsConfig if there are any from
 	// previous runs.
-	r.statusManager.SetNotFailing(statusmanager.OperatorConfig)
+	r.statusManager.Set(false, r.statusManager.SetNotFailing(statusmanager.OperatorConfig))
```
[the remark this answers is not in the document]
phoracek: Could we keep a dedicated function for this? It has static false and the argument that is generated from the setter of the same instance feels awkward

PR #1074, opened 2021-11-09
Add pod security admission

alonSadan:
This PR introduces the new [PodSecurity admission controller](https://kubernetes.io/docs/concepts/security/pod-security-admission/#applying-to-all-namespaces) which eventually should replace Openshift's SecurityContextConstraints.

This is an initial PR that is not intended to replace SCC, and the new PodSecurity admission is applied in a no-op
mod: not restrictions what so ever. This enables PodSecurity admission to co-exist with SCC and to be monitored and checked.

The new admission controller is applied in the namespace level, and this PR adds it to the namespace yamls of linux-bridge, ovs, and nmstate. Though in the current default behavior all of them will live under "cluster-network-addons" 
namespace,

A followup PR should drop SCC, reduce the "permissiveness"  of the PodSecurity admission, and perhaps apply the different components in different namespaces to apply each component with it's specific security restrictions. 

**Release note**:

```release-note
Introduce PodSecurity admission controller
```

### DIFF
--- a/pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
+++ b/pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
@@ -202,6 +202,15 @@ func (r *ReconcileNetworkAddonsConfig) Reconcile(request reconcile.Request) (rec
 		return reconcile.Result{}, nil
 	}
 
+	podSecurityEnabled, err := isPodSecurityEnabled(r.client)
+	if err != nil {
+		return reconcile.Result{}, errors.Wrap(err, "failed to check if PodSecurity admission enabled")
+	}
+	if podSecurityEnabled {
+		log.Printf("PodSecurity admission contorller is enabled")
+	}
+	r.clusterInfo.PodSecurityEnabled = podSecurityEnabled
+
 	// Check for NMState Operator
 	nmstateOperator, err := isRunningKubernetesNMStateOperator(r.client)
 	if err != nil {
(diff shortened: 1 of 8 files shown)


pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
```diff
@@ -710,3 +719,30 @@ func isOpenshiftSingleReplica(c k8sclient.Client) (bool, error) {
...
 }
+
+func isPodSecurityEnabled(c k8sclient.Client) (bool, error) {
+	apiPod, err := apiServerPod(c)
+	if err != nil {
+		return false, err
+	}
+
+	for _, cmd := range apiPod.Spec.Containers[0].Command {
+		if strings.Contains(cmd, "PodSecurity=true") {
```
rhrazdil, line 730: When this is no longer an alpha feature, could it happen that this would be enabled by default without the need of the explicit featuregate?
I'm asking because in that case, we may miss detecting `PodSecurity` to be enabled

alonSadan: It could happen that it will be enabled by default. The thing is by now CNAO can be applied on k8s versions lower than 1.22, and the API might not exist. I could not think of a better way to discover this.
Do you have an idea?

Perhaps when it is not alpha anymore it will appear in the `- --enable-admission-plugins` section but this is just speculation

phoracek:
As a part of HCO we are sharing the namespace with all the other components ... we are not owners of it and should not enforce any permissions, that should be on HCO if anything. Giving this a second thought, I'm starting to thing we should not proceed with this.

I originally thought this is a replacement for SCC. However, as Alon noted it is quite unrelated, this affects the whole namespace (which causes ownership problem) and it is also probably more useful for unprivileged clients. It is not clear to me how this security admission would be beneficial to us or a customer.

alonSadan:
> ll the other components ... we a

Thanks @phoracek 
So for this reasoning I am closing this PR.

Issue #1080, opened 2021-11-11
Unit tests on main branch are failing


oshoval:
**What happened**:
Unit tests on main branch are failing
the sha are missing for some reason
https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_cluster-network-addons-operator/1079/pull-cluster-network-addons-operator-unit-test/1458816986187304960#1:build-log.txt%3A9791

**What you expected to happen**:
Should pass

**How to reproduce it (as minimally and precisely as possible)**:
Just run it on main branch CI

**Anything else we need to know?**:

**Environment**:

RamLavi:
seems like an issue with skipeo? though I don't see it on local
```
+ BRIDGE_MARKER_TAG=0.9.1
+ BRIDGE_MARKER_IMAGE=quay.io/kubevirt/bridge-marker
+ BRIDGE_MARKER_IMAGE_TAGGED=quay.io/kubevirt/bridge-marker:0.9.1
++ docker-utils::get_image_digest quay.io/kubevirt/bridge-marker:0.9.1 quay.io/kubevirt/bridge-marker
+++ docker run --rm quay.io/skopeo/stable:latest inspect docker://quay.io/kubevirt/bridge-marker:0.9.1
+++ jq -r .Digest
runtime/cgo: pthread_create failed: Operation not permitted
SIGABRT: abort
PC=0x7efe3332b85c m=0 sigcode=18446744073709551610

goroutine 0 [idle]:
runtime: unknown pc 0x7efe3332b85c
stack: frame={sp:0x7ffecb633bc0, fp:0x0} stack=[0x7ffecae35120,0x7ffecb634160)
00007ffecb633ac0:  0000000000000000  00007ffecb633b08 
00007ffecb633ad0:  0000555db43103fa <runtime.mmap.func1+90>  00007efe1e8d1000 
00007ffecb633ae0:  0000000000001000  0000003200000003 
00007ffecb633af0:  00000000ffffffff  00007efe1e8d1000 
00007ffecb633b00:  00007ffecb633b48  00007ffecb633b80 
00007ffecb633b10:  0000555db42b127e <runtime.mmap+158>  00007ffecb633b50 
00007ffecb633b20:  00007ffecb633b70  0000555db42c74a5 <runtime.sysAlloc+261> 
00007ffecb633b30:  0000555db56aa578  00007ffecb633b60 
[...]

PR #1081, opened 2021-11-12
docker-utils, Remove seccomp to avoid pthread permission error

RamLavi:
seccomp is a sandboxing facility in the Linux kernel that acts
like a firewall for system calls (syscalls).
in recent quay.io/skopeo/stable:v1.5.*, certain commands were
added to the pthread_create function, that is not allowed by the
default seccomp profile.
to avoid pthread permission error on docker skopeo application,
and to avoid handeling of seccomp profile, disable the seccomp
policy entirely.
Since docker-utils is only used for bump scripts, and not the
actual binary, there is no need to use seccomp

fixes https://github.com/kubevirt/cluster-network-addons-operator/issues/1080

**Release note**:

```release-note
NONE
```

### DIFF
--- a/hack/components/docker-utils.sh
+++ b/hack/components/docker-utils.sh
@@ -13,7 +13,7 @@ set -xo pipefail
 # The image digest
 #
 function docker-utils::get_image_digest() {
-  echo "${2}@$(docker run --rm quay.io/skopeo/stable:latest inspect "docker://${1}" | jq -r '.Digest')"
+  echo "${2}@$(docker run --rm --security-opt seccomp=unconfined quay.io/skopeo/stable:latest inspect "docker://${1}" | jq -r '.Digest')"
 }
 
 # The check_image_exists function checks if an image already exists in the registry.
@@ -25,5 +25,5 @@ function docker-utils::get_image_digest() {
 # returns the image tag if found; else, returns an empty result
 #
 function docker-utils::check_image_exists() {
-  docker run --rm quay.io/skopeo/stable:latest list-tags "docker://${1}" | grep "\"${2}\""
+  docker run --rm --security-opt seccomp=unconfined quay.io/skopeo/stable:latest list-tags "docker://${1}" | grep "\"${2}\""
 }


RamLavi:
/hold
@phoracek in your opinion is it a risky to remove the setcomp policy?
I tend to say that since `docker skopeo` is only used when bumping/unit-testing - then it's not an issue.

phoracek:
I can't say with certainty. It sounds mostly harmless unless we are worried that skopeo escapes and causes harm in our release automation.

Can we avoid this decision? You haven't described why it started failing. If we understand that, maybe a cleaner solution will become apparent.

oshoval:
can't we just use an older skopeo version ?
we are using `quay.io/skopeo/stable:latest`

RamLavi:
> I can't say with certainty. It sounds mostly harmless unless we are worried that skopeo escapes and causes harm in our release automation.
> 
> Can we avoid this decision? You haven't described why it started failing. If we understand that, maybe a cleaner solution will become apparent.

it started failing because in the new version they started using a new command that is not allow by the seccom filter.
I asked skopeo maintainer to give me more information about it, but it looks like our options are:
1. As Or Shoval said, pin to a specific skopeo version. I am firmly against it, if we do that and it breaks in the future then we will need to maintain and fix it ourselves. 
2. understand which syscalls were added and create our own seccom profile and use it using the `--security-opt seccomp=new-profile` option. I don't like it, it's just another thing to handle.
3. disable seccomp policy. since it's not compromising anything then I vote to go with this approach.

oshoval:
> > I can't say with certainty. It sounds mostly harmless unless we are worried that skopeo escapes and causes harm in our release automation.
> > Can we avoid this decision? You haven't described why it started failing. If we understand that, maybe a cleaner solution will become apparent.
> 
> [updated commit with more info] it started failing because in the new version they started using a new command that is not allowed by the seccomp filter. I [asked](https://github.com/containers/skopeo/issues/1501#issuecomment-968326604) skopeo maintainer to give me more information about it, but it looks like our options are:
> 
> 1. As @oshoval said, pin to a specific skopeo version. I am firmly against it, if we do that and it breaks in the future then we will need to maintain and fix it ourselves.

I don't think its going to be broken so fast, (we just need to mirror it to quay so it won't get deleted, but this also won't be fast looking on the quay).
and until it broke the newer version will be fixed, and then we can switch to it.

> 2. understand which syscalls were added and create our own seccomp profile and use it using the `--security-opt seccomp=new-profile` option. I don't like it, it's just another thing to handle.
> 3. disable seccomp policy. since it's not compromising anything then I vote to go with this approach.

oshoval:
PR Desc says 1.4.1 is broken, but according the issue on skopeo 1.4.1 is still fine
"the errors appeared with 1.5.x, 1.4.1 is working fine."

RamLavi:
> > I can't say with certainty. It sounds mostly harmless unless we are worried that skopeo escapes and causes harm in our release automation.
> > Can we avoid this decision? You haven't described why it started failing. If we understand that, maybe a cleaner solution will become apparent.
> 
> [updated commit with more info] it started failing because in the new version they started using a new command that is not allowed by the seccomp filter. I [asked](https://github.com/containers/skopeo/issues/1501#issuecomment-968326604) skopeo maintainer to give me more information about it, but it looks like our options are:
> 
> 1. As @oshoval said, pin to a specific skopeo version. I am firmly against it, if we do that and it breaks in the future then we will need to maintain and fix it ourselves.
> 2. understand which syscalls were added and create our own seccomp profile and use it using the `--security-opt seccomp=new-profile` option. I don't like it, it's just another thing to handle.
> 3. disable seccomp policy. since it's not compromising anything then I vote to go with this approach.

Hey @qinqon @maiqueb, I would appreciate your opinions as to which option you think is best.
specifically, do you think pinning to the last working version is best (option 1) ?, or we should just ignore the problem altogether  (option 3)

maiqueb:
From what I've seen in the [skopeo issue](https://github.com/containers/skopeo/issues/1501#issuecomment-969235713) this shouldn't reproduce in newer versions of docker. 

If possible, I'd change the OCI binary directly to podman.

RamLavi:
@phoracek @qinqon could you take a look?

maiqueb:
I don't understand why this suddenly is again a good idea, rather than updating the container runtime, which was suggested https://github.com/containers/skopeo/issues/1501#issuecomment-969235713 . 

Running unconfined is not a good thing; it's literally saying to selinux process XYZ can do whatever it wants.

RamLavi:
> I don't understand why this suddenly is again a good idea, rather than updating the container runtime, which was suggested [containers/skopeo#1501 (comment)](https://github.com/containers/skopeo/issues/1501#issuecomment-969235713) .
> 
> Running unconfined is not a good thing; it's literally saying to selinux process XYZ can do whatever it wants.

Hey @maiqueb, I agree with everything you say, and would like to consider alternate safer choices (the current use is only for tests, but introducing this could result with it slipping to more important operations so again I'm totally with you on this).
 
Anyway, upgrading to the latest docker as suggested on the original issue does not work on my local ("nothing to upgrade" yet the issue persists) - see this [PR](https://github.com/kubevirt/cluster-network-addons-operator/pull/1098) trying to simulate that..
However, I do see that if I use fedora 35 the issue disappears. perhaps what we need is to use a more up to date OS? 
@fgimenez we are currently using [this](https://github.com/kubevirt/project-infra/blob/main/github/ci/prow-deploy/files/jobs/kubevirt/cluster-network-addons-operator/cluster-network-addons-operator-presubmits.yaml#L25) image, is there one with a more updated OS?

RamLavi:
@maiqueb could you review the project infra PR? https://github.com/kubevirt/project-infra/pull/1774

RamLavi:
issue is resolved by updating the lane image to a more updated fedora kubevirt/project-infra#1774
closing PR

PR #1082, opened 2021-11-18
hack/components, Move docker-utils to podman

RamLavi:
Recent skopeo releaeses updated the OS and moved to
use more advanced syscalls, ones that are currently
filtered by docker (issue [0]).
In order to move past this issue, moving to Podman.

[0] https://github.com/containers/skopeo/issues/1501

fixed #1081 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/hack/components/bump-bridge-marker.sh
+++ b/hack/components/bump-bridge-marker.sh
@@ -4,7 +4,7 @@ set -xeo pipefail
 
 source hack/components/yaml-utils.sh
 source hack/components/git-utils.sh
-source hack/components/docker-utils.sh
+source hack/components/podman-utils.sh
 
 #here we do all the object specific parametizing
 function __parametize_by_object() {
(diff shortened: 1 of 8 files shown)


oshoval (commented):
Kubevirt checks if either podman exists, and if not fallback to docker
I think it might deserve to do that, at least in the beginning ?
(just either add the seccomp / pin version so https://github.com/kubevirt/cluster-network-addons-operator/pull/1081 will be fixed as well)
installing podman on RHEL was pain IIRC

moreover unittest here suffer from the same problem still

Personally i think we should first fix on docker and only then migrate to podman if desired.

RamLavi:
>Kubevirt checks if either podman exists, and if not fallback to docker
I think it might deserve to do that, at least in the beginning ?
(just either add the seccomp / pin version so #1081 will be fixed as well)
installing podman on RHEL was pain IIRC
moreover if it is only for skopeo bad syscalls, i don't think we should enforce having podman
even just enforce having skopeo installed is better
moreover unittest here suffer from the same problem still
Personally i think we should first fix on docker and only then migrate to podman if desired.

The lane failed because podman is not installed on the test-image.
I don't think there is a problem using podman, we only use it for bumping stuff, it mainly it is an internal tool.
I do agree that since installing podman on RHEL is problematic, we might want to reconsider this PR..

phoracek:
@RamLavi what is the plan with this PR?

RamLavi:
@phoracek moving to podman doesn't seem like a good plan here.
I think I'll go back to removing the `seccomp` filtering approach. closing PR

maiqueb:
> > Kubevirt checks if either podman exists, and if not fallback to docker
> > I think it might deserve to do that, at least in the beginning ?
> > (just either add the seccomp / pin version so #1081 will be fixed as well)
> > installing podman on RHEL was pain IIRC
> > moreover if it is only for skopeo bad syscalls, i don't think we should enforce having podman
> > even just enforce having skopeo installed is better
> > moreover unittest here suffer from the same problem still
> > Personally i think we should first fix on docker and only then migrate to podman if desired.
> 
> The lane failed because podman is not installed on the test-image. I don't think there is a problem using podman, we only use it for bumping stuff, it mainly it is an internal tool. I do agree that since installing podman on RHEL is problematic, we might want to reconsider this PR..

Why is installing podman on RHEL problematic @RamLavi ? AFAIU, it is the default container runtime on rhel. [Quote from RHEL 8 docs](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html-single/building_running_and_managing_containers#assembly_starting-with-containers_building-running-and-managing-containers):
```
Unlike other container tools implementations, the tools described here do not center around the monolithic Docker container engine and docker command. 

Instead, Red Hat provides a set of command-line tools that can operate without a container engine. These include: 
- podman
- ...
```

And furthermore, what does RHEL have to do with this ? I know next to nothing about this repo's build tooling. And I have no logs to know what happened. (I guess that when you close a PR the associated builds are flushed - just a theory).

PR #1118, opened 2021-12-14
placement, infra: fallback to master node

qinqon:
At some environment control-plane is still not present, and nodes are
labeled with master only. This change replace toleratin + node selector
with a affinity rules that implement a fallback to master if
control-plane is not there.

**Release note**:

```release-note
Suport "master" control-plane node labels for infra placement.
```

### DIFF
--- a/pkg/network/placement_configuration.go
+++ b/pkg/network/placement_configuration.go
@@ -9,15 +9,44 @@ import (
 func GetDefaultPlacementConfiguration() cnao.PlacementConfiguration {
 	return cnao.PlacementConfiguration{
 		Infra: &cnao.Placement{
-			NodeSelector: map[string]string{
-				"beta.kubernetes.io/arch":               "amd64",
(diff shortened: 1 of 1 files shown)


pkg/network/placement_configuration.go
```diff
@@ -9,15 +9,32 @@ import (
 func GetDefaultPlacementConfiguration() cnao.PlacementConfiguration {
 	return cnao.PlacementConfiguration{
 		Infra: &cnao.Placement{
-			NodeSelector: map[string]string{
-				"beta.kubernetes.io/arch":               "amd64",
-				"node-role.kubernetes.io/control-plane": "",
-			},
-			Tolerations: []corev1.Toleration{
```
tiraboschi, line 16: I think you should keep this and also add back a toleration for node-role.kubernetes.io/master

tiraboschi (approved):

PR #1124, opened 2021-12-22
Add default linux OS nodeSelector to nmstate handlers

rhrazdil:
This change ensures that nmstate handler pods have the
`kubernetes.io/os: linux` node selector.

**Release note**:

```release-note
add kubernetes.io/os=linux nodeSelector to nmstate handler pods by default.
```

### DIFF
--- a/pkg/network/nmstate.go
+++ b/pkg/network/nmstate.go
@@ -17,25 +17,31 @@ import (
 	cnao "github.com/kubevirt/cluster-network-addons-operator/pkg/apis/networkaddonsoperator/shared"
 )
 
+const (
+	kubernetesOsSelector = "kubernetes.io/os"
+)
+
 // renderNMState generates the manifests of NMState handler
 func renderNMState(conf *cnao.NetworkAddonsConfigSpec, manifestDir string, clusterInfo *ClusterInfo) ([]*unstructured.Unstructured, error) {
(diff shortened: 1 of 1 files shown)


pkg/network/nmstate.go
```diff
@@ -169,3 +175,15 @@ func getMinNumberOfWebhookReplicas(clusterInfo *ClusterInfo) int32 {
 
 	return highlyAvailableMinWebhookReplicas
 }
+
+func pupulateDefaultConfiguration(conf *cnao.NetworkAddonsConfigSpec) *cnao.NetworkAddonsConfigSpec {
```
qinqon, line 179: This is all at the same package let mark it with "Nmstate"
```suggestion
func pupulateNmstateDefaultConfiguration(conf *cnao.NetworkAddonsConfigSpec) *cnao.NetworkAddonsConfigSpec {
```

qinqon (changes requested):

pkg/network/nmstate.go
rhrazdil: Ah, good point. [Updated](https://github.com/kubevirt/cluster-network-addons-operator/compare/d0cd05972a7bccd93e2cea4fce9739e169ae7ee6..e4d4ab762341112ecfdb5f7fb749bd52fa3bc833)

PR #1130, opened 2021-12-27
Operator image: Pinned Centos version

orelmisan:
Pinned Centos version to centos7.9.2009.
The Centos version was defined to be latest centos7.
This could have lead to non-uniform image builds.

~~Based on PR #1128, Please review the second commit only.~~
**Release note**:

```release-note
Using centos7.9.2009 as the base image.
```

### DIFF
--- a/build/operator/Dockerfile
+++ b/build/operator/Dockerfile
@@ -1,4 +1,4 @@
-FROM quay.io/centos/centos:centos7
+FROM quay.io/centos/centos:centos7.9.2009
 ENV ENTRYPOINT=/entrypoint \
     OPERATOR=/cluster-network-addons-operator \
     MANIFEST_TEMPLATOR=/manifest-templator \


build/operator/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM centos:centos7
+FROM quay.io/centos/centos:centos7.9.2009
```
oshoval, line 1: imo we don't need to be that strict, we usually pin the major only (for example on CNAO we used centos7 and when we need fedora we are using fedora32)
Moreover in case minor / patch versions are added, we better have them as well, because they can be either security / bug fix,
and in case they are breaking something, we will know and adapt our code, and not only when we change this line.
Changing it also might require maintenance that we would need to bump it manually from time to time

EdDev (approved):
I agree that pinning to a specific version/build is the right thing to do.

Having a "floating" version results in nondeterministic results each time a build is triggered.

Placing a hold to emphasize that there is a dependent PR that should be taken first.
/hold

oshoval:
Please see https://github.com/kubevirt/cluster-network-addons-operator/pull/1131
that fixes the rdoproject failures

btw it seems rdo is back now, so we can either keep 1131 or use it if it happens again

/retest

orelmisan:
rebase

orelmisan:
This PR was based on PR #1128 that was merged.
/hold cancel

build/operator/Dockerfile
qinqon: We can even use this PR to pin it to ubi8-minimal or tu centos 8 stream, centos7 is quite ancient.

qinqon (changes requested):

build/operator/Dockerfile
EdDev: One is pinning and the other is upgrading. I do not think it is the same, one has zero risk and the other requires much more work and has potential to find numerous issues.

orelmisan: I agree with @EdDev.
The replacement of a base image is a subject for another PR.

qinqon: Then let's just pin major.minor so we receive patch fixes.

```
quay.io/centos/centos:centos7.9
```

qinqon (changes requested):

build/operator/Dockerfile
EdDev: So you assume the patches do not cause any regression, right?
As I see it, it is still not under control... it will update the version per-build without any control.

I find this problematic, but it may worth checking how other project manage such things.
If I take golang as an example, they pin it with a specific hash. Which allows predictable builds without surprises.

I guess a policy needs to be define in general, something that all parts of the project can follow.

orelmisan: IMO builds should be reproducible.
For example, if you don't pin the base image version and you fix a bug in an older release and rebuild the image - you get new changes from the base image that weren't there when you've first released the image.

Its a trade-off between using known and tested packages vs. the cost of maintenance / automation.

phoracek:
I agree with Quique above. Since CNAO has virtually no system dependencies, I would rather switch to a smaller image an pin at most to the minor version, so we would still receive updates on rebuilds. The chance that any of these updates would break CNAO is negligable as CNAO is a mere binary with no dependencies on the system.

EdDev:
> has virtually no system dependencies

If that is correct, this PR should be replaced with one that is sourcing `scratch`.
And debugging is not a good reason to keep a full OS instead, for debugging there are dedicated containers that share the same ns/s with the target container.

The reasoning not to pin is also colliding (and contradicting) the reasoning to pin. On one hand it is not dependent on anything in the OS, and on the other hand you want it updated.

Anyway, this is surely not a deal breaker. All possible arguments have been said, I think it can be concluded one way or the other.

orelmisan:
The base image had changed to centos:stream9 that has no other tags.

PR #1131, opened 2021-12-28
open Vswitch: Remove RDO from open Vswitch install

oshoval:

Deploying CNAO fails because https://rdoproject.org is not accessible.
rdoproject is used when installing open Vswitch.
    
We are using already another version centos-release-openstack-wallaby
so we can remove rdoproject.
    
K8s-nmstate has tests of open Vswitch as part of CNAO gating,
and those tests are passing.

**Release note**:

```release-note
None
```


oshoval:
RDO is back online, so i will close this PR, unless it will be down again and again

PR #1132, opened 2021-12-30
kubevirtci: Bump kubevirtci in order to fix rate limits

oshoval:

Newer kubevirtci mirrors images such as the library/registry
to quay.io instead of using docker.io.
This will fix the docker.io rate limits.

**Release note**:

```release-note
None
```


oshoval:
tests are failing, dropping this until we need to bump for more reasons

PR #1145, opened 2022-01-24
labels: Do not label operator namespace with unneeded labels

oshoval:

CNAO labels the components it deploys with miscellaneous labels, for example relationship labels.
In case the component is in the same namespace as the CNA operator itself,
it will be labeled as well.

Since K8s logs are labeled with `kubernetes.namespace_name` according
the labels the namespace has, all the cnv logs in the namespace
would appear as their `kubernetes.io/component` equal the value CNAO sets.
Same for the other labels that CNAO adds to the namespace.

Therefore:

Do not add labels to the namespace of the operator itself.

In order to guarantee backward compatibility, CNAO will also remove
those labels from the operator namespace according to their key.
It is safe to do that, because those labels should not be added by anyone,
else the problem will reoccur.

Affected labels:

Relationship labels:
* `app.kubernetes.io/component`
* `app.kubernetes.io/part-of`
* `app.kubernetes.io/version`
* `app.kubernetes.io/managed-by`

Other labels:
* `prometheus.cnao.io`
* `networkaddonsoperator.network.kubevirt.io/version`
* `control-plane` (added because it exists on kubemacpool namespace)

Note:
Once there will be no more supported versions that added those labels to the namespace
it will be safe to remove `cleanUpNamespaceRelationLables`.

Fixes
https://bugzilla.redhat.com/show_bug.cgi?id=2033385

Fixed typo (crLables -> crLabels)

**Release note**:
```release-note
None
```

### DIFF
--- a/pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
+++ b/pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
@@ -431,10 +431,9 @@ func (r *ReconcileNetworkAddonsConfig) applyObjects(networkAddonsConfig metav1.O
 		// Don't set owner reference on namespaces if they are used by the operator itself
 		// Don't set owner reference on CRDs, they should survive removal of the operator
 		// Don't set owner reference on objects that explicitly rejected an owner
-		isOperatorNamespace := obj.GetKind() == "Namespace" && obj.GetName() == operatorNamespace
 		isCRD := obj.GetKind() == "CustomResourceDefinition"
 		_, isRejectingOwner := obj.GetAnnotations()[names.REJECT_OWNER_ANNOTATION]
-		if !isCRD && !isOperatorNamespace && !isRejectingOwner {
+		if !isCRD && !isOperatorNamespace(obj) && !isRejectingOwner {
 			if err := controllerutil.SetControllerReference(networkAddonsConfig, obj, r.scheme); err != nil {
 				log.Printf("could not set reference for (%s) %s/%s: %v", obj.GroupVersionKind(), obj.GetNamespace(), obj.GetName(), err)
 				err = errors.Wrapf(err, "could not set reference for (%s) %s/%s", obj.GroupVersionKind(), obj.GetNamespace(), obj.GetName())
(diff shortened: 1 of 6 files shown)


pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
```diff
@@ -520,39 +519,44 @@ func (r *ReconcileNetworkAddonsConfig) stopTrackingObjects() {
...
-		labels = updateLabelsFromCR(labels, crLables, appLabelKeys)
-		if err != nil {
-			return err
+			templateLabelKeys := append(appLabelKeys, names.PROMETHEUS_LABEL_KEY, names.MANAGED_BY_LABEL_KEY)
+			err = updateObjectTemplateLabels(obj, labels, templateLabelKeys)
+			if err != nil {
+				return err
+			}
+		} else {
+			delete(labels, names.KUBEMACPOOL_CONTROL_PLANE_KEY)
```
oshoval, line 547: if we remove this label from kmp
https://github.com/k8snetworkplumbingwg/kubemacpool/blob/e5c69e0641317336e13405c5476c7fb750e32ff4/config/default/manager/manager.yaml#L5

we can remove this line of code (but will still need to remove it at `cleanUpNamespaceLabels` for backward compability

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
```diff
@@ -520,39 +519,44 @@ func (r *ReconcileNetworkAddonsConfig) stopTrackingObjects() {
...
 	for _, obj := range objs {
 		labels := obj.GetLabels()
 		if labels == nil {
 			labels = map[string]string{}
 		}
-		// Label objects with version of the operator they were created by
-		labels[cnaov1.SchemeGroupVersion.Group+"/version"] = operatorVersionLabel
-		labels[names.PROMETHEUS_LABEL_KEY] = names.PROMETHEUS_LABEL_VALUE
-		labels[names.MANAGED_BY_LABEL_KEY] = names.MANAGED_BY_LABEL_DEFAULT_VALUE
+		if !isOperatorNamespace(obj) {
```
qinqon, line 529: Reduce identitation we don't need the else here.
```suggestion
		if isOperatorNamespace(obj) {
		    delete(labels, names.KUBEMACPOOL_CONTROL_PLANE_KEY)    
		}
		....
```

qinqon (changes requested):
Some code simplifications.

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
```diff
@@ -706,3 +710,7 @@ func isOpenshiftSingleReplica(c k8sclient.Client) (bool, error) {
 
 	return infraConfig.Status.InfrastructureTopology == osconfv1.SingleReplicaTopologyMode, nil
 }
+
+func isOperatorNamespace(obj *unstructured.Unstructured) bool {
+	return obj.GetKind() == "Namespace" && obj.GetName() == operatorNamespace
```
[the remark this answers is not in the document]
oshoval: the condition was not presented in this PR, just wrapped to a function
looked all over the code, didn't find a const that is set and then used, both on CNAO and vendors
I will use a const in this function for now, we can do a follow PR if needed to create some k8s consts types
wdyt?

pkg/network/network.go
```diff
@@ -246,3 +249,37 @@ func errorListToMultiLineString(errs []error) string {
...
+// to operator namespace
+func cleanUpNamespaceLabels(ctx context.Context, client k8sclient.Client) []error {
+	namespace := &v1.Namespace{}
+	err := client.Get(context.Background(), types.NamespacedName{Name: os.Getenv("OPERATOR_NAMESPACE")}, namespace)
+	if err != nil {
+		return []error{err}
+	}
+
+	labels := namespace.GetLabels()
+	if labels != nil {
```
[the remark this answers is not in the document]
oshoval: Thanks, changing to this, because also len(nil) is 0
```
	if len(labels) == 0 {
		return []error{}
	}
```

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
oshoval: We need to do lines 530 - 545 for non operator namespace, 
else we need to do only line 547, we can't do out of the box what you suggested, unless you mean something else?

oshoval:
Thanks, addressed the resolved comments

qinqon (approved):
/lgtm
/approve

pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go
phoracek: I think that removing it from KMP directly would be the best option. In general, I think that operators should not manager the Namespace they are deployed into

PR #1146, opened 2022-01-24
placement: Set linux as default OS

qinqon:
All the components at CNAO should deploy at linux node is not specified
differently. This change add the node selector that specify that at the
default placement.

**Release note**:

```release-note
Set default placement to os/linux nodes.
```


pkg/network/placement_configuration.go
```diff
@@ -53,6 +53,7 @@ func GetDefaultPlacementConfiguration() cnao.PlacementConfiguration {
 		Workloads: &cnao.Placement{
 			NodeSelector: map[string]string{
 				"beta.kubernetes.io/arch": "amd64",
+				"kubernetes.io/os":        "linux",
```
phoracek, line 56: What if a user overrides the placement configuration.

Meni raised that it would be desirable to clearly fail NNCP if it cannot configure the policy due to Windows nodes matching it.

PR #1152, opened 2022-01-26
0.65: Mark as static components with main

qinqon:
At a release branch if a component has not release branch itself it has
to be marked at static at components.yaml so autobumper will ignore new
tags.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/components.yaml
+++ b/components.yaml
@@ -3,7 +3,7 @@ components:
     url: https://github.com/kubevirt/bridge-marker
     commit: 46b2b0f81a807bf140f0e90b377c6cea0ca7eb9b
     branch: main
-    update-policy: tagged
+    update-policy: static
     metadata: 0.9.1
   kubemacpool:
     url: https://github.com/k8snetworkplumbingwg/kubemacpool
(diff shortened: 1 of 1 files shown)


rhrazdil (commented):
Don't we want to do this change on main branch as well?

qinqon:
> Don't we want to do this change on main branch as well?

Main branch shuld stay with `tagged` so we receive the github-bot PRs bumping components when they release a new tag, then at release branches we "freeze" those version changing with "static" the components that do not have a release branch, problem with that is that is very error prone, now at CNAO creating a release branch involve the following:
- Create the branch itself at git
- Change components.yaml to freeze the ones with release branchs
- Create the prow jobs for that release branch.

rhrazdil:
Got it, thanks!

> Maybe we can start thinking about automatization to create release branches at leat at CNAO.

Yes, that sounds helpful

qinqon:
/override pull-e2e-cnao-workflow-k8s-release-0.65

Not related the policy to static change is only used by the autobumper.

PR #1163, opened 2022-01-31
cluster: Enable just ovs repo

qinqon:
Some openstack repos does not support centos 8 stream yet. This change
point directly to the centos 8 stream ovs repo.

**Release note**:

```release-note
NONE
```


cluster/up.sh
```diff
@@ -35,9 +35,8 @@ fi
 if [[ "$KUBEVIRT_PROVIDER" =~ k8s- ]]; then
     echo 'Installing Open vSwitch and NetworkManager 1.34 on nodes'
     for node in $(./cluster/kubectl.sh get nodes --no-headers | awk '{print $1}'); do
-        ./cluster/cli.sh ssh ${node} -- sudo yum config-manager --set-enabled powertools
-        ./cluster/cli.sh ssh ${node} -- sudo yum install -y epel-release centos-release-openstack-wallaby epel-release centos-release-openstack-train https://rdoproject.org/repos/rdo-release.rpm
-        ./cluster/cli.sh ssh ${node} -- sudo yum install -y openvswitch libibverbs NetworkManager-1.34.0 NetworkManager-ovs-1.34.0
+        ./cluster/cli.sh ssh ${node} -- sudo dnf install centos-release-nfv-openvswitch -y
```
oshoval, line 38: can it we oneliner install ?
or at least move all the openvswitch to one separate line and the rest on the 2nd ?

please move the -y to the beginning as the other

oshoval (commented):
Thanks

cluster/up.sh
qinqon: Have to be in two lines since we are installing a repository, I will move the `-y` but will keep installation of RPM at one line so we don't take more time to update repo database again.

oshoval (approved):
Thanks

lets just make sure all the tests are passing

cluster/up.sh
```diff
@@ -35,9 +35,8 @@ fi
 if [[ "$KUBEVIRT_PROVIDER" =~ k8s- ]]; then
     echo 'Installing Open vSwitch and NetworkManager 1.34 on nodes'
     for node in $(./cluster/kubectl.sh get nodes --no-headers | awk '{print $1}'); do
-        ./cluster/cli.sh ssh ${node} -- sudo yum config-manager --set-enabled powertools
-        ./cluster/cli.sh ssh ${node} -- sudo yum install -y epel-release centos-release-openstack-wallaby epel-release centos-release-openstack-train https://rdoproject.org/repos/rdo-release.rpm
-        ./cluster/cli.sh ssh ${node} -- sudo yum install -y openvswitch libibverbs NetworkManager-1.34.0 NetworkManager-ovs-1.34.0
+        ./cluster/cli.sh ssh ${node} -- sudo dnf install -y centos-release-nfv-openvswitch
+        ./cluster/cli.sh ssh ${node} -- sudo dnf install -y openvswitch2.16 libibverbs NetworkManager-1.34.0 NetworkManager-ovs-1.34.0
```
rhrazdil, line 39: Maybe we don't need to list libibverbs explicitly as it comes from baseos repo, but can be removed later if we want to.

rhrazdil (approved):

cluster/up.sh
qinqon: Let's unblock CI first.

PR #1177, opened 2022-02-01
nmstate: Check operator pods not deployment

qinqon:
/kind bug

When installing knmstate CNAO try to find kubernetes-nmstate-operator on
the system and if it's there it will do the installation with it, to do
the search it filter the deployment by the "app" label, but this label
is not at deployment but at the pods. This change search for pods
instead of deployment to find if the NMState operator is installed.

**Release note**:

```release-note
Check nmstate-operator looking for its pods.
```


qinqon:
/hold

The nmstate-operator deployment should have that label https://github.com/nmstate/kubernetes-nmstate/blob/main/deploy/operator/operator.yaml#L8

PR #1219, opened 2022-02-10
multus: Bump to 3.8

### LINKED ISSUE #1230: Bump multus to v3.8
**What happened**:
There were a bug at the autobumper so CNAO got never PRs bumping multus, after fixing it we discover that latest multus is breaking CNAO.

The problem with CNAO and latest multus is that now removing a pod created after multus installatin fails if the multus service account is gone, this is the scenario at CNAO ci since it test network components removal.

CI not failing after bumping multus, we should not test removing multus is something that CNO is not doing.

**Anything else we need to know?**:

**Environment**:

qinqon:
Multus project has change a little:
- They don't have a manifests pert arch since they are using multiarch
  registry.
- They have bump golang to 1.17
- The CNI binary is not able to acess apiserver if daemonset is not there https://github.com/k8snetworkplumbingwg/multus-cni/issues/592.

This change moves away golang origin builder since there is non for
1.17

Closes #1230 

**Release note**:

```release-note
Bump multus to v3.8
```

### DIFF
--- a/automation/check-patch.e2e-multus-functests.sh
+++ b/automation/check-patch.e2e-multus-functests.sh
@@ -64,12 +64,6 @@ spec:
 EOF
 }
 
-function __remove-test-manifests() {
-    echo "Remove test manifests"
-    $KUBECTL delete -f ${TEST_YAML}
-    $KUBECTL delete -f ${CNI_PLUGIN_YAML}
-}
-
 main() {
     # Setup CNAO and artifacts temp directory
     source automation/check-patch.setup.sh
@@ -91,8 +85,6 @@ main() {
 
         echo "Run multus macvlan test"
         ./test-simple-macvlan1.sh
-
-        __remove-test-manifests
     )
 }
 
--- a/data/multus/001-multus.yaml
+++ b/data/multus/001-multus.yaml
@@ -23,6 +23,19 @@ spec:
             networks. More information available at: https://github.com/k8snetworkplumbingwg/multi-net-spec'
           type: object
           properties:
+            apiVersion:
+              description: 'APIVersion defines the versioned schema of this represen
+                tation of an object. Servers should convert recognized schemas to the
+                latest internal value, and may reject unrecognized values. More info:
+                https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
+              type: string
+            kind:
+              description: 'Kind is a string value representing the REST resource this
+                object represents. Servers may infer this from the endpoint the client
+                submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
+              type: string
+            metadata:
+              type: object
             spec:
               description: 'NetworkAttachmentDefinition spec defines the desired state of a network attachment'
               type: object
@@ -101,7 +114,6 @@ spec:
         name: kube-multus-ds-amd64
     spec:
       hostNetwork: true
-      nodeSelector: {{ toYaml .Placement.NodeSelector | nindent 8 }}
       tolerations: {{ toYaml .Placement.Tolerations | nindent 8 }}
       serviceAccountName: multus
       containers:
@@ -123,6 +135,7 @@ spec:
             - name: cnibin
               mountPath: /host/opt/cni/bin
           imagePullPolicy: {{ .ImagePullPolicy }}
+      terminationGracePeriodSeconds: 10
       volumes:
         - name: cni
           hostPath:
@@ -131,6 +144,7 @@ spec:
           hostPath:
             path: {{ .CNIBinDir }}
       priorityClassName: system-cluster-critical
+      nodeSelector: {{ toYaml .Placement.NodeSelector | nindent 8 }}
       affinity: {{ toYaml .Placement.Affinity | nindent 8 }}
 {{ if .EnableSCC }}
 ---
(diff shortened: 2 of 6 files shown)


qinqon:
/hold

This is wrong
https://github.com/kubevirt/cluster-network-addons-operator/pull/1219/files#diff-f1531df30500812ec42a33a52023c9ceca8a93d4972132fd241ad143b107c063R31

qinqon:
/hold cancel

Fixed

automation/check-patch.e2e-multus-functests.sh
```diff
@@ -66,8 +66,8 @@ EOF
 
 function __remove-test-manifests() {
     echo "Remove test manifests"
-    $KUBECTL delete -f ${TEST_YAML}
-    $KUBECTL delete -f ${CNI_PLUGIN_YAML}
+    $KUBECTL delete --ignore-not-found=true -f ${TEST_YAML}
+    $KUBECTL delete --ignore-not-found=true -f ${CNI_PLUGIN_YAML}
```
maiqueb, line 70: The multus tests cleanup after themselves, so this is required - or, the function should not be called.

maiqueb (approved):

automation/check-patch.e2e-multus-functests.sh
qinqon: You mean that is NOT requiered right ? calling the cleanup function

maiqueb: yes. The script you're executing is a full blown test - it clean up after itself (i.e. removes the pods + net-attach-def).

You remove the call to the cleanup function, along with its definition.

qinqon:
Removing the networkaddonsconfig end up with webhooks at terminating state with an issue related to multus

```
error killing pod: failed to "KillPodSandbox" for "8e924170-9741-4acf-8037-4be616a79aa6" with KillPodSandboxError: "rpc error: code = Unknown desc = failed to destroy network for pod sandbox k8s_nmstate-webhook-5dd5c69dc5-4jp5c_cluster-network-addons_8e924170-9741-4acf-8037-4be616a79aa6_0(f3e615986195dbb94d864419f56d984330600b105165dd4cba9836ca81019303): Multus: [cluster-network-addons/nmstate-webhook-5dd5c69dc5-4jp5c]: error getting pod: Unauthorized"
```
```
[ellorent@fedora cluster-network-addons-operator]$  ./cluster/kubectl.sh get pod -n cluster-network-addons
selecting docker as container runtime
NAME                                                  READY   STATUS        RESTARTS   AGE
cluster-network-addons-operator-796f9b4586-9dbd6      1/1     Running       0          6m
kubemacpool-mac-controller-manager-684c8cc5c4-xfss4   0/1     Terminating   0          3m6s
kubemacpool-mac-controller-manager-79764f66c9-tp4h7   0/1     Terminating   0          4m56s
nmstate-webhook-5dd5c69dc5-4jp5c                      0/1     Terminating   0          4m56s
nmstate-webhook-5dd5c69dc5-wpt5g                      0/1     Terminating   0          4m56s
```

maiqueb (changes requested):
/lgtm cancel

This is failing CI because the service account is being deleted. I recommend splitting the manifests - one for the service account, another for the daemonset - and only remove the daemonset.

qinqon:
We are still facing the issue, looks like after deleting the multus daemonset the `multus.kubeconfig` is not usable

```bash
$ssh node01 sudo kubectl get pod -n cluster-network-addons multus-xqgsx --kubeconfig=/etc/cni/net.d/multus.d/multus.kubeconfig 
error: You must be logged in to the server (Unauthorized)
```

maiqueb:
Removing the multus conf will make the runtime (k8s) start using the default cluster CNI, which is good. 

Good catch on the https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#service-account-token-volume-projection - deleting the multus pod invalidates that service account. 

Sorry for having misled you. 

@qinqon / @RamLavi I honestly don't know if it is necessary to indicate the running pods with secondary interfaces will leak (and their associated resources) after multus is deleted... I am assuming the users expect that. (if multus is gone, who's going to cleanup those resources ?...)

qinqon:
Looks like bumping to the new multus manifests is not possible

```
2022/02/24 08:08:37 could not apply (apps/v1, Kind=DaemonSet) cluster-network-addons/multus: could not update object (apps/v1, Kind=DaemonSet) cluster-network-addons/multus: DaemonSet.apps "multus" is invalid: spec.selector: Invalid value: v1.LabelSelector{MatchLabels:map[string]string{"name":"kube-multus-ds"}, MatchExpressions:[]v1.LabelSelectorRequirement(nil)}: field is immutable
```

qinqon:
Looks like removing multus daemonset before apply the new manifest introduce some instabilities to CI

data/multus/001-multus.yaml
```diff
@@ -135,6 +135,10 @@ spec:
             - name: cnibin
               mountPath: /host/opt/cni/bin
           imagePullPolicy: {{ .ImagePullPolicy }}
+          lifecycle:
+            preStop:
+              exec:
+                command: ["/bin/sh", "-c", "rm -f /host/etc/cni/net.d/00-multus.conf"]
```
rhrazdil, line 141: This is cool, so we actually clean multus cni from the nodes, instead of leaking the serviceAccount

maiqueb: well, this only handles the cni config, the binary still leaks. Not sure we want to handle that - might as well, since the binary can't be used without the configuration.

maiqueb: @qinqon ^ not sure you want to handle that scenario.

PR #1238, opened 2022-02-23
tests, metrics: Fix potential flakes on metrics scraping test

oshoval:
Since it might take few seconds for the end point to be ready,
we should not assert in the Eventually loop but retry
in case the end point is not ready in the first try.

Mostly seen when the endpoint is behind additional container,
so it takes time until the image of it is pulled.
This change is good as is, because it splits the logic, and use
assert in the right place.

```release-note
None
```

### DIFF
--- a/test/check/check.go
+++ b/test/check/check.go
@@ -707,7 +707,7 @@ func getMonitoringEndpoint() (*corev1.Endpoints, error) {
 	return endpoint, nil
 }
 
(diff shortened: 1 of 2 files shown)


oshoval:
Thanks @RamLavi 
Which branch is the CNAO one of 4.10.1 ?
Once we backport we will need to cherry-pick this PR as well (we can wait still)

PR #1253, opened 2022-03-04
multus: Delete cni cache at preStop

qinqon:
/kind bug

When multus is deleted the CNI resolution has to be forced for already
present pods or they will not be able to be removed. This change remove
the CNI cache before deleting the multus daemonset. This is related to
[1]

[1] https://github.com/kubevirt/cluster-network-addons-operator/pull/1219/commits/07036046c82b1f902a60ae4c058be61612e51f0d

**Release note**:

```release-note
Delete CNI cache /var/lib/cni/* at multus delete, the cache should be reconstructed by the CRI if it's using libcni.
```

### DIFF
--- a/data/multus/001-multus.yaml
+++ b/data/multus/001-multus.yaml
@@ -134,11 +134,13 @@ spec:
               mountPath: /host/etc/cni/net.d
             - name: cnibin
               mountPath: /host/opt/cni/bin
+            - name: cnicache
+              mountPath: /host/var/lib/cni
           imagePullPolicy: {{ .ImagePullPolicy }}
           lifecycle:
             preStop:
               exec:
-                command: ["/bin/sh", "-c", "rm -f /host/etc/cni/net.d/00-multus.conf"]
+                command: ["/bin/sh", "-c", "rm -rf /host/etc/cni/net.d/00-multus.conf /host/var/lib/cni/*"]
       terminationGracePeriodSeconds: 10
       volumes:
         - name: cni
@@ -147,6 +149,9 @@ spec:
         - name: cnibin
           hostPath:
             path: {{ .CNIBinDir }}
+        - name: cnicache
+          hostPath:
+            path: /var/lib/cni
       priorityClassName: system-cluster-critical
       nodeSelector: {{ toYaml .Placement.NodeSelector | nindent 8 }}
       affinity: {{ toYaml .Placement.Affinity | nindent 8 }}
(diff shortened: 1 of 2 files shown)


data/multus/001-multus.yaml
```diff
@@ -134,11 +134,13 @@ spec:
...
             - name: cnibin
               mountPath: /host/opt/cni/bin
+            - name: cnicache
+              mountPath: /host/var/lib/cni
           imagePullPolicy: {{ .ImagePullPolicy }}
           lifecycle:
             preStop:
               exec:
-                command: ["/bin/sh", "-c", "rm -f /host/etc/cni/net.d/00-multus.conf"]
+                command: ["/bin/sh", "-c", "rm -rf /host/etc/cni/net.d/00-multus.conf /host/var/lib/cni/*"]
```
maiqueb, line 143: I think it would be safer to only delete stuff associated w/ multus; I find it strange that you're destroying the cache for all CNIs.

Did you attempt to only remove the stuff belonging to multus ? 

Something like (I did not check if this works, my goal is mainly to show my intention):
```yaml 
command:
  - "/bin/sh"
  - "-c"
  - "find /host/var/lib/cni -name multus* -exec rm -f {} ; && rm -f /host/etc/cni/net.d/00-multus.conf"
```

maiqueb (commented):
I would prefer to only remove the data associated with multus, but I won't object if we have to blow up the entire cni cache.

data/multus/001-multus.yaml
qinqon: I tried it and didn't work, maybe I can try with grep.

qinqon:
/retest
```
+ git clone https://github.com/kubevirt/kubevirtci.git /tmp/cnao/cluster-network-addons-operator/_kubevirtci/
Cloning into '/tmp/cnao/cluster-network-addons-operator/_kubevirtci'...
fatal: error reading section header 'shallow-info'
```

maiqueb (changes requested):
Would you update the release notes indicating that runtimes relying on the libcni's cache will have  to find a way to reconstruct the cache ?

qinqon:
> runtimes relying on the libcni's cache will have to find a way to reconstruct the cache ?

Let's really verify that first.

qinqon:
> > runtimes relying on the libcni's cache will have to find a way to reconstruct the cache ?
> 
> Let's really verify that first.

@maiqueb Have confirmed that removing just the "*multus*" entries is not enough.

maiqueb (approved):
Thanks. 

Awesome work you did figuring this one out.

RamLavi (approved):
/lgtm
/approve

Great catch! Thanks

PR #1260, opened 2022-03-10
ginkgo, vendor, Bump to ginkgo v2

RamLavi:
This PR bumps ginkgo vendor to v2.

**Release note**:

```release-note
NONE
```


automation/check-patch.e2e-lifecycle-k8s.sh
```diff
@@ -37,7 +37,7 @@ main() {
     make cluster-up
     trap teardown EXIT SIGINT SIGTERM SIGSTOP
     make cluster-operator-push
-    make E2E_TEST_EXTRA_ARGS="-ginkgo.noColor --junit-output=$ARTIFACTS/junit.functest.xml" test/e2e/lifecycle
+    make E2E_TEST_EXTRA_ARGS="-ginkgo.noColor --ginkgo.junit-report=$ARTIFACTS/junit.functest.xml" test/e2e/lifecycle
```
qinqon, line 40: Move the $ARTIFACTS to --output-dir, it does not work if the dir is passed as --junit-report

qinqon (changes requested):

automation/check-patch.e2e-lifecycle-k8s.sh
RamLavi: Done

RamLavi:
seem like a vendor fetch hiccup
/retest

RamLavi:
/hold
checking artifacts copied by deliberately failing test

automation/check-patch.e2e-lifecycle-k8s.sh
RamLavi: I checked using `--ginkgo.junit-report=$ARTIFACTS/junit.functest.xml"` (+failing the test on purpose) and we do see the artifacts, see [here](https://gcsweb.ci.kubevirt.io/gcs/kubevirt-prow/pr-logs/pull/kubevirt_cluster-network-addons-operator/1260/pull-e2e-cluster-network-addons-operator-workflow-k8s/1503705101607374848/)

RamLavi:
looks like lifecycle lane [thinks](versionChanged) that version has changed. need to investigate

/hold

RamLavi:
/retest
CNAO operator not being ready does not seem related, investigated on another [PR](https://github.com/kubevirt/cluster-network-addons-operator/pull/1317)

RamLavi:
/hold
only thing to check is the artifacts issue

RamLavi:
/retest
prow is back but still hangs

automation/check-patch.e2e-lifecycle-k8s.sh
oshoval: Interesting that it works, didnt work on other repos such as knmstate, kubevirt
I think it worth still to consider please to indeed split it to junit-report and artifacts as this is the formal way, and this way in case something changes with ginkgo we will be already adapted,
Is there maybe some trap that copies the file ? if so it can explain it, unless the tests runs from the artifacts directory which can also explain why it works

oshoval (commented):
Can you please split to 2 commits, 
one that does the changes, and one with all the vendor changes?

automation/check-patch.e2e-lifecycle-k8s.sh
RamLavi: IIRC your way didn't work...

Makefile
```diff
@@ -56,7 +56,7 @@ $(GO):
 	hack/install-go.sh $(BIN_DIR)
 
 $(GINKGO): $(GO) go.mod
-	GOBIN=$$(pwd)/build/_output/bin/ $(GO) install ./vendor/github.com/onsi/ginkgo/ginkgo
+	GOBIN=$$(pwd)/build/_output/bin/ $(GO) install -mod=mod github.com/onsi/ginkgo/v2/ginkgo@v2.1.3
```
oshoval, line 59: in the go.mod we have other version already
https://github.com/kubevirt/cluster-network-addons-operator/pull/1260/commits/19add50bd4b44488203f38657103872f329607a1#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6R156

i think here it will be better to read the version from the go.mod so it will be auto updated,
you can see how Quique did it on knmstate PR

pkg/apis/networkaddonsoperator/shared/zz_generated.deepcopy.go
```diff
@@ -24,7 +24,7 @@ package shared
 import (
 	configv1 "github.com/openshift/api/config/v1"
 	conditionsv1 "github.com/openshift/custom-resource-status/conditions/v1"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
oshoval, line 27: needed?

tools.go
```diff
@@ -1,10 +1,11 @@
+//go:build tools
```
oshoval, line 1: needed ?
if so does the line below needed as well still ?

oshoval (commented):
Thanks

pkg/apis/networkaddonsoperator/shared/zz_generated.deepcopy.go
RamLavi: not really, this is a tailing fix, since CNAO does not check fmt checks on PRs.. 
If you insist I will remove and put in another PR

tools.go
RamLavi: This line is added by go, now that we use go 1.17 ...

oshoval: maybe need before that something like `go mod tidy -compat=$(GO_VERSION)` ?
weird that we have both lines now

Makefile
RamLavi: > in the go.mod we have other version already [19add50#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6R156](https://github.com/kubevirt/cluster-network-addons-operator/commit/19add50bd4b44488203f38657103872f329607a1#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6R156)

I see what you mean here...

> 
> i think here it will be better to read the version from the go.mod so it will be auto updated, you can see how Quique did it on knmstate PR
> 
> [nmstate/kubernetes-nmstate@fbda01d#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R4](https://github.com/nmstate/kubernetes-nmstate/commit/fbda01dfee7e5930e9e630407b5363d863a55751#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R4)

Not sure it's a fail safe idea.. unlike GO-VERSION which appears only once in the go.mod file, the `github.com/onsi/ginkgo/v2` import can appear multiple times, both on `require` and `replace` section..

oshoval: by taking the last occurrence in the file you will take the last one which will be either replace or the require

RamLavi: I don't want to rely on me knowing if there is a replace, it might be added in the future by someone else.
order of replace/require is not set.. I'll set a short script, tell me what you think

oshoval: maybe there is a `go mod list package` command ? that say what is the bottom line chosen ?

hack/go-version.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash -e
+grep ^go go.mod | awk '{print $2}'
```
oshoval, line 2: missing newline please

hack/ginkgo-version.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash -e
+
+gingko_vendor="github.com/onsi/ginkgo/v2"
+if [ -n "$(grep "=> ${gingko_vendor}" go.mod)" ]; then
+	grep "=> ${gingko_vendor}" go.mod | awk '{print $NF}'
+elif [ -n "$(grep "${gingko_vendor}" go.mod)" ]; then
```
oshoval, line 6: maybe here no need if, just the cmd

oshoval (commented):
gr8

Makefile
RamLavi: did not find such a tool/command

hack/ginkgo-version.sh
RamLavi: I think we need the if else.. if not then both greps might run. I can do an `exit` command but not sure it's any way better..

oshoval: i mean just else

RamLavi: ahh ack

automation/check-patch.e2e-lifecycle-k8s.sh
oshoval: well it worked on all other repos, worth another try if you can please

Makefile
```diff
@@ -56,7 +58,7 @@ $(GO):
 	hack/install-go.sh $(BIN_DIR)
 
 $(GINKGO): $(GO) go.mod
-	GOBIN=$$(pwd)/build/_output/bin/ $(GO) install ./vendor/github.com/onsi/ginkgo/ginkgo
+	GOBIN=$$(pwd)/build/_output/bin/ $(GO) install -mod=mod github.com/onsi/ginkgo/v2/ginkgo@$(GINKGO_VERSION)
```
qinqon, line 61: Pointing to the go-version branch is really needed ?

qinqon: btw using ginkgo CLI here does not brings much value, we can just use go test.

automation/check-patch.e2e-lifecycle-k8s.sh
qinqon: I am pretty sure it was not working if you don't specify `--output-dir`, maybe they have fix it at newer version ?

hack/ginkgo-version.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash -e
+
+gingko_vendor="github.com/onsi/ginkgo/v2"
```
qinqon, line 3: Let's just replace the ginkgo CLI with `go test` so we remove infra.

qinqon (changes requested):

hack/ginkgo-version.sh
RamLavi: ack, removed

automation/check-patch.e2e-lifecycle-k8s.sh
RamLavi: donno. didn't find any reference of such a fix..

oshoval: Please see
https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_cluster-network-addons-operator/1260/pull-e2e-cluster-network-addons-operator-monitoring-k8s/1511616494230835200
```
+ cp '/tmp/cnao//cluster-network-addons-operator/test_logs/e2e/workflow/*.log' /logs/artifacts
cp: cannot stat '/tmp/cnao//cluster-network-addons-operator/test_logs/e2e/workflow/*.log': No such file or directory
```

pkg/apis/networkaddonsoperator/shared/zz_generated.deepcopy.go
oshoval: can you try please to revert this file ? it doesn't seem related to this PR ?

RamLavi:
/hold 
something is wrong.. lifecycle lane only 6 tests??

automation/check-patch.e2e-lifecycle-k8s.sh
RamLavi: >using the right args, can remove the need of cp if related

that cp doesn't have to succeed, it has a `|| true`

>P.S panic: test timed out after 10m0s might also be due to this PR ?

think so..

oshoval: The fact that there is `|| true` doesnt surely mean it is allowed to fail, it was there in case there isn't a file imo.
Since a file is not there, it didnt manage to copy it and the junit failed to be displayed.
It smells like it might be because of the output-dir, this is what we saw at nmstate at least

The junit should exists if the test ran, even if they paniced

RamLavi:
> /hold something is wrong.. lifecycle lane only 6 tests??

fixed on https://github.com/kubevirt/cluster-network-addons-operator/pull/1328 
/unhold

PR #1272, opened 2022-03-16
DeleteOwnedObject, Change Delete handling log location

RamLavi:
Currently the "Handling deletion of X" is printed even if the object is
not owned, This may lead to misleading log data when running on HCO.
Moving the log to after object is filtered for ownership

**Release note**:

```release-note
NONE
```


qinqon (approved):
/lgtm
/approve

PR #1280, opened 2022-03-17
[release 0.65] bump kubevirtci and k8s 1.23

RamLavi:
This PR bumps KUBRVIRTCI_TAG and KUBEVIRT_PROVIDER to the correct k8s provider
Also, from this kubrvirtci bump there is no need to install a specific networkManager on cluster-up, as the version in kubevirtci is good enough.

**Release note**:

```release-note
NONE
```


RamLavi:
/retest
should work now that golang is bumped

RamLavi:
/retest
known flake

RamLavi:
the lifecycle test fails because the upgrade test is trying to deploy old versions, specifically KMP with `admissionregistration.k8s.io/v1beta1` which is AFAIU deprecated on k8s 1.22 and above(see [here](https://github.com/kubernetes/kubernetes/issues/82021))

amm.. any sugestions? @qinqon @rhrazdil

rhrazdil:
@RamLavi  I see, In that case I think it makes sense drop testing of upgrade from versions that can't run on the current k8s cluster.

If we have to support upgrade of those versions, we would need to add another lane with an older k8s cluster, but I don't think that's the case, is it?

RamLavi:
> @RamLavi I see, In that case I think it makes sense drop testing of upgrade from versions that can't run on the current k8s cluster.
> 
> If we have to support upgrade of those versions, we would need to add another lane with an older k8s cluster, but I don't think that's the case, is it?

You're right.
AFAIK on this stable branch we need to support from versions on branch `release-0.53` and above, and this already has `admissionregistration.k8s.io/v1` ([here](https://github.com/kubevirt/cluster-network-addons-operator/blob/v0.53.0/data/kubemacpool/kubemacpool.yaml#L8))

Now that I think of it we actually did something similar on main branch [here](https://github.com/kubevirt/cluster-network-addons-operator/pull/1252/commits/49c4af6858c10937c5abf3fdd8f4cb29991ebeca)

RamLavi:
/override pull-e2e-cnao-nmstate-functests-release-0.65
known flaky

RamLavi:
/override pull-e2e-cnao-nmstate-functests-release-0.65
known flaky, unrelated to PR
@rhrazdil this flaky happens more often lately, could you take a look that it is not a new issue?

PR #1292, opened 2022-03-22
Set degraded if nmstate is requested in networkAddonsConfig

rhrazdil:
This PR aims to block upgrade if still using CNAO to deploy nmstate by Setting NetworkAddonsConfig Degraded Status to true.
HCO will reconcile it and see that the NAC status is degraded, consequently it will set it's status upgradeable to false.

See HCO: https://github.com/kubevirt/hyperconverged-cluster-operator/blob/main/controllers/hyperconverged/hyperconverged_controller.go#L740

That will block upgrade until user installs KNO (Kubernetes Nmstate Operator).

HCO change:
- https://github.com/kubevirt/hyperconverged-cluster-operator/pull/1847

**Release note**:

```release-note
Block CNV upgrade if still using CNAO to deploy nmstate
```

### DIFF
--- a/pkg/controller/statusmanager/status_manager.go
+++ b/pkg/controller/statusmanager/status_manager.go
@@ -158,15 +158,30 @@ func (status *StatusManager) set(reachedAvailableLevel bool, conditions ...condi
 			},
 		)
 	} else if reachedAvailableLevel {
-		// If successfully deployed all components and is not failing on anything, mark as Available
-		status.eventEmitter.EmitAvailableForConfig()
-		conditionsv1.SetStatusConditionNoHeartbeat(&config.Status.Conditions,
-			conditionsv1.Condition{
-				Type:   conditionsv1.ConditionAvailable,
-				Status: corev1.ConditionTrue,
-			},
-		)
-		config.Status.ObservedVersion = operatorVersion
+		if config.Spec.NMState != nil {
+			// CNAO doesn't support nmstate deployment anymore, set Degraded state is nmstate is requested in NetworkAddonsConfig
+			reason := "InvalidConfiguration"
+			message := "NMState deployment is not supported by CNAO anymore, please install Kubernetes NMState Operator"
+			status.eventEmitter.EmitFailingForConfig(reason, message)
+			conditionsv1.SetStatusConditionNoHeartbeat(&config.Status.Conditions,
+				conditionsv1.Condition{
+					Type:    conditionsv1.ConditionDegraded,
+					Status:  corev1.ConditionTrue,
+					Reason:  reason,
+					Message: message,
+				},
+			)
+		} else {
+			// If successfully deployed all components and is not failing on anything, mark as Available
+			status.eventEmitter.EmitAvailableForConfig()
+			conditionsv1.SetStatusConditionNoHeartbeat(&config.Status.Conditions,
+				conditionsv1.Condition{
+					Type:   conditionsv1.ConditionAvailable,
+					Status: corev1.ConditionTrue,
+				},
+			)
+			config.Status.ObservedVersion = operatorVersion
+		}
 	}
 
 	// Make sure to expose deployed containers
--- a/test/e2e/workflow/recovery_test.go
+++ b/test/e2e/workflow/recovery_test.go
@@ -33,7 +33,7 @@ var _ = Describe("NetworkAddonsConfig", func() {
 
 			It("should turn from Failing to Available", func() {
 				CheckAvailableEvent(gvk)
-				CheckConfigCondition(gvk, ConditionAvailable, ConditionTrue, 5*time.Second, CheckDoNotRepeat)
+				CheckConfigCondition(gvk, ConditionAvailable, ConditionTrue, 10*time.Second, CheckDoNotRepeat)
 				CheckConfigCondition(gvk, ConditionDegraded, ConditionFalse, CheckImmediately, CheckDoNotRepeat)
 			})
 		})


rhrazdil:
Hello @tiraboschi @qinqon @RamLavi 
Could you please review this PR?

pkg/controller/statusmanager/status_manager.go
```diff
@@ -158,15 +159,32 @@ func (status *StatusManager) set(reachedAvailableLevel bool, conditions ...condi
...
-		)
-		config.Status.ObservedVersion = operatorVersion
+		if status.isRunningOnOpenshift411OrLater() && config.Spec.NMState != nil {
+			// In case the clusterVersion.Status.Desired.Version is 4.11 and we're deploying nmstate
+			// with CNAO, we'll set degraded state to block CNV upgrade, forcing migration of
+			// nmstate to kubernetes nmstate operator
+			reason := "InvalidConfiguration"
+			message := "NMState deployment is not provided in CNV >= 4.11, please install Kubernetes NMState Operator"
+			status.eventEmitter.EmitFailingForConfig(reason, message)
+			conditionsv1.SetStatusConditionNoHeartbeat(&config.Status.Conditions,
```
qinqon, line 169: This should change heartbeat right ? it's not a noop

pkg/controller/statusmanager/status_manager.go
```diff
@@ -158,15 +159,32 @@ func (status *StatusManager) set(reachedAvailableLevel bool, conditions ...condi
...
-		// If successfully deployed all components and is not failing on anything, mark as Available
-		status.eventEmitter.EmitAvailableForConfig()
-		conditionsv1.SetStatusConditionNoHeartbeat(&config.Status.Conditions,
-			conditionsv1.Condition{
-				Type:   conditionsv1.ConditionAvailable,
-				Status: corev1.ConditionTrue,
-			},
-		)
-		config.Status.ObservedVersion = operatorVersion
+		if status.isRunningOnOpenshift411OrLater() && config.Spec.NMState != nil {
```
qinqon, line 162: Maybe we don't need to check openshift at all at the end at this u/s release for wherever is running we will fail on invalid networkaddonsconfig meaning that if they ask for NMState we fail, openshift kubernetes or whatever.

pkg/controller/statusmanager/status_manager.go
```diff
@@ -158,15 +159,32 @@ func (status *StatusManager) set(reachedAvailableLevel bool, conditions ...condi
...
+					Type:    conditionsv1.ConditionDegraded,
+					Status:  corev1.ConditionTrue,
+					Reason:  reason,
+					Message: message,
+				},
+			)
+		} else {
+			// If successfully deployed all components and is not failing on anything, mark as Available
+			status.eventEmitter.EmitAvailableForConfig()
+			conditionsv1.SetStatusConditionNoHeartbeat(&config.Status.Conditions,
```
qinqon, line 180: We should change heratbeat if previous condition was not success, although I don't know if we can check that or if it's worth it.

qinqon (changes requested):
I think we don't need to check openshift.

pkg/controller/statusmanager/status_manager.go
rhrazdil: Right correct

pkg/controller/statusmanager/status_manager.go
rhrazdil: The thing is, if we don't check OCP 4.11, then we block not only upgrade from CNV 4.10 to 4.11, but also all z-stream upgrades on 4.10

rhrazdil: However, we could notify u/s users as you suggest

pkg/controller/statusmanager/status_manager.go
RamLavi: imo if we don't keep track about the heartbeat anymore then there's no need to update it in any scenario - neither good or bad.
We do have the TransitionTimestamp (when the status last changed value) which might be enough..

rhrazdil: Ah, so we don't want to update it in https://github.com/kubevirt/cluster-network-addons-operator/pull/1292#discussion_r839518033 either, correct?

qinqon: Then feels like heartbeat is kind of obsolete and replaced by TransitionTimestamp ?

pkg/controller/statusmanager/status_manager.go
rhrazdil: IIUC, we have to backport this to 4.10. If we didn't backport this a have this code in 4.11, then by the time users upgrade OCP to 4.11 and CNV to 4.11, it's already too late.

rhrazdil:
After discussing with @qinqon I'm setting the Degraded condition always when nmstate is requested.
In a backport to release-0.65, I only do so if we find we're running on OCP 4.11
https://github.com/kubevirt/cluster-network-addons-operator/pull/1323

pkg/controller/statusmanager/status_manager.go
RamLavi: it doesn't replace `TransitionTimestamp` becasue `TransitionTimestamp` is already there.
heartbeat is no longer published on CNAO. This is because it also causes a resourceVersion change (every 1 minute) and that triggers a repeating reconcile on HCO. As a result, we decided to remove it entirely on main and release-0.65 branches.

PR #1300, opened 2022-03-24
crds: sort tlsProfiles to generate deterministic crd

rhrazdil:

Sort tlsProfiles alphabetically, so that generated CRD is always the same

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/components/components.go
+++ b/pkg/components/components.go
@@ -588,8 +588,8 @@ func GetCrd() *extv1.CustomResourceDefinition {
 
 	cipherSuites := func() []extv1.JSON {
 		suites := []extv1.JSON{}
-		for _, p := range ocpv1.TLSProfiles {
-			for _, c := range p.Ciphers {
+		for _, p := range tlsProfiles(ocpv1.TLSProfiles).sortedKeys() {
+			for _, c := range ocpv1.TLSProfiles[p].Ciphers {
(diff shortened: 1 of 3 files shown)


pkg/components/components.go
```diff
@@ -588,8 +589,13 @@ func GetCrd() *extv1.CustomResourceDefinition {
 
 	cipherSuites := func() []extv1.JSON {
 		suites := []extv1.JSON{}
-		for _, p := range ocpv1.TLSProfiles {
-			for _, c := range p.Ciphers {
+		tlsProfiles := []ocpv1.TLSProfileType{}
+		for k := range ocpv1.TLSProfiles {
+			tlsProfiles = append(tlsProfiles, k)
+		}
+		sort.Sort(byTLSProfileName(tlsProfiles))
```
oshoval, line 596: can't we use native `sort.Strings` ?

you can implement also the sort method for the map itself,
so it will extract the keys and then sort them

rhrazdil: > can't we use native sort.Strings on a string slice ?

But for that I would need to cast the keys to `string`s and then to `ocpv1.TLSProfileType` when using them for indexing, which seems cumbersome.

See the [force-pushed](https://github.com/kubevirt/cluster-network-addons-operator/compare/20aa7e5889cbd322bdc0564529d853d5133f8b24..2f213ee0e96c03a2dbdae9f49e93c344396cb0ac) update. Also added unit tests.

RamLavi (approved):
/lgtm

PR #1302, opened 2022-03-24
[release-0.65] Add alert to notify of nmstate removal

rhrazdil:
kubernetes-nmstate is removed in the next CNAO version.
This change adds an alert that notifies users who have
knmstate deployed with CNAO, and point them to runbook,
that explains that standalone kubernetes-nmstate operator
should be installed.

**Release note**:

```release-note

```

### DIFF
--- a/data/monitoring/prom-rule.yaml
+++ b/data/monitoring/prom-rule.yaml
@@ -11,6 +11,20 @@ spec:
       rules:
         - expr: sum(up{namespace='{{ .Namespace }}', pod=~'cluster-network-addons-operator-.*'} or vector(0))
           record: kubevirt_cnao_num_up_operators
+        - expr: sum(kube_deployment_labels{deployment=~'kubernetes-nmstate-operator'} or vector(0))
+          record: kubevirt_nmstate_operator_deployments
+        - expr: sum(kube_daemonset_labels{namespace='{{ .Namespace }}', daemonset=~'nmstate-handler'} or vector(0))
+          record: kubevirt_cnao_nmstate_deployed
+        - alert: CnaoNmstateMigration
+          annotations:
+            summary: Nmstate will be removed from CNAO.
+            runbook_url: http://kubevirt.io/monitoring/runbooks/CnaoNmstateMigration
+          expr: sum(kubevirt_cnao_nmstate_deployed or vector(0)) > 0 and sum(kubevirt_nmstate_operator_deployments or vector(0)) == 0
+          for: 5m
+          labels:
+            severity: warning
+            kubernetes_operator_part_of: kubevirt
+            kubernetes_operator_component: cluster-network-addons-operator
         - alert: CnaoDown
           annotations:
             summary: CNAO pod is down.
--- a/data/monitoring/prom-rule.yaml
+++ b/data/monitoring/prom-rule.yaml
@@ -23,6 +23,7 @@ spec:
           for: 5m
           labels:
             severity: warning
+            priority: high
             kubernetes_operator_part_of: kubevirt
             kubernetes_operator_component: cluster-network-addons-operator
         - alert: CnaoDown
--- a/test/e2e/workflow/deployment_test.go
+++ b/test/e2e/workflow/deployment_test.go
@@ -431,8 +431,9 @@ var _ = Describe("NetworkAddonsConfig", func() {
 					Eventually(func() error {
 						nmstateHandlerDaemonSet := &v1.DaemonSet{}
 						return testenv.Client.Get(context.TODO(), types.NamespacedName{Name: NMStateComponent.DaemonSets[0], Namespace: "nmstate"}, nmstateHandlerDaemonSet)
-					}, 5*time.Minute, time.Second).Should(BeNil(), fmt.Sprintf("Timed out waiting for nmstate-operator daemonset"))
+					}, 5*time.Minute, time.Second).Should(BeNil(), "Timed out waiting for nmstate-operator daemonset")
 				})
+
 			})
 			Context("when it is not already deployed", func() {
 				BeforeEach(func() {
--- a/Makefile
+++ b/Makefile
@@ -36,7 +36,8 @@ E2E_TEST_EXTRA_ARGS ?=
 E2E_TEST_ARGS ?= $(strip -test.v -test.timeout $(E2E_TEST_TIMEOUT) -ginkgo.v $(E2E_TEST_EXTRA_ARGS))
 E2E_SUITES = \
 	test/e2e/lifecycle \
-	test/e2e/workflow
+	test/e2e/workflow \
+	test/e2e/monitoring
 
 BIN_DIR = $(CURDIR)/build/_output/bin/
 export GOROOT=$(BIN_DIR)/go/
--- a/automation/check-patch.e2e-monitoring-k8s.sh
+++ b/automation/check-patch.e2e-monitoring-k8s.sh
@@ -1,0 +1,31 @@
+#!/bin/bash -xe
+
+# This script should be able to execute workflow functional tests against
+# Kubernetes cluster on any environment with basic dependencies listed in
+# check-patch.packages installed and docker running.
+#
+# yum -y install automation/check-patch.packages
+# automation/check-patch.e2e-workflow-k8s.sh
+
+teardown() {
+    # Don't fail if there is no logs
+    cp ${E2E_LOGS}/workflow/*.log ${ARTIFACTS} || true
+    make cluster-down
+}
+
+main() {
+    export KUBEVIRT_PROVIDER='k8s-1.23'
+
+    source automation/check-patch.setup.sh
+    cd ${TMP_PROJECT_PATH}
+
+    export MONITORING_NAMESPACE="monitoring"
+    make cluster-down
+    make cluster-up
+    trap teardown EXIT SIGINT SIGTERM
+    make cluster-operator-push
+    make cluster-operator-install
+    make E2E_TEST_EXTRA_ARGS="-ginkgo.noColor --junit-output=$ARTIFACTS/junit.functest.xml" test/e2e/monitoring
+}
+
+[[ "${BASH_SOURCE[0]}" == "$0" ]] && main "$@"
--- a/cluster/up.sh
+++ b/cluster/up.sh
@@ -20,6 +20,8 @@ SCRIPTS_PATH="$(dirname "$(realpath "$0")")"
 source ${SCRIPTS_PATH}/cluster.sh
 
 export KUBEVIRT_DEPLOY_PROMETHEUS=true
+export KUBEVIRT_DEPLOY_PROMETHEUS_ALERTMANAGER=true
+export KUBEVIRT_DEPLOY_GRAFANA=true
 
 cluster::install
 
--- a/go.mod
+++ b/go.mod
@@ -28,6 +28,7 @@ require (
 	github.com/operator-framework/operator-sdk v0.18.0
 	github.com/pkg/errors v0.9.1
 	github.com/prometheus/client_golang v1.11.0
+	github.com/prometheus/common v0.28.0
 	github.com/spf13/pflag v1.0.5
 	github.com/thanhpk/randstr v1.0.4
 	github.com/tomnomnom/linkheader v0.0.0-20180905144013-02ca5825eb80 // indirect
@@ -126,8 +127,9 @@ require (
 	github.com/jmoiron/sqlx v1.2.0 // indirect
 	github.com/joho/godotenv v1.3.0 // indirect
 	github.com/josharian/intern v1.0.0 // indirect
+	github.com/jpillora/backoff v1.0.0 // indirect
 	github.com/json-iterator/go v1.1.12 // indirect
-	github.com/kevinburke/ssh_config v0.0.0-20190725054713-01f96b0aa0cd // indirect
+	github.com/kevinburke/ssh_config v0.0.0-20201106050909-4977a11b4351 // indirect
 	github.com/lann/builder v0.0.0-20180802200727-47ae307949d0 // indirect
 	github.com/lann/ps v0.0.0-20150810152359-62de8c46ede0 // indirect
 	github.com/lib/pq v1.3.0 // indirect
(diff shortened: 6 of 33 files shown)


RamLavi:
@sradco can you also review this ?

RamLavi (changes requested):
Can you also add an e2e to alerts_tests?

test/e2e/workflow/deployment_test.go
```diff
@@ -408,12 +408,33 @@ var _ = Describe("NetworkAddonsConfig", func() {
 		var (
 			configSpec cnao.NetworkAddonsConfigSpec
 		)
+		checkMetricValues := func(expectedMetricValueMap map[string]string) {
```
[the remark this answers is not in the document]
rhrazdil: It also means adding the monitoring lane, right?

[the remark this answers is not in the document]
rhrazdil: Done

data/monitoring/prom-rule.yaml
```diff
@@ -11,6 +11,20 @@ spec:
...
+        - alert: CnaoNmstateMigration
+          annotations:
+            summary: Nmstate will be removed from CNAO.
+            runbook_url: http://kubevirt.io/monitoring/runbooks/CnaoNmstateMigration
+          expr: sum(kubevirt_cnao_nmstate_deployed or vector(0)) > 0 and sum(kubevirt_nmstate_operator_deployments or vector(0)) == 0
+          for: 5m
+          labels:
+            severity: warning
+            kubernetes_operator_part_of: kubevirt
+            kubernetes_operator_component: cluster-network-addons-operator
```
yaacov, line 27: we want to highlight this alert in the UI virtualization overview page.

can we add a label that will tell the UI to show this alert on the top of the virtualization overview page for better visibility ?

for exampe:
`kubevirt_ui: highlighted`
does that makes sense?

yaacov: do we have a naming convention for alert labels?

rhrazdil: I don't think we have, but @sradco may confirm that.

rhrazdil: I would suggest

```
kubevirt_console_scope: [cluster,virtualmachine]
```
But IIUC, we need to provide more info in case an alert is cluster scoped (that's not case of the alert in this PR, just to make the mechanism general).

sradco: @yaacov Can you please explain what is the reasoning to highlight this specific alert? 
I believe alerts should be ordered based on their severity.

yaacov: `kubevirt_console_scope: [cluster,virtualmachine]`
is useful to specify the scope,
we will also need a way to specify that we want this alert to be shown in the top of the overview page.
`kubevirt_console_popup: [cluster,virtualmachine]`
?

rhrazdil: @sradco We want to show this specific alert in OCP Console UI as a banner on CNV overview page

rhrazdil: @yaacov My thinking was that is the kubevirt_console_scope is define, then we can implicitly assume that we want to show that alert in UI, no? Otherwise, if it's not there, UI wouldn't show it

sradco: @yaacov please see my question above. I don;t understand the need for this.
Having these "console" details in the alert as labels will be wrong practice. 
If you want to do this, it should be in the UI level only.
The alerts can be used is multiple envs, not only OCP UI and are also sent to the user with alert manager with all its labels. 
Adding this will create clutter that is not relevant to the alert itself.
The new labels we added about the operator name and component are directly related to the alert details and help the users to understand where the issue comes from.

yaacov: @sradco hi,

> the reasoning to highlight this specific alert?

IIUC the request is that an alert be shown on OCP Console UI as a banner on CNV overview page.
UI need some hint that this alert request to be shown this way.
( all alerts including this one will be shown in the regular alerts list as well. )

> If you want to do this, it should be in the UI level only.

AFAIU this is a property of the alert, the UI read this hint, and decide to show alerts that has this property differently.
this alert can then be updated by the people who create the alert rule, and decide to not show it if not needed anymore, or create a different alert that request to be displayed on top of the page.

It's an independent property from the severity, for example an alert can by warning or information and request to be highlighted.

yaacov: @sradco
do we currently have a way to know that an alert belong to a specific virtual machine?

AFAIU the request was that this alert will be shown in the virtualization overview, and another for a specific vm that needs some action because of some event.

sradco: I still think this should be done on the UI level. 
Perhaps there could be a place to state the alerts to show in the banner that is separate from the alerts . Then the UI can check based on it.
I strongly believe we should not  add these labels to the alerts.

sradco: We don't have alerts for specific VMs at this point. There are planned alerts for 4.12.

ronensdeor: @sradco We need a way to surface specific alerts so they are not "lost" in all the other alerts.
Those alert may have different severity, like a reboot is required, which can be warning, but the user must be aware of it.
The the current case, users that are using NMstate from CNV must install the stand alone operator before moving to 4.11. In this case it's not a critical alert the requires immediate action, but it is an important one. Having it as a warning will probably not be visible enough.

rhrazdil: @sradco We're looking for a generic way of displaying an alert in UI (not in Observe/metrics).
Marking an alert with an annoration/label/something allows leaving all logic for the backend, and UI only presents
what it's supposed to.
Wiring logic into UI, what is supposed to be displayed, what not, is very sub-optimal.

> I strongly believe we should not add these labels to the alerts.
Could you explain please?

I believe having an extra annotation would be a huge clutter for the users.
At the very least, the gain we get - a user being notified on a relevant place in the UI, instead of somewhere in Monitoring category, seems to be prevalent to me

yaacov: > Perhaps there could be a place to state the alerts to show in the banner that is separate from the alerts.

Also works for me, my concern is that maintainers that will update the alert rules forget to update the alert rules "impact to user" table that will be in a different file.
It will also be easier to a user looking at the active alert to filter by "impact to user", in case they don't use the UI.

sradco: In that case I think we should not use UI in the label name.
It should be somehow reflected that an action is required or something like that.
Like a sub label to severity.

yaacov: > It should be somehow reflected that an action is required

`admin_action_recommended: <some longer description>` 
or 
`recommended_action: yes/[no]` 

does that makes more sense?

sradco: I thought of something like:
severity_score / severity_level / severity_priority with  a few optional values, but not sure what they should be called.

yaacov: > Like a sub label to severity.

`severity_priority` makes sense to me because it indicate that handling this alert requires higher priority even if it's currently only a warning.

> What will happen if there are multiple ones?

They will all show in the top of the page, if they all require user to be aware of them, without raising them to critical and without hiding theme between all the other warnings and information alerts.

yaacov: maybe event just `priority` , without the `sevirity_`  prefix ?

data/monitoring/prom-rule.yaml
```diff
@@ -11,6 +11,20 @@ spec:
...
         - alert: CnaoDown
           annotations:
             summary: CNAO pod is down.
```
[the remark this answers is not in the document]
rhrazdil: We want to highlight the CnaoNmstateMigration alert, but yes, I get your point

data/monitoring/prom-rule.yaml
yaacov: we will also need to document how to use the labels, we don't want every critical alert to also be high priority (because it's critical)

sradco: I found a nice article that supports the naming and values for the label:
https://www.guru99.com/defect-severity-in-software-testing.html

I think we should keep it as priority.

sradco: The documentation for the label should be part of OCP metrics documentation.
Im not sure if we should have it as mandatory as the severity. 
The issue is that the label naming impacts also OCP and all other operators.
We should propose this in the monitoring forum and get their opinion.

yaacov: > I think we should keep it as priority.

makes sense to me too

> Im not sure if we should have it as mandatory as the severity.

IMHO it should not be mandatory (default to medium if missing) 
my reasoning is that unlike severity that is clear to define by the rule maintainer, priority may depend on things outside the scope of the alert rules maintainer, the cases where the rule maintainer wish to raise/lower priority are the odd cases.

sradco: I created https://github.com/openshift/enhancements/pull/1077. Please review.

yaacov: from https://github.com/openshift/enhancements/pull/1077 is looks like "priority: high"

@rhrazdil can you add the "priority" label?

phoracek:
Adding explicit hold so we don't merge this preemptively

/hold

data/monitoring/prom-rule.yaml
rhrazdil: Sorry, I've missed your comment.
Added the `priority: high` label in a new commit

data/monitoring/prom-rule.yaml
```diff
@@ -22,6 +22,7 @@ spec:
           expr: sum(kubevirt_cnao_nmstate_deployed or vector(0)) > 0 and sum(kubevirt_nmstate_operator_deployments or vector(0)) == 0
           for: 5m
           labels:
+            priority: high
```
sradco, line 25: @rhrazdil Please move the Priority label below the severity.

yaacov: IMHO semantically, the required fields should go before the optional ones,
using that way of thinking, `severity` that is  required label (and more important) will go before the optional `priority`.
I have nothing against going with alphabetical order if it's the common practice for labels, so that is also good option, don't know what is best.

sradco: @rhrazdil I agree with Kobi.

rhrazdil: Okay, I take it. [Updated](https://github.com/kubevirt/cluster-network-addons-operator/compare/46cadb76d39bfc90514006a71397e8553100edef..faf97e0175d2f1fa3bd071b60a3f44fcccd668ca)

pkg/monitoring/metrics.go
```diff
@@ -33,10 +33,15 @@ var (
 			Name: "kubevirt_cnao_cr_kubemacpool_deployed",
 			Help: "Kubemacpool is deployed by Cnao CR",
 		})
+	nmstateHandlerDeployed = prometheus.NewGauge(
+		prometheus.GaugeOpts{
+			Name: "kubevirt_cnao_nmstate_deployed",
```
RamLavi, line 38: hmmm.. I don't get something..
if you already defined `kubevirt_cnao_nmstate_deployed` with `kube_daemonset_labels`, i.e. without the use of a GAUGE variable, why do you need to define one here? isn't this redundant?

RamLavi (changes requested):
Please check if we really need to define `kubevirt_cnao_nmstate_deployed` as gauge, if you can define it with pure prometheus given metrics (like it seems you already have succeeded), then we can remove the gauges, which will allow us to remove some of the cherry picks which will make things simpler IMO)

pkg/monitoring/metrics.go
rhrazdil: Removed the variable in this [force-push](https://github.com/kubevirt/cluster-network-addons-operator/compare/faf97e0175d2f1fa3bd071b60a3f44fcccd668ca..7cc8bcd687897940a33fddda5cb0a38b1f71031a).

rhrazdil:
> Please check if we really need to define kubevirt_cnao_nmstate_deployed as gauge, if you can define it with pure prometheus given metrics (like it seems you already have succeeded), then we can remove the gauges, which will allow us to remove some of the cherry picks which will make things simpler IMO)

Updated (removed the redundant variable).
To the cherry-picks, the monitoring lane adds infra for testing alerts, we could add the infra alone without adding the lane, but it seems easier from maintenance perspective to cherry-pick the whole commit.
The following two commits:
[monitoring, e2e, Move endpoints test to monitoring lane ](https://github.com/kubevirt/cluster-network-addons-operator/pull/1302/commits/2ce50185e827c006c24403bc9e60246f5fb7e37b) 
[tests, metrics: Fix potential flakes on metrics scraping test](https://github.com/kubevirt/cluster-network-addons-operator/pull/1302/commits/b0cd67ef4ca55999e4585f96325c9bd15f8faf45)

Are not really neccessary, we could keep the tests in workflow/deployment, if you prefer

yaacov:
@rhrazdil hi,
should we merge the alert in the main branch too ?
it makes sense to me to have fixes in main when possible ? 

@RamLavi WDYT ?

cc:// @sradco @phoracek

rhrazdil:
Hello @yaacov,
we decided to only throw the alert in the release-0.65 branch, because cnao in newer release deletes kubernetes-nmstate, if it's installed.
So the alert couldn't fire in newer branch anyway.

sradco:
@rhrazdil I think it is important to add this alerts also to main branch. In case the user upgrade directly to the next majot version and not to the next minor z-stream.

rhrazdil:
We're blocking an upgrade when user upgrades from CNV 4.10 to CNV 4.11 and has kubernetes-nmstate deployed
via CNAO.
Hence, if user has CNV 4.11 (and thus also ocp 4.11) and knmstate deployed, we can safely assume, that it's deployed via Kubernetes Nmstate Operator. CNAO would delete the knmstate deployment within a minute.

So this alert has no meaning on 4.11

PR #1303, opened 2022-03-27
alerts, Add new tests

RamLavi:
This PR Add tests to check KubemacpoolDown and also a check that no alerts are
firing for no reason when all components are deployed

**Release note**:

```release-note
NONE
```


RamLavi:
/hold
anti KMP affinity is not working properly..

test/e2e/monitoring/alerts_test.go
```diff
@@ -54,6 +116,11 @@ var _ = Context("Prometheus Alerts", func() {
 			}
 		})
 
+		It("configuration: role-binding should point to the prometheus serviceAccount", func() {
```
RamLavi, line 119: The reason this test is added is because I had times where I ran the test manually and forgot to set the `MONITORING_NAMESPACE` env var + reset the CNAO manifests. This results in it being defaulted to the d/s value, and the tests to fail for "unexplained" reason.

rhrazdil (commented):
/lgtm

PR #1312, opened 2022-03-29
tls: Use intermediate TLS profile by default

qinqon:
/kind bug

When no TLS is specified at openshift the Intermediate profile is
selected [1]. This changes copy that behaviour at CNAO.

[1] https://github.com/openshift/library-go/blob/master/pkg/operator/configobserver/apiserver/observe_tlssecurityprofile.go#L81-L84

Closes https://bugzilla.redhat.com/show_bug.cgi?id=2069388

**Release note**:

```release-note
Use intermediate TLS profile by default
```


rhrazdil:
Thanks for taking this
/lgtm
/approve

qinqon:
/override pull-e2e-cluster-network-addons-operator-ovs-cni-functests

PR #1330, opened 2022-04-13
Add metrics from data/monitoring/prom-rule.yaml to docs

machadovilaca:

Metrics from `data/monitoring/prom-rule.yaml` are missing from metrics documentation.

**Release note**:

```release-note
None
```


rhrazdil (commented):
Hello, thanks for your PR.
Unfortunately, I'm not sure why adding non empty labels/annotations breaks the alert.
According to the docs, recording rules may not have annotations
https://prometheus.io/docs/prometheus/latest/configuration/recording_rules/#rule
but only labels.

I'm not however sure, what are the consequences (if any) of adding  labels to the rules.
Maybe the rule is now a subject to some filtering?

data/monitoring/prom-rule.yaml
```diff
@@ -11,6 +11,9 @@ spec:
       rules:
         - expr: sum(up{namespace='{{ .Namespace }}', pod=~'cluster-network-addons-operator-.*'} or vector(0))
           record: kubevirt_cnao_num_up_operators
+          annotations:
+            summary: Total count of running CNAO operators
+            type: Gauge
```
rhrazdil, line 16: If I understand correctly, this is a recording rule and not really a gauge.
If you look at `pkg/monitoring/metrics.go`
You'll we have two gauges there:
- kubevirt_cnao_cr_ready
- kubevirt_cnao_cr_kubemacpool_deployed
For that reason, I don't think we should label these records as gauges.
IIUC, these are metrics, not gauges.

data/monitoring/prom-rule.yaml
```diff
@@ -11,6 +11,9 @@ spec:
       rules:
         - expr: sum(up{namespace='{{ .Namespace }}', pod=~'cluster-network-addons-operator-.*'} or vector(0))
           record: kubevirt_cnao_num_up_operators
+          annotations:
+            summary: Total count of running CNAO operators
```
rhrazdil, line 15: Similarly, I'm not sure if using labels is the right thing for documentation strings.
IIUC, labels should be values of similar meaning, that one may use for filtering.
Surely, nobody will ever filter the record by the content of summary.

rhrazdil (changes requested):
Now to the reason why the alerts don't work.
It appears that when you add labels to a record, the alert that uses that record in `expr:`, we can't use the record name as a simple vector value anymore.

So for example, you would need to update
```
name: KubemacpoolDown
expr: kubevirt_cnao_cr_kubemacpool_deployed_total == 1 and kubevirt_cnao_kubemacpool_manager_num_up_pods_total == 0
```
To this
```
name: KubemacpoolDown
expr: sum(kubevirt_cnao_cr_kubemacpool_deployed_total) == 1 and sum(kubevirt_cnao_kubemacpool_manager_num_up_pods_total) == 0
```

I'm not sure I'm 100% correct on the terminology, I'm not expeirenced with Prometheus.

Anyway, as I explain in the comments inline, I don't think we should be adding labels for documentation purposes.

tools/metricsdocs/metricsdocs.go
```diff
@@ -4,13 +4,13 @@ import (
 	"bytes"
 	"fmt"
 	"log"
+	"sigs.k8s.io/controller-tools/pkg/markers"
```
rhrazdil, line 7: can we move this to a separate category?

tools/metricsdocs/metricsdocs.go
```diff
@@ -49,33 +49,61 @@ func metricsOptsToMetricList(metrics map[monitoring.MetricsKey]monitoring.Metric
...
+			Name  string      `yaml:"name"`
+			Rules []yaml.Node `yaml:"rules"`
+		} `yaml:"groups"`
+	} `yaml:"spec"`
+}
+
+type Rule struct {
+	Expr        string `yaml:"expr"`
+	Record      string `yaml:"record,omitempty"`
+	Alert       string `yaml:"alert,omitempty"`
```
rhrazdil, line 64: It seems to me that we only really make use of Record field. Can we just omit the remaining fields?

tools/metricsdocs/metricsdocs.go
```diff
@@ -84,6 +112,22 @@ func readFromPrometheusCR() metricList {
...
+		return "", ""
+	}
+
+	defn := reg.Lookup(ruleNode.HeadComment, markers.DescribesPackage)
+	rawComment, err := defn.Parse(ruleNode.HeadComment)
+	if err != nil {
+		return "", ""
+	}
+
+	comment := rawComment.(Comment)
```
rhrazdil, line 126: Can the assertion fail? Could you check that the type assertion succeeded? This function should return an error if the assertion doesn't succeed.

data/monitoring/prom-rule.yaml
```diff
@@ -9,11 +9,9 @@ spec:
   groups:
     - name: kubevirt.cnao.rules
       rules:
+        # +help:summary="Total count of running CNAO operators",type=Gauge
```
rhrazdil, line 12: Please, see comment in my previous review, I believe I have mentioned that I think this is really not a gauge, but a recording rule

hack/prom-rule-ci/prom-rules-tests.yaml
```diff
@@ -6,7 +6,7 @@ group_eval_order:
   - kubevirt.cnao.rules
 
 tests:
-# CnaoDown positive tests
```
rhrazdil, line 9: Could we just remove the commit that this commit reverts?

rhrazdil (changes requested):

tools/metricsdocs/metricsdocs.go
```diff
@@ -133,7 +171,22 @@ func metricDescriptionToMetric(rrd monitoring.MetricsOpts) metric {
 
 func (m metric) writeOut() {
 	fmt.Println("###", m.name)
-	fmt.Println(m.description + ". Type: " + m.mType + ".")
+
+	writeNewLine := false
+
+	if m.description != "" {
+		fmt.Print(m.description + ". ")
+		writeNewLine = true
```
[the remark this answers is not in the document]
machadovilaca: If using defer it will print two new lines when we have the description and the type. And we only want to print one.

tools/metricsdocs/metricsdocs.go
machadovilaca: done

tools/metricsdocs/metricsdocs.go
machadovilaca: done

tools/metricsdocs/metricsdocs.go
machadovilaca: check added

hack/prom-rule-ci/prom-rules-tests.yaml
machadovilaca: done

data/monitoring/prom-rule.yaml
machadovilaca: Yes, it is a recording rule, but in terms of monitoring and user documentation, it is more useful to understand the type of metric created through the use of the recording rules. Because each of those metrics types will have different characteristics and behaviors

machadovilaca:
@rhrazdil The `pull-e2e-cluster-network-addons-operator-workflow-k8s` keeps failing but I am pretty sure it is unrelated to this PR, would you be able to take a quick look, please?

rhrazdil (approved):
/lgtm

data/monitoring/prom-rule.yaml
```diff
@@ -43,8 +45,10 @@ spec:
             severity: warning
             kubernetes_operator_part_of: kubevirt
             kubernetes_operator_component: cluster-network-addons-operator
+        # +help:summary="Total count of running KubeMacPool manager pods",type=Gauge
         - expr: sum(up{namespace='{{ .Namespace }}', pod=~'kubemacpool-mac-controller-manager-.*'} or vector(0))
           record: kubevirt_cnao_kubemacpool_manager_num_up_pods_total
+        # +help:summary="Total count of deployed KubeMacPool CRs",type=Gauge
```
RamLavi, line 51: This is more accurate
```suggestion
        # +help:summary="Total count of KubeMacPool manager pods deployed by CNAO CR",type=Gauge
```

docs/metrics.md
```diff
@@ -6,7 +6,15 @@ All metrics documented here are auto-generated by the utility tool 'tools/metric
 ## Cluster Network Addons Operator Metrics List
 ### kubevirt_cnao_cr_kubemacpool_deployed
 KubeMacpool is deployed by CNAO CR. Type: Gauge.
+### kubevirt_cnao_cr_kubemacpool_deployed_total
+Total count of deployed KubeMacPool CRs. Type: Gauge.
```
RamLavi, line 10: need to change accordingly of course

RamLavi (changes requested):

data/monitoring/prom-rule.yaml
machadovilaca: updated

docs/metrics.md
machadovilaca: updated

PR #1338, opened 2022-04-22
Remove trailing spaces from CRD

phoracek:

With the current code, generated CRD would contain trailing spaces and
that would fail release unit tests lane.

**Release note**:

```release-note
NONE
```


rhrazdil (approved):
Thanks!

rhrazdil:
How come it didn't fail unit test before? IIUC, it should run make check as well

phoracek:
Aha! It may be because of humans run `make` before pushing, which in turn removes whitespace https://github.com/kubevirt/cluster-network-addons-operator/blob/main/Makefile#L80. The bot drafting the release PR is not calling that?

PR #1339, opened 2022-05-04
e2e, nmstate: Bump nmstate version in e2e tests

rhrazdil:
latest operator needs rights to list/get nodes

**Release note**:

```release-note
NONE
```

### DIFF
--- a/test/e2e/workflow/nmstate_test.go
+++ b/test/e2e/workflow/nmstate_test.go
@@ -25,7 +25,7 @@ import (
 )
 
 var _ = Describe("NMState", func() {
-	nmstateVersion := "v0.67.0"
+	nmstateVersion := "v0.70.1"
 	gvk := GetCnaoV1GroupVersionKind()
 	Context("installed as standalone", func() {
 		BeforeEach(func() {


test/e2e/workflow/nmstate_test.go
```diff
@@ -25,7 +25,7 @@ import (
 )
 
 var _ = Describe("NMState", func() {
-	nmstateVersion := "v0.67.0"
+	nmstateVersion := "v0.70.1"
```
qinqon, line 28: Can we discover this number from github release or is not worth it ?

rhrazdil: I'm not sure if it would always help, since the container image we pull for operator is latest anyway
see https://github.com/kubevirt/cluster-network-addons-operator/pull/1339/files#diff-b523712def10219cc1b79738fa982d92baf4c16bcc6a7a2a26941de04b1e0b1aR70
and
https://github.com/kubevirt/cluster-network-addons-operator/pull/1339/files#diff-b523712def10219cc1b79738fa982d92baf4c16bcc6a7a2a26941de04b1e0b1aR123

rhrazdil: Well, we wouldn't need to update this, but only make sure we cut a release in knmstate, when making such changes

PR #1392, opened 2022-07-18
e2e ovs: Increase timeout

oshoval:
Default is 10m which isn't enough.
Use 20m as OVS repo e2e tests are using.

This fixes the problem we saw after bumping ovs, but it isn't due to the bump
this problem exists even on clean main.

**Release note**:
```release-note
None
```


oshoval:
@RamLavi @qinqon @phoracek 
If you can review please
it works with it, see please https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_cluster-network-addons-operator/1389/pull-e2e-cluster-network-addons-operator-ovs-cni-functests/1549006946550419456

PR #1408, opened 2022-08-23
cluster deploy: Fix installing openvswitch

oshoval:
Installing openvswitch failed because it couldn't
find a working mirror for centos-release-nfv-openvswitch.
Refreshing the dnf cache fixes this problem.

Note that installing centos-release-nfv-openvswitch
can be removed actually because it is already part of
kubevirtci, but this can be done on a follow commit.

**Release note**:
```release-note
None
```


oshoval:
Maybe error gone now

PR #1415, opened 2022-09-14
nmstate, upgrade: Check nncp before failing

qinqon:
/bug 

For OCP 4.11 CNAO does not support NMState anymore at its networkaddonsconfig, the expected behaviour is that if the user has used the old nmstate (by populating some NNCPs) and it has not installed KNO for CNAO to replace old kubernetes-nmstate it has to fail the upgrade. This change add the missing part about checking NNCPs.

Closes https://bugzilla.redhat.com/show_bug.cgi?id=2126537

**Release note**:

```release-note
Check NNCPs before failing on OCP 4.10 -> 4.11 upgrade.
```


rhrazdil (approved):
Thanks
/lgtm

PR #1433, opened 2022-10-07
multus: bump to multus-v4 (thick plugin)

maiqueb:
The multus bump script was updated to:
- use a different multus manifest - `multus-daemonset-thick.yml`
- ignore comments in the manifest file. Comments in the manifests broke the `yaml-utils::split_yaml_by_seperator` function.

**Release note**:

```release-note
Consume multus v4, which operates as a thick plugin.
```

### DIFF
--- a/automation/check-patch.e2e-multus-functests.sh
+++ b/automation/check-patch.e2e-multus-functests.sh
@@ -37,13 +37,20 @@ function __prepare-test-environment() {
 
     echo "Deploy cni-plugins pod"
     CNI_PLUGIN_YAML="cni-install.yml"
+    cp "templates/${CNI_PLUGIN_YAML}.j2" ${CNI_PLUGIN_YAML}
     $KUBECTL create -f ${CNI_PLUGIN_YAML}
     $KUBECTL -n kube-system wait --for=condition=ready -l name=cni-plugins pod --timeout=300s
 
-    TEST_YAML="*macvlan1.yml"
+    mkdir -p yamls
+    TEST_YAML="simple-macvlan1.yml"
     # matching the test nodes to the CNAO cluster nodes
-    sed -i 's/kind-worker$/node01/g' ${TEST_YAML}
-    sed -i 's/kind-worker2$/node02/g' ${TEST_YAML}
+    sed 's/{{ CNI_VERSION }}/0.4.0/g' "templates/${TEST_YAML}.j2" > "yamls/${TEST_YAML}"
+    sed -i 's/kind-worker$/node01/g' "yamls/${TEST_YAML}"
+    sed -i 's/kind-worker2$/node02/g' "yamls/${TEST_YAML}"
+
+    TEST_YAML="default-route1.yml"
+    # matching the test nodes to the CNAO cluster nodes
+    sed 's/{{ CNI_VERSION }}/0.4.0/g' "templates/${TEST_YAML}.j2" > "yamls/${TEST_YAML}"
 
     echo "Deplopy kubernetes-nmstate"
     $KUBECTL apply -f https://github.com/nmstate/kubernetes-nmstate/releases/download/v0.64.7/nmstate.io_nmstates.yaml
--- a/components.yaml
+++ b/components.yaml
@@ -25,10 +25,10 @@ components:
     metadata: v0.10.1
   multus:
     url: https://github.com/k8snetworkplumbingwg/multus-cni
-    commit: f4c0adf54c99d7395d30a050a8d29b674b99d700
+    commit: f530d3eb84608fa25a59737bcf49991d4a037d82
     branch: master
     update-policy: tagged
-    metadata: v3.9.1
+    metadata: v4.0.0-alpha
   ovs-cni:
     url: https://github.com/k8snetworkplumbingwg/ovs-cni
     commit: 05da205a682694307d1df78fdeccde1321f563f6
--- a/components.yaml
+++ b/components.yaml
@@ -25,7 +25,7 @@ components:
     metadata: v0.10.1
   multus:
     url: https://github.com/k8snetworkplumbingwg/multus-cni
-    commit: f530d3eb84608fa25a59737bcf49991d4a037d82
+    commit: fa8a6e688081ba5a097998f42b3f7dddb94dbed4
     branch: master
     update-policy: tagged
     metadata: v4.0.0-alpha
--- a/pkg/components/components.go
+++ b/pkg/components/components.go
@@ -29,7 +29,7 @@ var (
 )
 
 const (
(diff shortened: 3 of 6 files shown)


components.yaml
```diff
@@ -25,10 +25,10 @@ components:
     metadata: v0.9.0
   multus:
     url: https://github.com/k8snetworkplumbingwg/multus-cni
-    commit: f4c0adf54c99d7395d30a050a8d29b674b99d700
```
RamLavi, line 28: Since there a re DNM commits in the PR then I am assuming this PR is still WIP, but just to understand - do you plan to pin multus to a specific commit? if so - please set `update-policy` to "static", and `metadata` to the virtual tag (the latter is not necessary but nice to have).

if not - then perhaps it's best to keep the `update-policy` at "tagged" and follow the thick plugin branch (I'm guessing [this](https://github.com/k8snetworkplumbingwg/multus-cni/tree/feature/multus-4.0))?

pkg/components/components.go
```diff
@@ -29,7 +29,7 @@ var (
 )
 
 const (
-	MultusImageDefault            = "ghcr.io/k8snetworkplumbingwg/multus-cni@sha256:829c27e9392d013eee5086ca7670d7326d723ebaec526237215e86086b5a3234"
+	MultusImageDefault            = "ghcr.io/k8snetworkplumbingwg/multus-cni:v3.9.1-62-g1aac2431"
```
RamLavi: CNAO is now working with sha digests in U/S, so we need to keep using the sha. 
Usually the last part of the bump script is handling getting the sha and replacing it here - so I would try to check why that part didn't work..

RamLavi (commented):
just some preliminary comments, feel free to disregard if it's not helpful to you :)

components.yaml
maiqueb: that branch was a feature development branch. It is no longer being pushed to.

There isn't a tag with multus v4; that's the reason why I'm pointing to a commit. Thus, I cannot point metadata to a tag.

I'm not saying I want to point to a commit - right now, I literally do not have a choice since multus does not have a tag I can consume. 

... I still have to figure out what's wrong w/ the macvtap tests.

pkg/components/components.go
RamLavi: well, if we're specific I think we need commit's image to be pushed to `ghcr.io/k8snetworkplumbingwg/multus-cni`. usually push occurs only for tagged commits via some post script, but some repos do it for every commit (kmp for [example](https://quay.io/repository/kubevirt/kubemacpool?tab=tags&tag=latest)) so that's an option as well, albeit tagged is the preferred way to go..

maiqueb: Check the next commit: https://github.com/kubevirt/cluster-network-addons-operator/pull/1433/commits/c3ca44bc17488cf56f966eaf11b6ae7fa9037c9d#diff-f1531df30500812ec42a33a52023c9ceca8a93d4972132fd241ad143b107c063R32

The image exists, since, as you say, multus pushes all imgs to their container registry. What's broken is we [look for the tags attached to a commit](https://github.com/kubevirt/cluster-network-addons-operator/blob/main/hack/components/bump-multus.sh#L127); when none is found, somehow we create that weird thing.

RamLavi: ah yes you're right. when the tag is not there, it will use a virtual tag thing, cause this is how we tag it on [quay](https://quay.io/repository/kubevirt/kubemacpool?tab=tags&tag=latest) as well.. well if you need to change it to fit to how multus will now be used (w/wo tags) then feel free to do so :) .

components.yaml
maiqueb: Done. 

Let's see if this makes CI happy ... :crossed_fingers:

maiqueb:
Had to rebase latest master to address the macvtap issue.

maiqueb:
We're being held back for https://github.com/k8snetworkplumbingwg/multus-cni/issues/944 ...

automation/check-patch.e2e-multus-functests.sh
```diff
@@ -37,13 +37,20 @@ function __prepare-test-environment() {
...
     # matching the test nodes to the CNAO cluster nodes
-    sed -i 's/kind-worker$/node01/g' ${TEST_YAML}
-    sed -i 's/kind-worker2$/node02/g' ${TEST_YAML}
+    sed 's/{{ CNI_VERSION }}/0.4.0/g' "templates/${TEST_YAML}.j2" > "yamls/${TEST_YAML}"
+    sed -i 's/kind-worker$/node01/g' "yamls/${TEST_YAML}"
+    sed -i 's/kind-worker2$/node02/g' "yamls/${TEST_YAML}"
+
+    TEST_YAML="default-route1.yml"
+    # matching the test nodes to the CNAO cluster nodes
+    sed 's/{{ CNI_VERSION }}/0.4.0/g' "templates/${TEST_YAML}.j2" > "yamls/${TEST_YAML}"
```
RamLavi, line 53: this is duplicate with line 47, is that intentional?

RamLavi (changes requested):
Can you move the last commit `multus, temporal hack: use untagged multus version` to be the first?
then the "old first commit" aka `multus: bump to multus-v4 (thick plugin)` will only deal with the changes made to the bump script and it's outputs

automation/check-patch.e2e-multus-functests.sh
maiqueb: Yes; it is templating a different file, `default-route1.yml`, while the other is templating `simple-macvlan1.yml`.

RamLavi: ah sorry missed that

maiqueb: All good. If both required equal templating, I'd haven put this in a loop, which would have made this obvious.

... unfortunately, they don't.

maiqueb:
> > > Can you move the last commit `multus, temporal hack: use untagged multus version` to be the first? then the "old first commit" aka `multus: bump to multus-v4 (thick plugin)` will only deal with the changes made to the bump script and it's outputs
> > 
> > 
> > I can, but I don't see the advantage. What does that achieve ?
> 
> Since commits in this repo are squashed to 1 commit per PR, then it doesn't really matter, but if someone looks back at this PR then the first commit seems to use a wrong commit, that will surely break cause the bump fails to retrieve the tag --> SHA digest.. My intent is only to make it more readable, but if it is to much of a hustle, I won't insist

The first commit is the correct one - it uses the tag, that unfortunately has a bug. The top most commit is the patched workaround for us to move forward.

I'd prefer to leave this as is. 
@RamLavi

RamLavi:
Ok
/lgtm

/qinqon can you have a look as well?

RamLavi:
@qinqon can you have a look as well?

PR #1452, opened 2022-11-14
manifest: Restricted PSA on install

qinqon:
The CNAO workload cannot run on restricted PSA but it can be set on operator install to check audit-logs and then change it to privileged.

**Release note**:

```release-note
NONE
```


RamLavi (commented):
hmm.. @qinqon why do we use restricted on "namespace" manifest but change it on the fly to privileged on "operator-install"?
why not use privileged and be done with it?

PR #1466, opened 2022-11-24
makefile: Add missing component to bump-all target

oshoval:
This way it will be checked as part of the unit test lane that runs the bump-all target.

We need to support auto detection imo.

**Release note**:
```release-note
None
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -205,7 +205,7 @@ auto-bumper: $(GO)
 
 bump-%:
 	CNAO_VERSION=${VERSION} ./hack/components/bump-$*.sh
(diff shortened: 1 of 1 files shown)


RamLavi (approved):
/lgtm
/approve

maiqueb:
What do the unit tests have to do w/ make bump ?

oshoval:
> What do the unit tests have to do w/ make bump ?

It is in this lane
`automation/check-patch.unit-test.sh:    make bump-all`
After it, it runs `hack/check.sh` which checks `git status` is clean

PR #1474, opened 2022-12-07
components: Remove KubeSecondaryDNS

oshoval:
Until https://github.com/go-git/go-git/pull/485 is fixed since
it blocks the auto bumper when cloning the KSD repository.

**Release note**:
```release-note
None
```


components.yaml
```diff
@@ -41,9 +41,3 @@ components:
     branch: main
     update-policy: tagged
     metadata: v0.29.1
-  kube-secondary-dns:
```
RamLavi, line 44: I don't think removing this component entirely is a good idea, we still want to be able to manually bump, and for that we need this.

Can I suggest that we change the bumper so we don't exitWithError if it fails handling a specific component?

oshoval:
https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_cluster-network-addons-operator/1474/pull-e2e-cnao-kube-secondary-dns-functests/1600416437669728256

considering this fails, i will check your approach

oshoval:
Addressed comments, changed to skip the component instead of trying it

components.yaml
RamLavi: yeah that's what I meant

Regarding my suggestion, if we allow a component to fail auto-bump and continue to the other components, then at least we won't clock other components..

tools/bumper/bumper.go
```diff
@@ -50,6 +50,10 @@ func main() {
 
 	var failedComponents []string
 	for componentName, component := range componentsConfig.Components {
+		if componentName == "kube-secondary-dns" {
+			logger.Printf("~~Skipping %s because of known problem with go-git~~", componentName)
```
RamLavi, line 54: I would prefer that we fail and continue (so that the fail will be more visible and annoy us to fix it ASAP) but I don't insist on it for this PR.

Please add the link to the go-git issue

oshoval: Added https://github.com/go-git/go-git/issues/495

oshoval:
Thanks

oshoval:
Is there a presubmit lane that checks auto-bumper compiles ?
If not worth to add imo

PR #1476, opened 2022-12-07
components: Remove double export

oshoval:
`KUBEVIRT_PROVIDER` is already exported at the sourced `cluster.sh`, 
no need to override it.
It just confuse and harder to maintain because it will ignore the `cluster.sh` value
which is the main one we use.

**Release note**:
```release-note
None
```

### DIFF
--- a/automation/components-functests.setup.sh
+++ b/automation/components-functests.setup.sh
@@ -22,7 +22,6 @@ source hack/components/yaml-utils.sh
 source cluster/cluster.sh
 
 # Spin up Kubernetes cluster
-export KUBEVIRT_PROVIDER='k8s-1.23'
 make cluster-down cluster-up
 
 # Export .kubeconfig full path, so it will be possible


oshoval:
/cc @qinqon 
can you take a look please ?

ormergi:
Please note that on the other hand, one would need to dig in other scripts if he likes to understand which provider/k8s version is being used.
IIRC it's there for visibility.

oshoval:
> Please note that on the other hand, one would need to dig in other scripts if he likes to understand which provider/k8s version is being used. IIRC it's there for visibility.

We need only one place that we change and affect what version we run,
and we need that the CNAO version on its own lane is the same one that is running on the components that CNAO is loading.
It isn't right to need to change multi places.
It is sourced, and easily seen in the log as we have +x all over.

We will soon also have some automation or semi automation that will bump the k8s version, and it will touch only the cluster.sh file which is common on all network repos.

PR #1485, opened 2022-12-18
Fix undeploy components

oshoval:
The current config was used instead of the previous.
In the current config it is empty, we need to pass the previous config, 
so in case the component was configured in the previous config it will render its 
object and add it to the delete list.

**Release note**:
```release-note
NONE
```


RamLavi (approved):
/lgtm
/approve
good catch!

oshoval:
Thanks
can you please create a new release ? 
so it will be consumed by HCO and fix this problem

PR #1494, opened 2022-12-27
Add `operator_health_impact` label to CNAO alerts

assafad:
Signed-off-by: assafad <aadmi@redhat.com>

This PR adds to each of CNAO alerts a label named `operator_health_impact`, which indicates how the alerts impact the operator health. This will enable us filtering alerts based on their impact on the operator health, and to later have an operator health metric, which will tell the user what is the aggregated operator health.

This label gets one of the following values:
- `critical` indicates that the alert fires due to an issue that impacts the operator's functionality badly, and an action must be taken immediately in order to solve it.
-  `warning` indicates that the alert fires due to an issue that soon might impact the operator's functionality badly, and the user should be warned in order to solve the issue.
- `none` indicates that the alert fires due to an issue that doesn't affect the operator health, and in most cases is related to the workload rather than the operator itself.

**Which issue(s) this PR fixes**:
https://issues.redhat.com/browse/CNV-18923

**Release note**:

```release-note
None
```

### DIFF
--- a/data/monitoring/prom-rule.yaml
+++ b/data/monitoring/prom-rule.yaml
@@ -20,6 +20,7 @@ spec:
           for: 5m
           labels:
             severity: warning
+            operator_health_impact: warning
             kubernetes_operator_part_of: kubevirt
             kubernetes_operator_component: cluster-network-addons-operator
         - alert: NetworkAddonsConfigNotReady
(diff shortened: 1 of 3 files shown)


RamLavi (commented):
@assafad can you explain why some some are labeled critical when the severity is warning?

RamLavi (approved):
/lgtm

@sradco please review so I could merge the PR

PR #1499, opened 2023-01-10
[ksd] Parameterize "runAsNonRoot" in the yaml

AlonaKaplan:
In order to make ksd pod to work with restricted psa policy it should have `runAsNonRoot: true`.
However, once the pod's securityContext has `runAsNonRoot: true` it should also have a specific `rusAsUser`.
Having a specific user is problematic on Openshift since it may configure a range of valid ids (example of a possible error - `spec.containers[0].securityContext.runAsUser: Invalid value: 1000: must be in the ranges: [1000670000, 1000679999]`).

According to "What happens if I use runAsUser in my workloads?" section in https://github.com/redhat-openshift-ecosystem/community-operators-prod/discussions/1417 the suggested solution is - 
```
If you are unable to change the image, for Openshift distributions you MUST leave the RunAsUser and RunAsNonRoot fields empty and let the Openshift SCC inject the values when your workload is qualified to the SCC policy.
```

Since one of the kds pod containers is CoreDNS and not controlled by the ksd repository, this is the preferred solution in the described scenario.

**Release note**:

```release-note
[KSD] set `runAsNonRoot: null` and `runAsUser:null` in case of Openshift environment.
```

### DIFF
--- a/pkg/network/kube_secondary_dns_controller.go
+++ b/pkg/network/kube_secondary_dns_controller.go
@@ -26,6 +26,13 @@ func renderKubeSecondaryDNS(conf *cnao.NetworkAddonsConfigSpec, manifestDir stri
 	data.Data["NameServerIp"] = conf.KubeSecondaryDNS.NameServerIP
 	data.Data["KubeSecondaryDNSImage"] = os.Getenv("KUBE_SECONDARY_DNS_IMAGE")
 	data.Data["CoreDNSImage"] = os.Getenv("CORE_DNS_IMAGE")
+	if clusterInfo.SCCAvailable {
+		data.Data["RunAsNonRoot"] = "null"
+		data.Data["RunAsUser"] = "null"
+	} else {
+		data.Data["RunAsNonRoot"] = "true"
+		data.Data["RunAsUser"] = "1000"
+	}
 
 	objs, err := render.RenderDir(filepath.Join(manifestDir, "kube-secondary-dns"), &data)
 	if err != nil {
--- a/data/kube-secondary-dns/secondarydns.yaml
+++ b/data/kube-secondary-dns/secondarydns.yaml
@@ -76,8 +76,8 @@ spec:
     spec:
       serviceAccountName: secondary
       securityContext:
-        runAsUser: 1000
-        runAsNonRoot: true
+        runAsUser: {{ .RunAsUser }}
+        runAsNonRoot: {{ .RunAsNonRoot }}
         seccompProfile:
           type: "RuntimeDefault"
       containers:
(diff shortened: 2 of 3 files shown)


pkg/network/kube_secondary_dns_controller.go
```diff
@@ -26,6 +26,13 @@ func renderKubeSecondaryDNS(conf *cnao.NetworkAddonsConfigSpec, manifestDir stri
 	data.Data["NameServerIp"] = conf.KubeSecondaryDNS.NameServerIP
 	data.Data["KubeSecondaryDNSImage"] = os.Getenv("KUBE_SECONDARY_DNS_IMAGE")
 	data.Data["CoreDNSImage"] = os.Getenv("CORE_DNS_IMAGE")
+	if clusterInfo.SCCAvailable {
+		data.Data["RunAsNonRoot"] = "null"
```
phoracek, line 30: Passing these null strings looks odd to me and is out of sync with what we do with the other yamls.

Could you please just export the boolean here, maybe

```
data.Data["RunAsNonRoot"] = clusterInfo.SCCAvailable
```

And then us `if` in the template? https://github.com/kubevirt/cluster-network-addons-operator/blob/main/data/linux-bridge/001-rbac.yaml#L1

phoracek (changes requested):

pkg/network/kube_secondary_dns_controller.go
AlonaKaplan: I find this option a bit problematic. 
Something like -
```
{{ if .runAsNonRoot }}
        runAsNonRoot: true
        runAsUser: 1000
{{ end }}
```
has to be injected in the middle of the yaml.
We will need to have a script finding `runAsNonRoot` and `runAsUser` and surrounding it with the condition.
What if for example ksd will add some other param between runAsNonRoot and runAsUser? It will break cnao.

I think setting the `null` values should be safer a cleaner.
https://www.educative.io/answers/how-to-represent-null-values-in-yaml

phoracek: Makes sense. 

Not sure about the way you use null in this PR though. It is quoted and thus a string, not a real null. So it would be evaluated as an invalid value, instead of absent attribute. No?

AlonaKaplan: Thanks for the comment. I tested it, it is working.
I deployed `kubeSecondaryDNS` with cnao. The deployment and pods were created and running.
Their yaml doesn't contain both `runAsNonRoot` and `runAsUser`.

phoracek: I see, ok

phoracek (approved):
/lgtm
/approve

PR #1510, opened 2023-02-07
Fix nmstate service cleanup

phoracek:

The original code was using a wrong API group. With this patch, the group is fixed and a test is added.

**Release note**:

```release-note
NONE
```


phoracek:
/hold

Checking if the test fails with the original code.

phoracek:
/hold cancel

The test has reproduced the issue

PR #1513, opened 2023-03-02
Multus: Enable deploying Multus V3.X (thin CNI plugin form)

ormergi:
Currently, the Multus components deploy Multus V4 (thick CNI plugin form) which is not GA'ed yet.

This PR adds an additional Multus component in order to enable deploying Multus V3.X (thin CNI plugin form) which is GA version.

-  Having CNAO able to deploy both version enable testing Kubevirt integration with both versions and help the development of the hotplug NICs feature.
- When the `NetworkAddonsConfig` spec is changed to use one component instead of the other  (i.e: replace `multus: {}` with `multusV3: {}` or the other way around), the correct version will be deployed.
- Having both `Multus` and `MultusV3` components specified in the `NetworkAddonsConfig` spec is not supported.

**Release note**:

```release-note

```


components.yaml
```diff
@@ -35,6 +35,12 @@ components:
     branch: master
     update-policy: static
     metadata: ""
+  multusV3:
+    url: https://github.com/k8snetworkplumbingwg/multus-cni
+    commit: f4c0adf54c99d7395d30a050a8d29b674b99d700
+    branch: master
+    update-policy: tagged
+    metadata: v3.9.1
```
maiqueb, line 43: Why 3.9.1 instead of [3.9.3](https://github.com/k8snetworkplumbingwg/multus-cni/releases/tag/v3.9.3) (latest) ?

pkg/network/multus.go
```diff
@@ -107,3 +107,32 @@ func renderMultus(conf *cnao.NetworkAddonsConfigSpec, manifestDir string, opensh
...
+	// render manifests from disk
+	data := render.MakeRenderData()
+	data.Data["Namespace"] = os.Getenv("OPERAND_NAMESPACE")
+	data.Data["MultusImage"] = os.Getenv("MULTUS_V3_IMAGE")
+	data.Data["ImagePullPolicy"] = conf.ImagePullPolicy
+	data.Data["Placement"] = conf.PlacementConfiguration.Workloads
+	if clusterInfo.OpenShift4 {
+		data.Data["CNIConfigDir"] = cni.ConfigDirOpenShift4
+		data.Data["CNIBinDir"] = cni.BinDirOpenShift4
+	} else {
```
maiqueb, line 126: I thought we wanted to prevent multus from being deployed in openshift.

Am I missing something ?

maiqueb (changes requested):
shouldn't we make the multus components mutually exclusive ? You either have one or the other. 

And add unit tests for that please.

README.md
```diff
@@ -37,10 +37,23 @@ metadata:
...
+```yaml
+apiVersion: networkaddonsoperator.network.kubevirt.io/v1
+kind: NetworkAddonsConfig
+metadata:
+  name: cluster
+spec:
+  multusV3: {}
+```
+
+> **_NOTE:_** Having both `multus` and `multusV3` specified is not supported and should be avoided.
```
oshoval, line 55: can it be enforced?

components.yaml
```diff
@@ -35,6 +35,12 @@ components:
     branch: master
     update-policy: static
     metadata: ""
+  multusV3:
```
oshoval, line 38: missing bump script imo
it will fail the auto bumper isnt it ?

pkg/network/multus.go
```diff
@@ -107,3 +107,32 @@ func renderMultus(conf *cnao.NetworkAddonsConfigSpec, manifestDir string, opensh
...
+		data.Data["CNIConfigDir"] = cni.ConfigDir
+		data.Data["CNIBinDir"] = cni.BinDir
+	}
+	data.Data["EnableSCC"] = clusterInfo.SCCAvailable
```
oshoval, line 130: do we need it ?

oshoval (commented):
Thanks
not a deep view

EdDev:
> shouldn't we make the multus components mutually exclusive ? You either have one or the other.

I would prefer to have no such logic in CNAO. We can document it, but not go beyond that.

I would suggest to update the release-notes section in the PR and add this warning.

maiqueb:
> > shouldn't we make the multus components mutually exclusive ? You either have one or the other.
> 
> I would prefer to have no such logic in CNAO. We can document it, but not go beyond that.

Why ? 

> 
> I would suggest to update the release-notes section in the PR and add this warning.

EdDev:
> > > shouldn't we make the multus components mutually exclusive ? You either have one or the other.
> > 
> > 
> > I would prefer to have no such logic in CNAO. We can document it, but not go beyond that.
> 
> Why ?

Because if someone wants to intentionally configure something wrongly, we do not have to protect against it.
It adds a level of complexity to this and we wanted to keep it as simple as possible.

maiqueb:
> > > > shouldn't we make the multus components mutually exclusive ? You either have one or the other.
> > > 
> > > 
> > > I would prefer to have no such logic in CNAO. We can document it, but not go beyond that.
> > 
> > 
> > Why ?
> 
> Because if someone wants to intentionally configure something wrongly, we do not have to protect against it. It adds a level of complexity to this and we wanted to keep it as simple as possible.

My opinion is we should help the user by at least making it hard to get things misconfigured, and throw meaningful error messages back.

But I won't insist.

maiqueb:
> > > > > shouldn't we make the multus components mutually exclusive ? You either have one or the other.
> > > > 
> > > > 
> > > > I would prefer to have no such logic in CNAO. We can document it, but not go beyond that.
> > > 
> > > 
> > > Why ?
> > 
> > 
> > Because if someone wants to intentionally configure something wrongly, we do not have to protect against it. It adds a level of complexity to this and we wanted to keep it as simple as possible.
> 
> My opinion is we should help the user by at least making it hard to get things misconfigured, and throw meaningful error messages back.
> 
> But I won't insist.

EDIT: I have no idea what will happen if both are enabled, and given our "amazing" upgrade track record, I am not sure how to recover it once this happens - other than manually deleting binaries from the node's file system.

EdDev:
I understand the cons here, but I still would prefer not to manage it. It creates a dependency between these two and I think it is a bad move.

I do not know if CNAO supports the removal of deployments, but if it does, one could remove the other before getting deployed.

maiqueb:
> I understand the cons here, but I still would prefer not to manage it. It creates a dependency between these two and I think it is a bad move.
> 
> I do not know if CNAO supports the removal of deployments, but if it does, one could remove the other before getting deployed.

AFAIU, it doesn't. That's why I think it is important for us to prevent users from making preventable mistakes.

EdDev:
> AFAIU, it doesn't. That's why I think it is important for us to make it hard for users to make preventable mistakes.

Ok, so if that is the case, I would consider not using CNAO to deploy Multus and do it ourself in kubevirtci.
I prefer a very simple solution or no solution at all.

As I see it, CNAO is broken because of what it deploys, but this is obviously not under any consensus. This was an idea to overcome the problem, but if it complicates CNAO core, it does not worth it.

Lets see what the maintainers say about this.

oshoval:
CNAO does support removal

components.yaml
ormergi: I just reverted to what CNAO used to deploy before the transition to Multus v4 (taken from v0.80.0)

pkg/network/multus.go
ormergi: I followed the example of Multus component we have today.

components.yaml
ormergi: > missing bump script imo (even blank if should do nothing)
> it will fail the auto bumper isnt it ?

What do you mean by that?

> also which one should be the default for now by CNAO / HCO , not the v3 ?

The default remains  Multus v4.

pkg/network/multus.go
ormergi: I followed the example of Multus component we have today.

https://github.com/kubevirt/cluster-network-addons-operator/pull/1513#discussion_r1123363404

components.yaml
oshoval: > > missing bump script imo (even blank if should do nothing)
> > it will fail the auto bumper isnt it ?
> 
> What do you mean by that?

Each component has a script
`./hack/components/bump-<COMPONENET_NAME>.sh`
the auto bumper is a script that consumes the yaml and runs the scripts,
if a script is missing it will fail

I would suggest to look on other component adding and check if more stuff is missing

README.md
ormergi: I think it's possible, the idea of the current implementation is to keep things simple as possible and avoid the maintenance overhead.

This is an intermediate state until Multus V4 is GA'ed on U/S.

ormergi:
CNAO does support the removal of components, when a component is removed for the CR spec it will be reconciled and perform removal of that component.
In the scenerio where one switch between Multus components as follows:
```
---
apiVersion: networkaddonsoperator.network.kubevirt.io/v1
kind: NetworkAddonsConfig
metadata:
  name: cluster
spec:
  multus: {}

         |     /\
        \/     |

---
apiVersion: networkaddonsoperator.network.kubevirt.io/v1
kind: NetworkAddonsConfig
metadata:
  name: cluster
spec:
  multusV3: {}
```

The old Multus version will be removed and the new one deployed.

RamLavi (changes requested):
Hey @ormergi thank you for this change.
I've made some inline comments.
Also:
1. If we are "returning the wheel" back then the CI (test/lifecycle, test/workflow, multus-functest)
needs to split as well to: 
- lane that runs with multusv3
- lane that runs with cutting edge, non GA, still developing multusv4 one, as a non blocking lane. suggestion: could you please change `components-functests.setup.sh` so that it supports a flag to deploy multusv4 instead of v3 (which will be the default)

2. I think we should not allow deploying the 2 multueses together. it is CNAO job to delpoy, and it should also prevent false deployment. I agree with @maiqueb it would be easier to prevent than to fix.

Makefile
```diff
@@ -155,6 +155,7 @@ gen-manifests: manifest-templator
 	CONTAINER_PREFIX=$(IMAGE_REGISTRY) \
 	CONTAINER_TAG=$(IMAGE_TAG) \
 	MULTUS_IMAGE=$(MULTUS_IMAGE) \
+	MULTUS_V3_IMAGE=$(MULTUS_V3_IMAGE) \
```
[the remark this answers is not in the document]
ormergi: Maybe I specified the obvious, but since its another flavor of Multus component one could think it works differently, so I wanted to emphasize it should work the same as any other component.
I can remove it if you think it's redundant.

cluster/operator-push.sh
```diff
@@ -16,13 +16,20 @@
...
+manifest_registry=""
 if [[ "${KUBEVIRT_PROVIDER}" == external ]]; then
     if [[ ! -v DEV_IMAGE_REGISTRY ]]; then
         echo "Missing DEV_IMAGE_REGISTRY variable"
         return 1
     fi
     manifest_registry=$DEV_IMAGE_REGISTRY
-    push_registry=$manifest_registry
+
+    if [[ -z DEV_PUSH_REGISTRY ]]; then
```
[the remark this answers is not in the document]
ormergi: It slipped in mistaknly, I will remove it from this PR.

Makefile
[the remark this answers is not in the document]
RamLavi: I would prefer, it left me more perplexed than anything :) 
The better place to put this info is on the README, and you already did that.

pkg/network/multus.go
oshoval: we wont reach here due to 
https://github.com/kubevirt/cluster-network-addons-operator/pull/1513/files#diff-f905e8c1fce38a3ea9acd2b9fe20ab7b1217840a4e47c65595a9682491dbaf5bR113

right ?
so worth imo to remove for the new function in this PR

maiqueb:
> CNAO does support the removal of components, when a component is removed for the CR spec it will be reconciled and perform removal of that component. In the scenerio where one switch between Multus components as follows:
> 
> ```
> ---
> apiVersion: networkaddonsoperator.network.kubevirt.io/v1
> kind: NetworkAddonsConfig
> metadata:
>   name: cluster
> spec:
>   multus: {}
> ...
> 
>         \/  /\  
> 
> ---
> apiVersion: networkaddonsoperator.network.kubevirt.io/v1
> kind: NetworkAddonsConfig
> metadata:
>   name: cluster
> spec:
>   multusV3: {}
> ...
> ```
> 
> The old Multus version will be removed and the new one deployed.

What do you mean with "removed" ? I don't think the binaries nor CNI configurations are deleted from the file-system, which means they'll still be available . 

It **may** be OK if both multus versions generate configurations using the same name: i.e.  you'll overwrite the `00-multus.conf` configurations on the CNI conf folder.

Please confirm this, try one upgrade path locally from v4 to v3 and back to v4, and assure you can still create pods.

oshoval:
> > CNAO does support the removal of components, when a component is removed for the CR spec it will be reconciled and perform removal of that component. In the scenerio where one switch between Multus components as follows:
> > ```
> > ---
> > apiVersion: networkaddonsoperator.network.kubevirt.io/v1
> > kind: NetworkAddonsConfig
> > metadata:
> >   name: cluster
> > spec:
> >   multus: {}
> > ...
> > 
> >         \/  /\  
> > 
> > ---
> > apiVersion: networkaddonsoperator.network.kubevirt.io/v1
> > kind: NetworkAddonsConfig
> > metadata:
> >   name: cluster
> > spec:
> >   multusV3: {}
> > ...
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The old Multus version will be removed and the new one deployed.
> 
> What do you mean with "removed" ? I don't think the binaries nor CNI configurations are deleted from the file-system, which means they'll still be available .
> 
> It **may** be OK if both multus versions generate configurations using the same name: i.e. you'll overwrite the `00-multus.conf` configurations on the CNI conf folder.
> 
> Please confirm this, try one upgrade path locally from v4 to v3 and back to v4, and assure you can still create pods.

only k8s resources are deleted by CNAO, unless the pods have a del trigger which is not the case
side effects are kept, it is indeed the question if reconfig will override everything without leftovers that can do mess
for example lower id configs etc

RamLavi:
@ormergi as we plan to not introduce multusv3 as an individual component at the moment, can you please close this PR?

ormergi:
/close 

As we agree to not proceed with this change at the moment.
Thanks for the time and review :slightly_smiling_face:

PR #1515, opened 2023-03-08
placement-configuration: Add infra tolerations

RamLavi:
Currently the default toleration for pods is to evict a node after 5 minutes that the node is not-ready/unschedulable.
In pods running on master nodes, this eviction needs to be faster in order to not delay a node's reboot.

This PR is adding infra tolerations that allows a pod to evict after 1 minute in case the node it is scheduled on is not-ready/unschedulable. 
As currently the only component using infra placement is Kubemacpool, this is the only component that will be affected.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/network/placement_configuration.go
+++ b/pkg/network/placement_configuration.go
@@ -7,6 +7,7 @@ import (
 )
 
 func GetDefaultPlacementConfiguration() cnao.PlacementConfiguration {
+	var nodeNotNReadyTolerationInSeconds int64 = 60
 	return cnao.PlacementConfiguration{
 		Infra: &cnao.Placement{
 			Tolerations: []corev1.Toleration{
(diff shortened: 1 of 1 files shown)


rhrazdil (commented):
I would perhaps improve the wording of the commit message, as the way I understand it, this toleration will allow KMP to keep running on unreachable/not-ready node for up to a minute, after which it is finally evicted. Isn't it so?

RamLavi:
> I would perhaps improve the wording of the commit message, as the way I understand it, this toleration will allow KMP to keep running on unreachable/not-ready node for up to a minute, after which it is finally evicted. Isn't it so?

yeah, the default toleration of these annotations is 5 mins. We want KMP to evict faster in case of master node reboot.

rhrazdil:
>> I would perhaps improve the wording of the commit message, as the way I understand it, this toleration will allow KMP to keep running on unreachable/not-ready node for up to a minute, after which it is finally evicted. Isn't it so?

> yeah, the default toleration of these annotations is 5 mins. We want KMP to evict faster in case of master node reboot.
Added more info to the commit and PR desc. wdyt?

Oh, I didn't know about the 5 minute default. Good to know.
It makes perfect sense now.

rhrazdil (approved):

Issue #1512, opened 2023-02-27
macvtap-deviceplugin-config keeps getting overwritten


davidpanic:
**What happened**:

When trying to apply the macvtap config, something keeps overwriting it. It is always being set to `DP_MACVTAP_CONF: '[]'`.

**What you expected to happen**:

I expect to be able to set up the config.

**How to reproduce it (as minimally and precisely as possible)**:

Apply the following config:

```yaml
kind: ConfigMap
apiVersion: v1
metadata:
  name: macvtap-deviceplugin-config
  namespace: kubevirt-hyperconverged
data:
  DP_MACVTAP_CONF: |
    [ {
        "name" : "localnet",
        "lowerDevice" : "bond0",
        "mode": "bridge",
        "capacity" : 50
    } ]
```

`$ kubectl apply -f macvtap.yaml`
```
configmap/macvtap-deviceplugin-config configured
```

...Wait about 1-2 minutes

`$ kubectl -n kubevirt-hyperconverged get configmap macvtap-deviceplugin-config -o yaml`
```yaml
apiVersion: v1
data:
  DP_MACVTAP_CONF: '[]'
kind: ConfigMap
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"v1","data":{"DP_MACVTAP_CONF":"[ {\n    \"name\" : \"localnet\",\n    \"lowerDevice\" : \"bond0\",\n    \"mode\": \"bridge\",\n    \"capacity\" : 50\n} ]\n"},"kind":"ConfigMap","metadata":{"annotations":{},"name":"macvtap-deviceplugin-config","namespace":"kubevirt-hyperconverged"}}
  creationTimestamp: "2023-02-27T12:39:39Z"
  labels:
    app.kubernetes.io/component: network
    app.kubernetes.io/managed-by: cnao-operator
    app.kubernetes.io/part-of: hyperconverged-cluster
    app.kubernetes.io/version: 1.8.0
    networkaddonsoperator.network.kubevirt.io/version: v0.79.1
    prometheus.cnao.io: "true"
  name: macvtap-deviceplugin-config
  namespace: kubevirt-hyperconverged
  ownerReferences:
  - apiVersion: networkaddonsoperator.network.kubevirt.io/v1
    blockOwnerDeletion: true
    controller: true
    kind: NetworkAddonsConfig
    name: cluster
[...]

PR #1519, opened 2023-03-08
macvtap: allows overriding the macvtap device plugin configuration

maiqueb:
This PR allows users to define the macvtap-cni device-plugin configuration instead of using link auto-discovery (i.e. all links available in the host would be allowed as macvtap lower devices).

For that, we need macvtap-cni [v.0.11.0](https://github.com/kubevirt/macvtap-cni/releases/tag/v0.11.0), which templates the name of the config map holding the device plugin configuration.

Fixes: #1512 

**Release note**:

```release-note
Allows overriding the macvtap-cni device plugin configuration.
```

### DIFF
--- a/pkg/apis/networkaddonsoperator/shared/networkaddonsconfig_types.go
+++ b/pkg/apis/networkaddonsoperator/shared/networkaddonsconfig_types.go
@@ -81,7 +81,11 @@ type KubeMacPool struct {
 }
 
 // MacvtapCni plugin allows users to define Kubernetes networks on top of existing host interfaces
-type MacvtapCni struct{}
+type MacvtapCni struct {
+	// DevicePluginConfig allows the user to override the name of the
+	// `ConfigMap` where the device plugin configuration is held.
+	DevicePluginConfig string `json:"devicePluginConfig,omitempty"`
+}
 
 // NetworkAddonsConfigStatus defines the observed state of NetworkAddonsConfig
 type NetworkAddonsConfigStatus struct {
--- a/pkg/components/components.go
+++ b/pkg/components/components.go
@@ -35,7 +35,7 @@ const (
 	LinuxBridgeMarkerImageDefault     = "quay.io/kubevirt/bridge-marker@sha256:5d24c6d1ecb0556896b7b81c7e5260b54173858425777b7a84df8a706c07e6d2"
 	KubeMacPoolImageDefault           = "quay.io/kubevirt/kubemacpool@sha256:0cc5ad824fc163d6dea5e9bd872467c691eaa9a88944008b5d746495b2a72214"
 	OvsCniImageDefault                = "quay.io/kubevirt/ovs-cni-plugin@sha256:5f7290e2294255ab2547c3b4bf48cc2d75531ec5a43e600366e9b2719bef983f"
-	MacvtapCniImageDefault            = "quay.io/kubevirt/macvtap-cni@sha256:5a288f1f9956c2ea8127fa736b598326852d2aa58a8469fa663a1150c2313b02"
+	MacvtapCniImageDefault            = "quay.io/kubevirt/macvtap-cni@sha256:434420511e09b2b5ede785a2c9062b6658ffbc26fbdd4629ce06110f9039c600"
 	KubeRbacProxyImageDefault         = "quay.io/openshift/origin-kube-rbac-proxy@sha256:baedb268ac66456018fb30af395bb3d69af5fff3252ff5d549f0231b1ebb6901"
 	KubeSecondaryDNSImageDefault      = "ghcr.io/kubevirt/kubesecondarydns@sha256:b489a7c5d05b000f776c9c302985b8b7f29ff31f577a1480912ed625c8772d6b"
 	CoreDNSImageDefault               = "k8s.gcr.io/coredns/coredns@sha256:5b6ec0d6de9baaf3e92d0f66cd96a25b9edbce8716f5f15dcd1a616b3abd590e"
--- a/pkg/components/components.go
+++ b/pkg/components/components.go
@@ -824,6 +824,12 @@ func GetCrd() *extv1.CustomResourceDefinition {
 						"macvtap": extv1.JSONSchemaProps{
 							Description: "MacvtapCni plugin allows users to define Kubernetes networks on top of existing host interfaces",
 							Type:        "object",
+							Properties: map[string]extv1.JSONSchemaProps{
+								"devicePluginConfig": extv1.JSONSchemaProps{
+									Description: "DevicePluginConfig allows the user to override the name of the `ConfigMap` where the device plugin configuration is held",
+									Type:        "string",
+								},
+							},
 						},
 						"multus": extv1.JSONSchemaProps{
 							Description: "Multus plugin enables attaching multiple network interfaces to Pods in Kubernetes",
--- a/pkg/network/macvtap.go
+++ b/pkg/network/macvtap.go
@@ -24,6 +24,7 @@ func renderMacvtapCni(conf *cnao.NetworkAddonsConfigSpec, manifestDir string, cl
 	data.Data["ImagePullPolicy"] = conf.ImagePullPolicy
 	data.Data["EnableSCC"] = clusterInfo.SCCAvailable
 	data.Data["MacvtapImage"] = os.Getenv("MACVTAP_CNI_IMAGE")
+	data.Data["DevicePluginConfigName"] = conf.MacvtapCni.DevicePluginConfig
 	if clusterInfo.OpenShift4 {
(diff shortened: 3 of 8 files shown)


pkg/network/macvtap.go
```diff
@@ -37,3 +37,12 @@ func renderMacvtapCni(conf *cnao.NetworkAddonsConfigSpec, manifestDir string, cl
...
 	return objs, nil
 }
+
+func fillMacvtapDefaults(conf *cnao.NetworkAddonsConfigSpec) []error {
+	// https://github.com/kubevirt/macvtap-cni/blob/be1528fb09e9ac3c490a5df31330851d7e1f8b0a/manifests/macvtap.yaml#L23
+	const defaultMacvtapDevicePluginConfigMapName = "macvtap-deviceplugin-config"
+	if conf.MacvtapCni != nil && conf.MacvtapCni.DevicePluginConfig == "" {
+		conf.MacvtapCni.DevicePluginConfig = defaultMacvtapDevicePluginConfigMapName
+	}
+	return nil
```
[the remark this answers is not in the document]
maiqueb: No point in creating an empty slice when a nil does the same thing. 

The only difference would be when marshalling: the nil would marshall into `null` while the empty slice would marchall to `[]` . 

I think using nil is more idiomatic since that's the value it sets when you `var slice []<type>` . 

I can use your suggestion if you prefer.

pkg/network/macvtap.go
```diff
@@ -37,3 +37,12 @@ func renderMacvtapCni(conf *cnao.NetworkAddonsConfigSpec, manifestDir string, cl
 
 	return objs, nil
 }
+
+func fillMacvtapDefaults(conf *cnao.NetworkAddonsConfigSpec) []error {
```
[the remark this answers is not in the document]
maiqueb: I think I understand. But don't we want the opposite ? i.e. if you don't specify you want to **disable** something, it should be enabled ? 

I mean this was intentional. Does my argument make sense ? This is, all in all a boolean, which has a default. 

I guess this is one of the reasons the `optional` json tag exists ... What should we do then ?

pkg/network/network.go
```diff
@@ -61,6 +61,7 @@ func FillDefaults(conf, previous *cnao.NetworkAddonsConfigSpec) error {
 	errs = append(errs, fillDefaultsSelfSignConfiguration(conf, previous)...)
 	errs = append(errs, fillDefaultsImagePullPolicy(conf, previous)...)
 	errs = append(errs, fillDefaultsKubeMacPool(conf, previous)...)
+	fillMacvtapDefaults(conf)
```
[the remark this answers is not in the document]
maiqueb: Yes, I think not returning errors is preferable, since there's no way this can "fail".

pkg/network/macvtap.go
```diff
@@ -37,3 +38,12 @@ func renderMacvtapCni(conf *cnao.NetworkAddonsConfigSpec, manifestDir string, cl
...
+	// https://github.com/kubevirt/macvtap-cni/blob/be1528fb09e9ac3c490a5df31330851d7e1f8b0a/manifests/macvtap.yaml#L23
+	const defaultMacvtapDevicePluginConfigMapName = "macvtap-deviceplugin-config"
+	if conf.MacvtapCni != nil && conf.MacvtapCni.DevicePluginConfig == "" {
+		conf.MacvtapCni.DevicePluginConfig = defaultMacvtapDevicePluginConfigMapName
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
maiqueb: OK.

pkg/network/macvtap.go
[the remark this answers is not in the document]
RamLavi: that's ok no need to change..

pkg/network/macvtap.go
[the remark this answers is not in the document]
maiqueb: let's disregard this; I'll follow what's being done for other components.

pkg/network/network.go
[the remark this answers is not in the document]
maiqueb: Done.

maiqueb:
... now I should just understand how can I debug the e2e test I've added.

maiqueb:
/test pull-e2e-cnao-multus-dynamic-networks-functests

wow a flake here .. I'll keep an eye out.

maiqueb:
@RamLavi could you take a look ?

pkg/apis/networkaddonsoperator/shared/networkaddonsconfig_types.go
```diff
@@ -81,7 +81,11 @@ type KubeMacPool struct {
 }
 
 // MacvtapCni plugin allows users to define Kubernetes networks on top of existing host interfaces
-type MacvtapCni struct{}
+type MacvtapCni struct {
+	// DevicePluginConfig allows the user to override the name of the
+	// `ConfigMap` where the device plugin configuration is held.
```
RamLavi, line 86: do we need to add this `// +optional` (like in [here](https://github.com/kubevirt/kubevirt/blob/main/staging/src/kubevirt.io/api/core/v1/componentconfig.go#L19))?

pkg/components/components.go
```diff
@@ -35,7 +35,7 @@ const (
 	LinuxBridgeMarkerImageDefault     = "quay.io/kubevirt/bridge-marker@sha256:5d24c6d1ecb0556896b7b81c7e5260b54173858425777b7a84df8a706c07e6d2"
 	KubeMacPoolImageDefault           = "quay.io/kubevirt/kubemacpool@sha256:0cc5ad824fc163d6dea5e9bd872467c691eaa9a88944008b5d746495b2a72214"
 	OvsCniImageDefault                = "quay.io/kubevirt/ovs-cni-plugin@sha256:5f7290e2294255ab2547c3b4bf48cc2d75531ec5a43e600366e9b2719bef983f"
-	MacvtapCniImageDefault            = "quay.io/kubevirt/macvtap-cni@sha256:5a288f1f9956c2ea8127fa736b598326852d2aa58a8469fa663a1150c2313b02"
+	MacvtapCniImageDefault            = "quay.io/kubevirt/macvtap-cni@sha256:434420511e09b2b5ede785a2c9062b6658ffbc26fbdd4629ce06110f9039c600"
```
RamLavi, line 38: I think this belongs to the bump commit, can you move it?

pkg/network/macvtap_test.go
```diff
@@ -0,0 +1,42 @@
...
+				MacvtapCni: &cnao.MacvtapCni{DevicePluginConfig: configMapNameFromOldDays},
+			}
+			fillMacvtapDefaults(&config, &oldConfig)
+			Expect(config).To(Equal(
+				cnao.NetworkAddonsConfigSpec{
+					MacvtapCni: &cnao.MacvtapCni{
+						DevicePluginConfig: configMapNameFromOldDays},
+				}))
+		})
+	})
```
RamLavi, line 41: please add another scenario, where:
```
oldConfig := cnao.NetworkAddonsConfigSpec{
	MacvtapCni: &cnao.MacvtapCni{DevicePluginConfig: configMapNameFromOldDays},
}
```
and config has a new value to DevicePluginConfig.
Expected config after `fillMacvtapDefaults`: config has the new value..

test/e2e/workflow/deployment_test.go
```diff
@@ -361,7 +361,7 @@ var _ = Describe("NetworkAddonsConfig", func() {
 			})
 		})
 
-		Context("with an overridden configuration", func() {
+		Context("with a non-default configuration", func() {
```
RamLavi, line 364: I think this belongs to the former commit, can you move it please?

RamLavi (changes requested):

pkg/apis/networkaddonsoperator/shared/networkaddonsconfig_types.go
maiqueb: why would we ? I see the KubeMAcPool / SecondaryDNS configurations do not have it.

pkg/components/components.go
maiqueb: Done.

pkg/network/macvtap_test.go
maiqueb: Done.

test/e2e/workflow/deployment_test.go
maiqueb: Done.

pkg/apis/networkaddonsoperator/shared/networkaddonsconfig_types.go
RamLavi: Was just think aloud, good point.

RamLavi (approved):
/lgtm
/approve

PR #1521, opened 2023-03-21
KSD: Bump to v0.0.10

AlonaKaplan:
**What this PR does / why we need it**:
KSD: Bump to v0.0.10

**Special notes for your reviewer**:

**Release note**:

```release-note
bump ksd to v0.0.10
```

PR #1522, opened 2023-03-22
ksd: Update bump to include CoreDNS as well

oshoval:
Add the missing CoreDNS to KSD bump script.
In case we change the image on KSD, it will be reflected on CNAO as well.

Use the original manifest because the updated one is parameterized already,
and isn't split to yamls (so can't use yaml-utils).

Note that sed won't fail in case `CORE_DNS_IMAGE` isn't found,
we can improve it on the future, it isn't since it doesn't change a lot.
It is the same case for the rest of the images.
 
**Release note**:
```release-note
None
```


oshoval:
/cc @AlonaKaplan 

Depends on your [PR](https://github.com/kubevirt/cluster-network-addons-operator/pull/1521)
Need to rebase once it is merged

oshoval:
/hold 

Depends on the mentioned PR

oshoval:
/cc @RamLavi @AlonaKaplan 

Ready for review
Thanks

RamLavi (approved):
/approve

@AlonaKaplan please give your lgtm

[update] - I see you already added it . never mind then 
/lgtm

oshoval:
Thanks

PR #1547, opened 2023-05-03
Use env RUNBOOK_URL_TEMPLATE for the runbooks URL template

assafad:
When `RUNBOOK_URL_TEMPLATE` env variable is set, its value will be used as the runbooks URL template. Otherwise, keep using the default runbook URL template `https://kubevirt.io/monitoring/runbooks/`. 
This allows using different runbooks versions, in accordance with the installation.

**Which issue(s) this PR fixes**: 
Jira-ticket - https://issues.redhat.com/browse/CNV-14385

**Release note**:

```release-note
None
```


pkg/components/components.go
```diff
@@ -286,6 +287,10 @@ func GetDeployment(version string, operatorVersion string, namespace string, rep
 									Name:  "MONITORING_SERVICE_ACCOUNT",
 									Value: "prometheus-k8s",
 								},
+								{
+									Name:  "RUNBOOK_URL_TEMPLATE",
+									Value: monitoring.GetRunbookURLTemplate(),
```
assafad, line 292: Currently, in an upstream installation, this will get the upstream template (i.e. `https://kubevirt.io/monitoring/runbooks/`) , although `RUNBOOK_URL_TEMPLATE` is not set.
Is this misleading? better to have it empty in upstream?

assafad:
@ormergi, @EdDev Hi, can you please review this PR?

pkg/components/components.go
qinqon: Now we have the template we are forced to put something

qinqon:
@assafad are we passing the RUNBOOK_URL_TEMPLATE env var at HCO on d/s ?

assafad:
> @assafad are we passing the RUNBOOK_URL_TEMPLATE env var at HCO on d/s ?

No, on CNAO - https://gitlab.cee.redhat.com/cpaas-midstream/openshift-virtualization/cluster-network-addons-operator/-/merge_requests/265

pkg/components/components.go
assafad: but the env var won't necessarily be set (in U/S it's not).
So, in U/S, do we want this field to be the actual env var value (empty), or to have the value of the template being used (`https://kubevirt.io/monitoring/runbooks/`)?

PR #1561, opened 2023-05-28
auto bumper: Enable KSD auto bumping

oshoval:
Since https://github.com/go-git/go-git/pull/484 is fixed,
go-git can fetch repos which it couldn't before.
Enable KSD auto bumping, it was disabled because of this problem.

go-git was bumped (v5.7.0) in order to consume the new version.

**Release note**:
```release-note
None
```


phoracek:
/lgtm
/approve

This is great, thanks a lot for adding this automation

PR #1566, opened 2023-06-14
kubemacpool: Remove namespace labels

qinqon:
The kubemacpool project was adding restricted label to test security context. This change remove labels altogether and configure them at CI to test it.

Since the CNAO namespace is shared between components is done there instead of kubemacpool that has it's own namespace.

**Release note**:

```release-note
Remove kubemacpool namespace labels.
```

### DIFF
--- a/automation/components-functests.setup.sh
+++ b/automation/components-functests.setup.sh
@@ -32,6 +32,18 @@ export KUBECONFIG=$(cluster::kubeconfig)
 make cluster-operator-push
 make cluster-operator-install
 
+# Test kubemacpool with restricted
+if [ "$COMPONENT" == "kubemacpool" ]; then
+    cluster/kubectl.sh apply -f - <<EOF
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: cluster-network-addons
+  labels:
+    pod-security.kubernetes.io/enforce: restricted
+EOF
+fi
+
 # Deploy all network addons components with CNAO
     cat <<EOF | cluster/kubectl.sh apply -f -
 apiVersion: networkaddonsoperator.network.kubevirt.io/v1
--- a/data/kubemacpool/kubemacpool.yaml
+++ b/data/kubemacpool/kubemacpool.yaml
@@ -1,9 +1,6 @@
 apiVersion: v1
 kind: Namespace
 metadata:
-  labels:
-    control-plane: mac-controller-manager
-    pod-security.kubernetes.io/enforce: restricted
   name: '{{ .Namespace }}'
 ---
 apiVersion: rbac.authorization.k8s.io/v1
--- a/hack/components/bump-kubemacpool.sh
+++ b/hack/components/bump-kubemacpool.sh
@@ -49,12 +49,16 @@ patches:
     kind: Deployment
     name: mac-controller-manager
     namespace: system
-- path: cnao_mac-range_patch.json
+- path: cnao_mac-range_patch.yaml
   target:
     version: v1
     kind: ConfigMap
     name: mac-range-config
     namespace: system
+- path: cnao_remove-labels_patch.yaml
+  target:
+    version: v1
+    kind: Namespace
 EOF
 
     cat <<EOF > config/cnao/cnao_kubemacpool_manager_patch.yaml
@@ -116,13 +120,17 @@ EOF
   value: "{{ toYaml .Placement.Tolerations | nindent 8 }}"
 EOF
 
-    cat <<EOF > config/cnao/cnao_mac-range_patch.json
+    cat <<EOF > config/cnao/cnao_mac-range_patch.yaml
 - op: replace
   path: /data/RANGE_START
   value: "{{ .RangeStart }}"
 - op: replace
   path: /data/RANGE_END
   value: "{{ .RangeEnd }}"
+EOF
+    cat <<EOF > config/cnao/cnao_remove-labels_patch.yaml
+- op: remove
+  path: /metadata/labels
 EOF
 
 


data/kubemacpool/kubemacpool.yaml
```diff
@@ -1,9 +1,6 @@
 apiVersion: v1
 kind: Namespace
 metadata:
-  labels:
-    control-plane: mac-controller-manager
```
oshoval, line 5: why `control-plane: mac-controller-manager` is removed as well ?
it was not part of the latest breaking change

oshoval:
Worth to enclose the failure of HCO please on PR desc

data/kubemacpool/kubemacpool.yaml
RamLavi: +1, it should stay. I think we already delete it for non-namespaced objects ([here](https://github.com/kubevirt/cluster-network-addons-operator/blob/main/pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go#L537)).

automation/check-patch.e2e-kubemacpool-functests.sh
```diff
@@ -38,6 +38,17 @@ main() {
     source automation/check-patch.setup.sh
     cd ${TMP_PROJECT_PATH}
 
+    kubectl apply -f - <<EOF
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: cluster-network-addons
+  labels:
+    control-plane: mac-controller-manager
```
RamLavi, line 47: I think adding/removing the `control-plane` label should not be part of this PR,

data/kubemacpool/kubemacpool.yaml
oshoval: Btw some of the logic there was in order to maintain backward compatability
maybe the versions we support /deployed on customer site are new enough so we can remove this patch?
(more info here https://github.com/kubevirt/cluster-network-addons-operator/pull/1145)

qinqon: I would not depend on the controller and keep the manifest as we want them to be, so we don't need time for them to be fixed.

hack/components/bump-kubemacpool.sh
```diff
@@ -116,14 +120,18 @@ EOF
...
-    cat <<EOF > config/cnao/cnao_mac-range_patch.json
+    cat <<EOF > config/cnao/cnao_mac-range_patch.yaml
 - op: replace
   path: /data/RANGE_START
   value: "{{ .RangeStart }}"
 - op: replace
   path: /data/RANGE_END
   value: "{{ .RangeEnd }}"
 EOF
+    cat <<EOF > config/cnao/cnao_remove-labels_patch.yaml
```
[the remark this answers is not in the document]
qinqon: I am going to remove all the labels since controller is removing the other one too.

data/kubemacpool/kubemacpool.yaml
oshoval: The controller is removing them only for backward compatibility, for versions where CNAO label stuff that it didn't need to have and then those versions were upgraded to one that fixed the labeling.

You can see here a discussion that tells that we better remove it from the source itself - KMP, IIRC
https://github.com/kubevirt/cluster-network-addons-operator/pull/1145/files#r796372158

oshoval: need to rerun bump right ? since you changed the script

oshoval:
https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_cluster-network-addons-operator/1566/pull-e2e-cnao-multus-dynamic-networks-functests/1668906222926958592
we have this here btw (don't think it is related to this PR IIRC, but maybe worth to double check if interesting)

```
W0614 09:24:17.766599   36399 warnings.go:70] would violate PodSecurity "restricted:latest": allowPrivilegeEscalation != false (container "samplepod" must set securityContext.allowPrivilegeEscalation=false), unrestricted capabilities (container "samplepod" must set securityContext.capabilities.drop=["ALL"]), runAsNonRoot != true (pod or container "samplepod" must set securityContext.runAsNonRoot=true), seccompProfile (pod or container "samplepod" must set securityContext.seccompProfile.type to "RuntimeDefault" or "Localhost")
```

RamLavi (approved):
/lgtm

I think that removing all the labels in the yaml patch may bite us in the future, but we can remove the `control-plane` label  from the NS on the KMP repo in parallel, then consider refining the yaml patch to remove exactly the labels we want.

oshoval:
it still fails

qinqon:
hiccup 
```
pull-e2e-cluster-network-addons-operator-kubemacpool-functest
```

> /lgtm
> 
> I think that removing all the labels in the yaml patch may bite us in the future, but we can remove the `control-plane` label from the NS on the KMP repo in parallel, then consider refining the yaml patch to remove exactly the labels we want.

Is very dangerous to label the shared namespace, it affects all the network components, we should remove the label, if we see kubemacpool depends on something, we will fix it.

oshoval:
Sumup here for future possible AI, thanks

https://github.com/kubevirt/cluster-network-addons-operator/issues/1567

automation/components-functests.setup.sh
```diff
@@ -32,6 +32,18 @@ export KUBECONFIG=$(cluster::kubeconfig)
 make cluster-operator-push
 make cluster-operator-install
 
+# Test kubemacpool with restricted
+if [ "$COMPONENT" == "kubemacpool" ]; then
+    cluster/kubectl.sh apply -f - <<EOF
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: cluster-network-addons
```
oshoval, line 41: nit: we don't need to parameterize the ns name ?
anyhow if needed can be done on follow PR

PR #1568, opened 2023-06-18
Kubemacpool: Parameterize "runAsNonRoot" in the yaml

RamLavi:
In order to make kubemacpool pods work with restricted psa policy it should have runAsNonRoot: true.
However, once the pod's securityContext has runAsNonRoot: true it should also have a specific rusAsUser.
Having a specific user is problematic on Openshift since it may configure a range of valid ids (example of a possible error - spec.containers[0].securityContext.runAsUser: Invalid value: 127: must be in the ranges: [1000670000, 1000679999]).
(issue [here](https://issues.redhat.com/browse/CNV-29823))

According to "What happens if I use runAsUser in my workloads?" section in https://github.com/redhat-openshift-ecosystem/community-operators-prod/discussions/1417 the suggested solution is -
```
If you are unable to change the image, for Openshift distributions you MUST leave the RunAsUser and RunAsNonRoot fields empty and let the Openshift SCC inject the values when your workload is qualified to the SCC policy.
```
In order to support both cases where SCC is deployed or not, this is the solution chosen.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/network/kubemacpool.go
+++ b/pkg/network/kubemacpool.go
@@ -96,7 +96,7 @@ func changeSafeKubeMacPool(prev, next *cnao.NetworkAddonsConfigSpec) []error {
 }
 
 // renderLinuxBridge generates the manifests of Linux Bridge
-func renderKubeMacPool(conf *cnao.NetworkAddonsConfigSpec, manifestDir string) ([]*unstructured.Unstructured, error) {
+func renderKubeMacPool(conf *cnao.NetworkAddonsConfigSpec, manifestDir string, clusterInfo *ClusterInfo) ([]*unstructured.Unstructured, error) {
 	if conf.KubeMacPool == nil {
 		return nil, nil
 	}
--- a/pkg/network/kubemacpool.go
+++ b/pkg/network/kubemacpool.go
@@ -115,6 +115,14 @@ func renderKubeMacPool(conf *cnao.NetworkAddonsConfigSpec, manifestDir string, c
 	data.Data["CertRotateInterval"] = conf.SelfSignConfiguration.CertRotateInterval
 	data.Data["CertOverlapInterval"] = conf.SelfSignConfiguration.CertOverlapInterval
 
+	if clusterInfo.SCCAvailable {
+		data.Data["RunAsNonRoot"] = "null"
+		data.Data["RunAsUser"] = "null"
+	} else {
+		data.Data["RunAsNonRoot"] = "true"
+		data.Data["RunAsUser"] = "107"
+	}
+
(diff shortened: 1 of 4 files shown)


RamLavi:
getting `could not apply (apps/v1, Kind=Deployment) cluster-network-addons/kubemacpool-cert-manager: could not create (apps/v1, Kind=Deployment) cluster-network-addons/kubemacpool-cert-manager: Deployment in version "v1" cannot be handled as a Deployment: json: cannot unmarshal string into Go struct field PodSecurityContext.spec.template.spec.securityContext.runAsNonRoot of type bool`

also tried to use the more specific parameter types:
```
var nonRootTrue = true
var nonRootUser int64 = 107
data.Data["RunAsNonRoot"] = &nonRootTrue
data.Data["RunAsUser"] = &nonRootUser
```
didn't work as well..

@oshoval did you encounter this issue? I would expect ksd to have the same issue..

oshoval:
There is no quote here
https://github.com/kubevirt/cluster-network-addons-operator/pull/1499/files#diff-8041a1f861576736acd26091b5c7806c2735246aa93d254b335c5bd8795e8255R79

Maybe try to align the same thing ?
I don't know, because Alona worked on that, please ask her

pkg/network/kubemacpool.go
```diff
@@ -115,6 +115,14 @@ func renderKubeMacPool(conf *cnao.NetworkAddonsConfigSpec, manifestDir string) (
 	data.Data["CertRotateInterval"] = conf.SelfSignConfiguration.CertRotateInterval
 	data.Data["CertOverlapInterval"] = conf.SelfSignConfiguration.CertOverlapInterval
 
+	if clusterInfo.SCCAvailable {
+		data.Data["RunAsNonRoot"] = "null"
+		data.Data["RunAsUser"] = "null"
```
qinqon, line 120: this is the same as not set them up ? I would prefer not to set it at all instead of "null"

qinqon (changes requested):

RamLavi:
> There is no quote here https://github.com/kubevirt/cluster-network-addons-operator/pull/1499/files#diff-8041a1f861576736acd26091b5c7806c2735246aa93d254b335c5bd8795e8255R79

yep, that was it. But sorry to bother you I thought you worked on it

oshoval:
> > There is no quote here https://github.com/kubevirt/cluster-network-addons-operator/pull/1499/files#diff-8041a1f861576736acd26091b5c7806c2735246aa93d254b335c5bd8795e8255R79
> 
> yep, that was it. But sorry to bother you I thought you worked on it

all good, gg

oshoval:
/retest 

podman issues, need to bump job image for those

qinqon (approved):
/lgtm
/approve

PR #1573, opened 2023-06-27
linux bridge: Consolidate CNIs

oshoval:
Some Kubernetes vendors deploy their own bridge and tuning CNI binaries.
In such cases, CNAO can just rely on these binaries instead of deploying
(and maintaining its own).

In case an existing binary is found, create a soft link between
bridge and cnv-bridge. Otherwise build and deploy cnv-bridge.
Soft link it to bridge.

With this change, we maintain backwards compatibility, while avoiding the
need to support bridge and tuning CNIs where we don't have to.

**Release note**:
```release-note
None
```

### DIFF
--- a/data/linux-bridge/002-linux-bridge.yaml
+++ b/data/linux-bridge/002-linux-bridge.yaml
@@ -39,23 +39,24 @@ spec:
             - /bin/bash
             - -ce
             - |
-              echo 'Installing bridge and tuning CNIs'
               cni_mount_dir=/opt/cni/bin
-              sourcebinpath=/usr/src/github.com/containernetworking/plugins/bin
-              cp --remove-destination ${sourcebinpath}/bridge ${cni_mount_dir}/cnv-bridge
-              cp --remove-destination ${sourcebinpath}/tuning ${cni_mount_dir}/cnv-tuning
+              components=("bridge" "tuning")
+              for component in "${components[@]}"; do
+                if find "${cni_mount_dir}/${component}" &>/dev/null && ! test -L "${cni_mount_dir}/${component}"; then
+                  echo "${component} binary found, creating symbolic link ${cni_mount_dir}/cnv-${component}"
+                  ln -sf "${cni_mount_dir}/${component}" "${cni_mount_dir}/cnv-${component}"
+                else
+                  echo "installing cnv-${component}, and creating symbolic link ${cni_mount_dir}/${component}"
 
-              echo 'Checking bridge and tuning CNIs deployment on node'
-              printf -v bridgechecksum "%s" "$(<$sourcebinpath/bridge.checksum)"
-              printf -v tuningchecksum "%s" "$(<$sourcebinpath/tuning.checksum)"
-              printf "%s %s" "${bridgechecksum% *}" "${cni_mount_dir}/cnv-bridge" | sha256sum --check
-              printf "%s %s" "${tuningchecksum% *}" "${cni_mount_dir}/cnv-tuning" | sha256sum --check
+                  sourcebinpath="/usr/src/github.com/containernetworking/plugins/bin"
+                  cp --remove-destination "${sourcebinpath}/${component}" "${cni_mount_dir}/cnv-${component}"
 
-              # Some projects (e.g. openshift/console) use cnv- prefix to distinguish between
-              # binaries shipped by OpenShift and those shipped by KubeVirt (D/S matters).
-              # Following two lines make sure we will provide both names when needed.
-              find ${cni_mount_dir}/bridge &>/dev/null || ln -s ${cni_mount_dir}/cnv-bridge ${cni_mount_dir}/bridge
-              find ${cni_mount_dir}/tuning &>/dev/null || ln -s ${cni_mount_dir}/cnv-tuning ${cni_mount_dir}/tuning
+                  printf -v component_checksum "%s" "$(<${sourcebinpath}/${component}.checksum)"
+                  printf "%s %s" "${component_checksum% *}" "${cni_mount_dir}/cnv-${component}" | sha256sum --check
+
+                  ln -sf "${cni_mount_dir}/cnv-${component}" "${cni_mount_dir}/${component}"
+                fi
+              done
               echo 'Entering sleep... (success)'
               sleep infinity
           resources:


data/linux-bridge/002-linux-bridge.yaml
```diff
@@ -39,23 +39,20 @@ spec:
...
-              cp --remove-destination ${sourcebinpath}/bridge ${cni_mount_dir}/cnv-bridge
-              cp --remove-destination ${sourcebinpath}/tuning ${cni_mount_dir}/cnv-tuning
+              if find ${cni_mount_dir}/bridge &>/dev/null; then
+                ln -s ${cni_mount_dir}/bridge ${cni_mount_dir}/cnv-bridge
+                ln -s ${cni_mount_dir}/tuning ${cni_mount_dir}/cnv-tuning
+              else
+                cni_mount_dir=/opt/cni/bin
+                sourcebinpath=/usr/src/github.com/containernetworking/plugins/bin
+                cp --remove-destination ${sourcebinpath}/bridge ${cni_mount_dir}/cnv-bridge
+                cp --remove-destination ${sourcebinpath}/tuning ${cni_mount_dir}/cnv-tuning
```
phoracek, line 49: We still need to deploy the `bridge` and `tuning` binaries as we used to. Otherwise KubeVirt e2e tests will start failing and users won't be able to use their old NetworkAttachmentDefinitions. To maintain the existing behavior, we still need to install `bridge` and `cnv-bridge` on non-OpenShift environments

data/linux-bridge/002-linux-bridge.yaml
```diff
@@ -39,23 +39,20 @@ spec:
             - /bin/bash
             - -ce
             - |
-              echo 'Installing bridge and tuning CNIs'
-              cni_mount_dir=/opt/cni/bin
-              sourcebinpath=/usr/src/github.com/containernetworking/plugins/bin
-              cp --remove-destination ${sourcebinpath}/bridge ${cni_mount_dir}/cnv-bridge
-              cp --remove-destination ${sourcebinpath}/tuning ${cni_mount_dir}/cnv-tuning
+              if find ${cni_mount_dir}/bridge &>/dev/null; then
```
phoracek, line 42: This won't be enough if the comment below gets addressed.

phoracek (changes requested):

oshoval:
please retake a look
thanks

data/linux-bridge/002-linux-bridge.yaml
oshoval: changed https://github.com/kubevirt/cluster-network-addons-operator/compare/cd82a3236f8ffba41a0862f91456e7b1c7518811..419fedfbc1b9b72932e6c4ec63b35d836c2100f1

but willing to hear more ideas, thanks

oshoval:
ithink we should fix the fact pull-e2e-cnao-multus-dynamic-networks-functests  is flaky

data/linux-bridge/002-linux-bridge.yaml
oshoval: addressed the comment below
can you please explain why it is not enough?
any other alternatives?

phoracek: With this condition we would never update the plugins.

On the first run of CNAO on a Kubernetes cluster, the else branch would install `bridge` and `cnv-bridge`. On any consecutive run, this condition would be evaluated as true. Thus, we will never update existing binaries with newer ones.

I simple solution for this would be to include a "flag" in the container image. Our upstream container image would stay as is. Any downstream image that wants to only symlink could `touch /opt/symlink-cnis`. The condition in this daemon set would look for that file.

oshoval: interesting,
but would be better to touch only upstream if possible,
we can add some kind of mechanism
that if at the first time there was no `bridge`, it touches `${cni_mount_dir}/upstream`
with this we can change the condition to be `&&` the above file doesn't exist
wdyt?

phoracek: Let's try it an see how it looks

oshoval:
addressed comments

data/linux-bridge/002-linux-bridge.yaml
```diff
@@ -39,23 +39,23 @@ spec:
...
-              # Following two lines make sure we will provide both names when needed.
-              find ${cni_mount_dir}/bridge &>/dev/null || ln -s ${cni_mount_dir}/cnv-bridge ${cni_mount_dir}/bridge
-              find ${cni_mount_dir}/tuning &>/dev/null || ln -s ${cni_mount_dir}/cnv-tuning ${cni_mount_dir}/tuning
+                printf -v bridgechecksum "%s" "$(<$sourcebinpath/bridge.checksum)"
+                printf -v tuningchecksum "%s" "$(<$sourcebinpath/tuning.checksum)"
+                printf "%s %s" "${bridgechecksum% *}" "${cni_mount_dir}/cnv-bridge" | sha256sum --check
+                printf "%s %s" "${tuningchecksum% *}" "${cni_mount_dir}/cnv-tuning" | sha256sum --check
+                find ${cni_mount_dir}/bridge &>/dev/null || ln -s ${cni_mount_dir}/cnv-bridge ${cni_mount_dir}/bridge
+                find ${cni_mount_dir}/bridge &>/dev/null || ln -s ${cni_mount_dir}/cnv-tuning ${cni_mount_dir}/tuning
+              fi
```
oshoval, line 58: in case of redeployment, wouldn't this keep the first bridge that was created ?
we can use 'ln -sf` for upstream if desired

data/linux-bridge/002-linux-bridge.yaml
oshoval: instead of a flag, we can check if the `bridge` is a real file or softlink, wdyt?
less files is better

oshoval:
Using `test` instead of a file, less files are better imho.
Using `ln -sf` to force override the soft link, this will make it more predictable.

oshoval:
Would be nice to declare please EOL for "tuning" from some point, wdyt ?

data/linux-bridge/002-linux-bridge.yaml
```diff
@@ -39,23 +39,23 @@ spec:
...
-              cni_mount_dir=/opt/cni/bin
-              sourcebinpath=/usr/src/github.com/containernetworking/plugins/bin
-              cp --remove-destination ${sourcebinpath}/bridge ${cni_mount_dir}/cnv-bridge
-              cp --remove-destination ${sourcebinpath}/tuning ${cni_mount_dir}/cnv-tuning
+              if [ find ${cni_mount_dir}/bridge &>/dev/null ] && [ ! test -L ${cni_mount_dir}/bridge ]; then
+                ln -s ${cni_mount_dir}/bridge ${cni_mount_dir}/cnv-bridge
+                ln -s ${cni_mount_dir}/tuning ${cni_mount_dir}/cnv-tuning
+              else
+                cni_mount_dir=/opt/cni/bin
+                sourcebinpath=/usr/src/github.com/containernetworking/plugins/bin
```
phoracek, line 47: Does not this need to be above the first if?

phoracek (changes requested):
Please make sure to test this (at least manually) on both Kubernetes and OCP

data/linux-bridge/002-linux-bridge.yaml
oshoval: why? it is used only below this line, and the if (of line 42 checks only `${cni_mount_dir}/bridge`
please correct me if you want something else

oshoval: You are right
`cni_mount_dir=/opt/cni/bin` should be before the if, (or just use hardcoded `/opt/cni/bin`)
Thanks

oshoval:
Fixed according comments, (used anchors if its ok)
Will update here once tested both on k8s and CRC (if suitable, otherwise OCP)
Thanks

data/linux-bridge/002-linux-bridge.yaml
```diff
@@ -39,23 +39,23 @@ spec:
...
             - |
-              echo 'Installing bridge and tuning CNIs'
               cni_mount_dir=/opt/cni/bin
-              sourcebinpath=/usr/src/github.com/containernetworking/plugins/bin
-              cp --remove-destination ${sourcebinpath}/bridge ${cni_mount_dir}/cnv-bridge
-              cp --remove-destination ${sourcebinpath}/tuning ${cni_mount_dir}/cnv-tuning
+              if find ${cni_mount_dir}/bridge &>/dev/null && ! test -L ${cni_mount_dir}/bridge; then
+                ln -s ${cni_mount_dir}/bridge ${cni_mount_dir}/cnv-bridge
+                ln -s ${cni_mount_dir}/tuning ${cni_mount_dir}/cnv-tuning
+              else
```
oshoval, line 46: we used to have conditional soft link creation
`find ${cni_mount_dir}/bridge &>/dev/null || ln -s ${cni_mount_dir}/cnv-bridge ${cni_mount_dir}/bridge`
but now we don't

and even on U/S we have already binaries `bridge`, `tuning`, that are installed
as part of provision by this command
```
dnf install --skip-broken --nobest --nogpgcheck --disableexcludes=kubernetes -y \
    kubeadm-${version} \
    kubelet-${version} \
    kubectl-${version} \
    kubernetes-cni
```
(note that removing `kubernetes-cni` wont help, it will still be installed)

It means that until today on U/S, we had bridge that belong to k8s and cnv-bridge that was installed by us.

On OCP it will be fine, but we need to decide please what to do on U/S

Maybe we are fine to keep U/S and D/S the same in case `bridge` binary exists ? (best imho if covers every need)
or we want to try and delete those two binaries, during cluster-up of CNAO (but it is prone to be less predictable)

oshoval:
used `ln -sf` on the "if" branch, because otherwise, when redeployed it will fail to create the soft link, as it already exists

data/linux-bridge/002-linux-bridge.yaml
phoracek: If we use the OpenShift strategy on Kubernetes, we may damage existing users who don't deploy Kubernetes from the RPMs you shared above. We need to keep shipping and updating our binaries, at least for now.

What about this logic:

```
if cnv-bridge symlink exists:
    do nothing
elif bridge binary exists:
   create cnv-bridge symlink
else:
  install cnv-bridge binary
  create bridge symlink
```

This way we should:
* Maintain the existing behavior on Kubernetes
* Always symlink to `bridge` binary if it is not installed by us
* Keep updating the binary if it is deployed by us
* Not need any additional files or flags

Is there any gap in my logic? @oshoval

oshoval: Thanks

Few notes to consider please (maybe it is fine, just raising)

It won't keep the current logic on kubevirtci
because bridge exists so it will "create cnv-bridge symlink"
before this PR, we did install cnv-network binary, but kept bridge binary as is

"If we use the OpenShift strategy on Kubernetes, we may damage existing users who don't deploy Kubernetes from the RPMs you shared above"
if they don't deploy it and don't have bridge, we install cnv-network for them and then create bridge softlink
isn't it fine ?

beside that sounds good

oshoval: Here is the code, the only change is the first if
i feel it isn't adding anything major comparing to current PR state, as it is already encapsulated in the 2nd condition ?
```
              cni_mount_dir=/opt/cni/bin
              if find ${cni_mount_dir}/cnv-bridge &>/dev/null && test -L ${cni_mount_dir}/cnv-bridge; then
                echo "cnv-bridge symbolic link found, nothing to be done"
              if find ${cni_mount_dir}/bridge &>/dev/null && ! test -L ${cni_mount_dir}/bridge; then
                echo "bridge binary found, creating cnv-bridge symbolic link"
                ln -sf ${cni_mount_dir}/bridge ${cni_mount_dir}/cnv-bridge
                ln -sf ${cni_mount_dir}/tuning ${cni_mount_dir}/cnv-tuning
              else
                echo "installing cnv-bridge and creating bridge symbolic link"
                sourcebinpath=/usr/src/github.com/containernetworking/plugins/bin
                cp --remove-destination ${sourcebinpath}/bridge ${cni_mount_dir}/cnv-bridge
                cp --remove-destination ${sourcebinpath}/tuning ${cni_mount_dir}/cnv-tuning

                printf -v bridgechecksum "%s" "$(<$sourcebinpath/bridge.checksum)"
                printf -v tuningchecksum "%s" "$(<$sourcebinpath/tuning.checksum)"
                printf "%s %s" "${bridgechecksum% *}" "${cni_mount_dir}/cnv-bridge" | sha256sum --check
                printf "%s %s" "${tuningchecksum% *}" "${cni_mount_dir}/cnv-tuning" | sha256sum --check

                ln -sf ${cni_mount_dir}/cnv-bridge ${cni_mount_dir}/bridge
                ln -sf ${cni_mount_dir}/cnv-tuning ${cni_mount_dir}/tuning
              fi
              echo 'Entering sleep... (success)'
              sleep infinity
```

phoracek: Ah you are right, the only difference would be that with 3 conditions we would not recreate the symlink, right?

I'm fine with the 2 condition approach.

phoracek: > 
> > I'm fine with the 2 condition approach.
> 
> so the PR as is right ? it only means that we won't deploy cnv-bridge on kubevirtci upstream atm, because bridge binary is there are we fine with it ? do we want to try and delete the binary ?

I'm fine with that.

Please make sure to test this change on Kubernetes and on OCP in various scenarios. I'm ok without all of them being automated, but let's at least try it manually.

oshoval: all looks fine

on CRC it has bridge binary and creates cnv-bridge sym link

on U/S it also has bridge binary and creates cnv-bridge sym link
and if i manually delete the bridge binary (and any cnv-bridge sym link leftovers from previous run)
it will deploy cnv bridge binary and create bridge symlink this time as expected

data/linux-bridge/002-linux-bridge.yaml
```diff
@@ -39,23 +39,23 @@ spec:
             - /bin/bash
             - -ce
             - |
-              echo 'Installing bridge and tuning CNIs'
               cni_mount_dir=/opt/cni/bin
-              sourcebinpath=/usr/src/github.com/containernetworking/plugins/bin
-              cp --remove-destination ${sourcebinpath}/bridge ${cni_mount_dir}/cnv-bridge
-              cp --remove-destination ${sourcebinpath}/tuning ${cni_mount_dir}/cnv-tuning
+              if find ${cni_mount_dir}/bridge &>/dev/null && ! test -L ${cni_mount_dir}/bridge; then
```
phoracek, line 43: Let's not make assumptions about the tuning also accompanying bridge. Please split this into two if-elses, one for bridge, one for tuning.

Also please add echos explaining what happens and why, to make it easier for cluster admins to understand what's going on (and what are our intentions) while looking in the logs.

phoracek (changes requested):
The logic looks good to me. Just some comments about logs, commit message and structure.

Please make the commit message more generic:

Some Kubernetes vendors deploy their own bridge and tuning CNI binaries.
In such cases, CNAO can just rely on these binaries instead of deploying
(and maintaining its own).

In case an existing binary is found, create a soft link between
bridge and cnv-bridge. Otherwise build and deploy cnv-bridge.
Soft link it to bridge.

With this change, we maintain backwards compatibility, while avoiding the
need to support bridge and tuning CNIs where we don't have to.

data/linux-bridge/002-linux-bridge.yaml
oshoval: Done

oshoval:
Addressed comments

oshoval:
are we fine that this ovs-mirror fails ?
https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_cluster-network-addons-operator/1573/pull-e2e-cluster-network-addons-operator-ovs-cni-functests/1683032870244847616

we have now bridge binary, and cnv-bridge symbol
before this PR we had both binaries
maybe it doesnt affect the ovs tests ...

oshoval:
/test pull-e2e-cluster-network-addons-operator-ovs-cni-functests 

retesting

phoracek:
Yeah, the OVS mirror CNI is flaky for a long time... I have to set aside some time to get it resolved on https://github.com/k8snetworkplumbingwg/ovs-cni (it's problematic there too).

phoracek (approved):
/lgtm
/approve

Thanks a lot!

PR #1587, opened 2023-07-19
Allow restriction of Multus RBAC

phoracek:

The Multus component requires an a powerful cluster-wide role to
operate:

```
  - apiGroups:
    - k8s.cni.cncf.io
    resources:
    - '*'
    verbs:
    - '*'
```

At the same time, Multus is an optional component.

As part of an effort to limit CNAO to the least amount of privileges it
must posses, this patch introduces a new flag to the CNAO manifest
generator, allowing to omit the Multus RBAC from the CNAO RBAC. Default
value of the flag is true, in this case Multus RBAC will be included.

* If this new flag is set to false and Multus is never deployed, it
  limits the possible attack surface.
* If this flag is set to false and a user attempts to deploy Multus,
  CNAO will turn into a degraded state, unable to deploy Multus because
  of missing RBAC permissions.

Typically, this flag should set to false on OpenShift, where Multus is
managed by the CNO, and thus these additional RBAC permissions on CNAO
aren't necessary.

This PR does not change any existing behavior. It only allows integrators to tweak CNAO manifests per their needs.

This change was successfully tested on OpenShift CRC as well.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/components/components.go
+++ b/pkg/components/components.go
@@ -463,7 +463,7 @@ func GetRole(namespace string) *rbacv1.Role {
 	return role
 }
 
-func GetClusterRole() *rbacv1.ClusterRole {
+func GetClusterRole(allowMultus bool) *rbacv1.ClusterRole {
 	role := &rbacv1.ClusterRole{
 		TypeMeta: metav1.TypeMeta{
 			APIVersion: "rbac.authorization.k8s.io/v1",
@@ -657,12 +657,15 @@ func GetClusterRole() *rbacv1.ClusterRole {
 	}
 
 	role.Rules = append(role.Rules, componentsClusterRoles()...)
+	if allowMultus {
+		role.Rules = append(role.Rules, multusClusterRoles()...)
+	}
 
 	return role
 }
 
 func componentsClusterRoles() []rbacv1.PolicyRule {
-	rules := []rbacv1.PolicyRule{
+	return []rbacv1.PolicyRule{
 		{
 			APIGroups: []string{
 				"",
@@ -700,17 +703,6 @@ func componentsClusterRoles() []rbacv1.PolicyRule {
 				"create",
 				"patch",
 				"update",
-			},
-		},
-		{
-			APIGroups: []string{
-				"k8s.cni.cncf.io",
-			},
-			Resources: []string{
-				"*",
-			},
-			Verbs: []string{
-				"*",
 			},
 		},
 		{
@@ -857,8 +849,25 @@ func componentsClusterRoles() []rbacv1.PolicyRule {
 			},
 		},
 	}
+}
 
-	return rules
+// Multus CNCF rules are kept separate and optional to allow vendors
+// who ship Multus through different channels to not deploy Multus'
+// highly privileged cluster-wide RBAC with CNAO.
+func multusClusterRoles() []rbacv1.PolicyRule {
+	return []rbacv1.PolicyRule{
+		{
+			APIGroups: []string{
+				"k8s.cni.cncf.io",
+			},
+			Resources: []string{
+				"*",
+			},
+			Verbs: []string{
+				"*",
+			},
+		},
+	}
 }
 
 func GetCrd() *extv1.CustomResourceDefinition {
--- a/tools/manifest-templator/manifest-templator.go
+++ b/tools/manifest-templator/manifest-templator.go
@@ -136,7 +136,7 @@ func marshallObject(obj interface{}, writer io.Writer) error {
 	return nil
 }
 
-func getCNA(data *templateData) {
+func getCNA(data *templateData, allowMultus bool) {
 	writer := strings.Builder{}
 
 	// Get CNA Deployment
@@ -177,7 +177,7 @@ func getCNA(data *templateData) {
 
 	// Get CNA ClusterRole
 	writer = strings.Builder{}
-	clusterRole := components.GetClusterRole()
+	clusterRole := components.GetClusterRole(allowMultus)
 	marshallObject(clusterRole, &writer)
 	clusterRoleString := writer.String()
 
@@ -240,6 +240,7 @@ func main() {
 	imageName := flag.String("image-name", components.Name, "The operator image's name")
 	containerTag := flag.String("container-tag", "latest", "The operator image's container tag")
 	imagePullPolicy := flag.String("image-pull-policy", "Always", "The pull policy to use on the operator image")
+	allowMultus := flag.Bool("allow-multus", true, "Install Multus RBAC, making it possible to deploy Multus through the operator (should be disabled on OpenShift to limit CNAO's RBAC to the necessary minimum)")
 	multusImage := flag.String("multus-image", components.MultusImageDefault, "The multus image managed by CNA")
 	linuxBridgeCniImage := flag.String("linux-bridge-cni-image", components.LinuxBridgeCniImageDefault, "The linux bridge cni image managed by CNA")
 	linuxBridgeMarkerImage := flag.String("linux-bridge-marker-image", components.LinuxBridgeMarkerImageDefault, "The linux bridge marker image managed by CNA")
@@ -280,7 +281,7 @@ func main() {
 	}
 
 	// Load in all CNA Resources
-	getCNA(&data)
+	getCNA(&data, *allowMultus)
 
 	if *inputFile == "" {
 		panic("Must specify input file")


pkg/components/components.go
```diff
@@ -463,7 +463,7 @@ func GetRole(namespace string) *rbacv1.Role {
 	return role
 }
 
-func GetClusterRole() *rbacv1.ClusterRole {
+func GetClusterRole(allowMultus bool) *rbacv1.ClusterRole {
```
oshoval, line 466: Thanks

nit: what about a new function that adds the rbac according needs, instead passing this argument ?

phoracek:
@oshoval thank you for the review. I have made the requested change. Let me know if this is what you had in mind.

pkg/components/components.go
phoracek: I did it on a lower level.

Now when I read your comment again, I think you wanted the if to me moved to the templator instead?

oshoval: You did exactly as i meant :) thanks
unless you think there is a better way (i.e in the templator as you suggested)
this is good imho

pkg/components/components.go
```diff
@@ -657,12 +657,15 @@ func GetClusterRole() *rbacv1.ClusterRole {
 	}
 
 	role.Rules = append(role.Rules, componentsClusterRoles()...)
+	if allowMultus {
+		role.Rules = append(role.Rules, multusClusterRoles()...)
+	}
```
oshoval, line 662: nit: would be nice that it will be before line 659
this way it is more encapsulated by groups
as multus is closer to its real place where it belongs

even one blank line in between can be added additionally

wdyt ?

tools/manifest-templator/manifest-templator.go
```diff
@@ -240,6 +240,7 @@ func main() {
 	imageName := flag.String("image-name", components.Name, "The operator image's name")
 	containerTag := flag.String("container-tag", "latest", "The operator image's container tag")
 	imagePullPolicy := flag.String("image-pull-policy", "Always", "The pull policy to use on the operator image")
+	allowMultus := flag.Bool("allow-multus", true, "Install Multus RBAC, making it possible to deploy Multus through the operator (should be disabled on OpenShift to limit CNAO's RBAC to the necessary minimum)")
```
oshoval, line 243: PR desc does says what happen with and without the flag but
worth to mention please on PR desc that Openshift must disable it
as this comment says

oshoval (commented):
Thanks

PR desc typo newer -> never

oshoval (approved):
Thanks

just worth to add imo this info to PR desc

https://github.com/kubevirt/cluster-network-addons-operator/pull/1587/files#r1269084079

phoracek:
Thanks for the review @oshoval. I addressed all of you comments. Please take another look when you find some time.

oshoval:
Thanks
PR desc / commit message - please fix "are an unnecessary." -> "aren't unnecessary."

Consider please adding what CNO stands for, CNO (Openshift's Cluster Network Operator)
because we have both CNAO and CNO mentioned, and the first time i saw it it was confusing a bit.

RamLavi:
> If this new flag is used and Multus is never deployed, it limits the possible surface area. If this flag is used and a user attempts to deploy Multus, CNAO will turn into a degraded state, unable to install Multus' RBAC.

This suggests the flag omits when used (i.e. flag ON --> no RBAC).

> 
> Typically, this flag should be disabled on OpenShift, where Multus is managed by the CNO, and thus these additional RBAC permissions on CNAO are an unnecessary.

This suggest the flag does not omit when used (i.e. flag OFF --> no RBAC)..

oshoval:
> Having trouble understanding the PR Desc & commit message:
> 
> > If this new flag is used and Multus is never deployed, it limits the possible surface area. If this flag is used and a user attempts to deploy Multus, CNAO will turn into a degraded state, unable to install Multus' RBAC.
> 
> This suggests the flag omits when used (i.e. flag ON --> no RBAC).
> 
> > Typically, this flag should be disabled on OpenShift, where Multus is managed by the CNO, and thus these additional RBAC permissions on CNAO are an unnecessary.
> 
> This suggest the flag does not omit when used (i.e. flag OFF --> no RBAC)..

it means that the presenting the flag disabling multus, but i understand the working might be confusing

RamLavi:
> > Having trouble understanding the PR Desc & commit message:
> > > If this new flag is used and Multus is never deployed, it limits the possible surface area. If this flag is used and a user attempts to deploy Multus, CNAO will turn into a degraded state, unable to install Multus' RBAC.
> > 
> > 
> > This suggests the flag omits when used (i.e. flag ON --> no RBAC).
> > > Typically, this flag should be disabled on OpenShift, where Multus is managed by the CNO, and thus these additional RBAC permissions on CNAO are an unnecessary.
> > 
> > 
> > This suggest the flag does not omit when used (i.e. flag OFF --> no RBAC)..
> 
> it means that the presenting the flag disabling multus, but i understand the working might be confusing

May I suggest this phrasing:
```
As part of an effort to limit CNAO to the least amount of privileges it must posses, this patch introduces a new flag "allow-multus" to the CNAO manifest generator, allowing to permit/omit the Multus RBAC from the CNAO RBAC.

If this new flag is set to `true` and Multus is never deployed, it limits the possible surface area. If this flag is set to `false` and a user attempts to deploy Multus, CNAO will turn into a degraded state, unable to install Multus' RBAC.

Typically, this flag should be set to `false` on OpenShift, where Multus is
managed by the CNO, and thus these additional RBAC permissions on CNAO
are an unnecessary.
```

oshoval:
> > > Having trouble understanding the PR Desc & commit message:
> > > > If this new flag is used and Multus is never deployed, it limits the possible surface area. If this flag is used and a user attempts to deploy Multus, CNAO will turn into a degraded state, unable to install Multus' RBAC.
> > > 
> > > 
> > > This suggests the flag omits when used (i.e. flag ON --> no RBAC).
> > > > Typically, this flag should be disabled on OpenShift, where Multus is managed by the CNO, and thus these additional RBAC permissions on CNAO are an unnecessary.
> > > 
> > > 
> > > This suggest the flag does not omit when used (i.e. flag OFF --> no RBAC)..
> > 
> > 
> > it means that the presenting the flag disabling multus, but i understand the working might be confusing
> 
> May I suggest this phrasing:
> 
> ```
> As part of an effort to limit CNAO to the least amount of privileges it must posses, this patch introduces a new flag "allow-multus" to the CNAO manifest generator, allowing to permit/omit the Multus RBAC from the CNAO RBAC.
> 
> corner cases:
> - If this flag is set to `true` and Multus is never deployed, it limits the possible surface area. 
> - If this flag is set to `false` and a user attempts to deploy Multus, CNAO will turn into a degraded state, unable to install Multus' RBAC.
> 
> Typically, this flag should be set to `false` on OpenShift, where Multus is
> managed by the CNO, and thus these additional RBAC permissions on CNAO
> are an unnecessary.
> ```

Thanks, imho we can indeed improve but i wouldn't use the corner cases (even that they do appear)
but something like

```
As part of an effort to limit CNAO to the least amount of privileges it must posses, this patch introduces a new flag to the CNAO manifest generator, allowing to omit the Multus RBAC from the CNAO RBAC.
Default value of the flag is `true`, in this case Multus RBAC will be included.

If this new flag is set to `false` and Multus is never deployed, it limits the possible surface area.
If this flag is set to `false` and a user attempts to deploy Multus, CNAO will turn into a degraded state, unable to deploy  Multus because of missing RBAC permissions.

Typically, this flag should set to `false` on OpenShift, where Multus is
managed by the CNO, and thus these additional RBAC permissions on CNAO aren't necessary.
```

wdyt ?

phoracek:
@oshoval @RamLavi fixed. Please take a look

oshoval (approved):
Thanks

Nit to consider please - adding a new line before the word `Default`

```
generator, allowing to omit the Multus RBAC from the CNAO RBAC. Default
value of the flag is true, in this case Multus RBAC will be included.
```
->
```
generator, allowing to omit the Multus RBAC from the CNAO RBAC. 
Default value of the flag is true, in this case Multus RBAC will be included.
```

phoracek:
> Nit to consider please - adding a new line before the word `Default`

It is a single paragraph aligned to the recommended commit message length. I don't want to fiddle with it, it does not add much to the readability, there is no clear rule for it, and it would require people who let their IDE format the commit message to rewrite it manually.

PR #1592, opened 2023-08-02
linux bridge: Fix softlink path

oshoval:
The soft link should point to the real path on the host, 
not to the mount path of the container.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=2228240

**Release note**:
```release-note
None
```

### DIFF
--- a/data/linux-bridge/002-linux-bridge.yaml
+++ b/data/linux-bridge/002-linux-bridge.yaml
@@ -40,11 +40,12 @@ spec:
             - -ce
             - |
               cni_mount_dir=/opt/cni/bin
+              host_dir={{ .CNIBinDir }}
               components=("bridge" "tuning")
               for component in "${components[@]}"; do
                 if find "${cni_mount_dir}/${component}" &>/dev/null && ! test -L "${cni_mount_dir}/${component}"; then
                   echo "${component} binary found, creating symbolic link ${cni_mount_dir}/cnv-${component}"
-                  ln -sf "${cni_mount_dir}/${component}" "${cni_mount_dir}/cnv-${component}"
+                  ln -sf "${host_dir}/${component}" "${cni_mount_dir}/cnv-${component}"
                 else
                   echo "installing cnv-${component}, and creating symbolic link ${cni_mount_dir}/${component}"
 
@@ -54,7 +55,7 @@ spec:
                   printf -v component_checksum "%s" "$(<${sourcebinpath}/${component}.checksum)"
                   printf "%s %s" "${component_checksum% *}" "${cni_mount_dir}/cnv-${component}" | sha256sum --check
 
-                  ln -sf "${cni_mount_dir}/cnv-${component}" "${cni_mount_dir}/${component}"
+                  ln -sf "${host_dir}/cnv-${component}" "${cni_mount_dir}/${component}"
                 fi
               done
               echo 'Entering sleep... (success)'


data/linux-bridge/002-linux-bridge.yaml
```diff
@@ -40,11 +40,12 @@ spec:
             - -ce
             - |
               cni_mount_dir=/opt/cni/bin
+              host_dir={{ .CNIBinDir }}
               components=("bridge" "tuning")
               for component in "${components[@]}"; do
                 if find "${cni_mount_dir}/${component}" &>/dev/null && ! test -L "${cni_mount_dir}/${component}"; then
                   echo "${component} binary found, creating symbolic link ${cni_mount_dir}/cnv-${component}"
-                  ln -sf "${cni_mount_dir}/${component}" "${cni_mount_dir}/cnv-${component}"
+                  ln -sf "${host_dir}/${component}" "${cni_mount_dir}/cnv-${component}"
```
phoracek, line 48: Should not be the source for the link on host_dir as well?

oshoval: the source is the changed one
we can see on OCP that the soft link is created in the right folder (since this folder is mounted)
but the target is wrong as it points to the mount instead the real path
so i think it is good, make sense ?
will double check on CRC

Before fix, `/opt/cni/bin` is the mount point instead the real host path where it should point
```
$ ls -l /var/lib/cni/bin/
total 398436
-rwxr-xr-x. 1 root root  4010197 Jul 20 16:43 bandwidth
-rwxr-xr-x. 1 root root  4497776 Jul 20 16:44 bond
-rwxr-xr-x. 1 root root  4531999 Jul 20 16:43 bridge
lrwxrwxrwx. 1 root root       19 Jul 30 17:24 cnv-bridge -> /opt/cni/bin/bridge   ## binary is missing
lrwxrwxrwx. 1 root root       19 Jul 30 17:24 cnv-tuning -> /opt/cni/bin/tuning   ## binary is missing
```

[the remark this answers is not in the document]
oshoval: as we can see on the `ls` here
https://github.com/kubevirt/cluster-network-addons-operator/pull/1592#discussion_r1281543418
the files that are created are created fine, because it is mounted,
we need to change the pointers only iiuc,

will double check on CRC by tricking it to do both conditions, one at a time

oshoval:
CRC with the fix

```
sh-4.4# ls -l
total 204292
-rwxr-xr-x. 1 root root  3882795 Aug  2 07:59 bandwidth
-rwxr-xr-x. 1 root root  4320904 Aug  2 07:59 bond
-rwxr-xr-x. 1 root root  4336323 Aug  2 07:59 bridge
lrwxrwxrwx. 1 root root       23 Aug  2 08:14 cnv-bridge -> /var/lib/cni/bin/bridge
lrwxrwxrwx. 1 root root       23 Aug  2 08:14 cnv-tuning -> /var/lib/cni/bin/tuning
```

and if i delete all relevant files and redeploy linux bridge (now the `else` is taken)
```
lrwxrwxrwx. 1 root root       27 Aug  2 08:17 bridge -> /var/lib/cni/bin/cnv-bridge
-rwxr-xr-x. 1 root root  4771000 Aug  2 08:17 cnv-bridge
-rwxr-xr-x. 1 root root  3555724 Aug  2 08:17 cnv-tuning
lrwxrwxrwx. 1 root root       27 Aug  2 08:17 tuning -> /var/lib/cni/bin/cnv-tuning
```

oshoval:
/test pull-e2e-cluster-network-addons-operator-workflow-k8s 
/test pull-e2e-cluster-network-addons-operator-lifecycle-k8s 

didnt start for around 5h

PR #1613, opened 2023-09-27
tools: Fix fmt issues

RamLavi:
This PR fixes an fmt issues on the tools folder

To be clear - This folder is not used as part of the production operation of CNAO.

changed are auto created by `make fmt vet`

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tools/metricsdocs/metricsdocs.go
+++ b/tools/metricsdocs/metricsdocs.go
@@ -5,7 +5,7 @@ import (
 	"sort"
 
 	"github.com/kubevirt/cluster-network-addons-operator/pkg/monitoring"
-	"github.com/kubevirt/cluster-network-addons-operator/tools/metrics-parser"
+	metricsparser "github.com/kubevirt/cluster-network-addons-operator/tools/metrics-parser"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


oshoval:
Should we consider adding `make fmt vet` to unit tests ?

tools/metricsdocs/metricsdocs.go
```diff
@@ -5,7 +5,7 @@ import (
 	"sort"
 
 	"github.com/kubevirt/cluster-network-addons-operator/pkg/monitoring"
-	"github.com/kubevirt/cluster-network-addons-operator/tools/metrics-parser"
+	metricsparser "github.com/kubevirt/cluster-network-addons-operator/tools/metrics-parser"
```
oshoval, line 8: should we disable fmt / vet alerts on this ?
because it is nice to have, not mandatory
maybe it suggest it here because the `-` which makes the two names not the same

oshoval (approved):
Thanks

tools/metricsdocs/metricsdocs.go
RamLavi: I don't to keep this level of fmt, fmt adds it automatically anyway, it's not like I need to do anything.
But we can certainly examine further how much we want fmt to alert us in a separate PR.

RamLavi:
> Should we consider adding `make fmt vet` to unit tests ?

It could be a good idea to add `make fmt vet` and then doing a porcelin check to make sure that we don't slip any fmt issues again.. 
But let's experiment that on another PR.

PR #1638, opened 2023-11-01
workflows, auto-bumper: Update supported stable branches

RamLavi:
CNAO will auto bump only the stable branches that are set on the workflow branch matrix.
Updating the supported branches.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/.github/workflows/component-bumper.yaml
+++ b/.github/workflows/component-bumper.yaml
@@ -14,11 +14,10 @@ jobs:
       matrix:
         branch:
           - main
-          - release-0.53
-          - release-0.58
           - release-0.65
           - release-0.76
           - release-0.79
+          - release-0.85
           - release-0.89
     steps:
       - name: Login to Quay


oshoval (approved):

.github/workflows/component-bumper.yaml
```diff
@@ -14,11 +14,10 @@ jobs:
       matrix:
         branch:
           - main
-          - release-0.53
-          - release-0.58
           - release-0.65
```
oshoval, line 17: nit:
maybe worth to add sanity check that RELEASES_SELECTOR lower than oldest doesn't appear here ?
and that all that do appear there appear here as well ?

RamLavi: That's a connection between D/S and U/S that I'm not sure we want to have

oshoval: implicitly yes, but both are on U/S
anyhow whatever you prefer

oshoval:
ovs-mirror

/override pull-e2e-cluster-network-addons-operator-ovs-cni-functests

PR #1653, opened 2023-11-02
Bump kube-rbac-proxy

oshoval:
Includes https://github.com/openshift/kube-rbac-proxy/pull/80
Affects CNAO and KMP.

`KubeRbacProxyImageDefault` was updated manually to a version
that contains the mention fix (we don't have automation, nor need atm for kube-rbac-proxy).

Before this PR we used `baedb268ac66456018fb30af395bb3d69af5fff3252ff5d549f0231b1ebb6901`
everywhere (via variable).

This PR did the following:
1. Changed from using variable to use static hash, because it is wrong to use variable which exists only at 
`pkg/components/components.go` but might be different between versions.
Since the image might change, it can't use the variable which exists only on `pkg/components/components.go`,
but should have an hard coded value per each release.
See https://github.com/kubevirt/cluster-network-addons-operator/pull/1563 for more info.

2. Now that [1] is fixed, manually bump the latest release candid (`test/releases/99.0.0.go`) and `components.go` to the new hash.

Note: since the tests validate the release on the upgrade test,
both for the old and latest, if there were any errors we would see.

We might want to backport it, but each release will need its own version (4.14, 4.13 etc).
Note that this affects only U/S, so not a must imo.

Automation is not needed but note that when a bump happens next time
only those file should be updated with the new hash: 
`pkg/components/components.go`
`test/releases/99.0.0.go`
It is safe because once one is updated, the tests would fail if the 2nd isn't.

**Release note**:

```release-note
None
```

### DIFF
--- a/pkg/components/components.go
+++ b/pkg/components/components.go
@@ -37,7 +37,7 @@ const (
 	KubeMacPoolImageDefault           = "quay.io/kubevirt/kubemacpool@sha256:2f7a4ed0532909176b21eb3b80dec88a14e4f23d0c0aae129acd699636d058a3"
 	OvsCniImageDefault                = "quay.io/kubevirt/ovs-cni-plugin@sha256:e16ac74343da21abb8fb668ce71e728053d00503a992dae2164b9e94a280113e"
 	MacvtapCniImageDefault            = "quay.io/kubevirt/macvtap-cni@sha256:850b89343ace7c7ea6b18dd8e11964613974e9d1f7377af03854d407fb15230a"
-	KubeRbacProxyImageDefault         = "quay.io/openshift/origin-kube-rbac-proxy@sha256:baedb268ac66456018fb30af395bb3d69af5fff3252ff5d549f0231b1ebb6901"
+	KubeRbacProxyImageDefault         = "quay.io/openshift/origin-kube-rbac-proxy@sha256:e2def4213ec0657e72eb790ae8a115511d5b8f164a62d3568d2f1bff189917e8"
 	KubeSecondaryDNSImageDefault      = "ghcr.io/kubevirt/kubesecondarydns@sha256:e87e829380a1e576384145f78ccaa885ba1d5690d5de7d0b73d40cfb804ea24d"
 	CoreDNSImageDefault               = "registry.k8s.io/coredns/coredns@sha256:a0ead06651cf580044aeb0a0feba63591858fb2e43ade8c9dea45a6a89ae7e5e"
 )
--- a/test/releases/0.76.0.go
+++ b/test/releases/0.76.0.go
@@ -2,7 +2,6 @@ package releases
 
 import (
 	cnao "github.com/kubevirt/cluster-network-addons-operator/pkg/apis/networkaddonsoperator/shared"
-	"github.com/kubevirt/cluster-network-addons-operator/pkg/components"
 )
 
 func init() {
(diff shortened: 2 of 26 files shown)


oshoval:
need to fix some stuff

RamLavi:
I don't think it's OK that we're retroactively changing the image on old versions. 
@oshoval can you elaborate why you want to do this?

oshoval:
We don't change it retroactively 
we just replace variable with the hash
see please first paragraph of 
https://github.com/kubevirt/cluster-network-addons-operator/pull/1563

RamLavi:
> We don't change it retroactively we just replace variable with the hash (each value reflects the one in its own test/release/X file) see please first paragraph of #1563

ah I get it..

oshoval:
> > We don't change it retroactively we just replace variable with the hash (each value reflects the one in its own test/release/X file) see please first paragraph of #1563
> 
> ah I get it.. The please add to the PR and commit desc this logic:
> 
> ```
> v0.0.11 and above (PR) --> registry.k8s.io/coredns/coredns@sha256:a0ead06651cf580044aeb0a0feba63591858fb2e43ade8c9dea45a6a89ae7e5e
> v0.0.10  --> registry.k8s.io/coredns/coredns@sha256:5b6ec0d6de9baaf3e92d0f66cd96a25b9edbce8716f5f15dcd1a616b3abd590e
> < v0.0.10  --> k8s.gcr.io/coredns/coredns@sha256:5b6ec0d6de9baaf3e92d0f66cd96a25b9edbce8716f5f15dcd1a616b3abd590e
> ```

It is not the same image

oshoval:
> > We don't change it retroactively we just replace variable with the hash (each value reflects the one in its own test/release/X file) see please first paragraph of #1563
> 
> ah I get it.. The please add to the PR and commit desc this logic:
> 
> ```
> v0.0.11 and above (PR) --> registry.k8s.io/coredns/coredns@sha256:a0ead06651cf580044aeb0a0feba63591858fb2e43ade8c9dea45a6a89ae7e5e
> v0.0.10  --> registry.k8s.io/coredns/coredns@sha256:5b6ec0d6de9baaf3e92d0f66cd96a25b9edbce8716f5f15dcd1a616b3abd590e
> < v0.0.10  --> k8s.gcr.io/coredns/coredns@sha256:5b6ec0d6de9baaf3e92d0f66cd96a25b9edbce8716f5f15dcd1a616b3abd590e
> ```

Took slightly different approach 
more details on PR desc (updated).
Also note that there are just 2 versions in game, so for all the old i changed to the old one
and for the latest i changed to the latest which was added only on this PR.

oshoval:
/cc @RamLavi 
can you please take another look ?

RamLavi (commented):
Can you please update the commit desc?

oshoval:
> Can you please update the commit desc?

It has the gist
```
KubeRbacProxyImageDefault was updated manually.
Each test/release/x.go file was updated according the version it had
in its variable.
Since the tests validate the release on the upgrade test,
both for the old and latest, if there were any errors we would see.
```
The rest isn't as important imho, if you find something specific you think should be added please elaborate

oshoval:
basically the gist was already there
added few more sentences

oshoval:
Added also a note about next time we change the hash in the PR desc

RamLavi:
> /cc @RamLavi can you please take another look ?

From what I see - you chose 2 URLs for different version. 
The commit desc should explain what is the logic behind it. 
This addresses it but I don't feel it's clear enough:
```
Each test/release/x.go file was updated according the version it had
in its variable (basically all stable to the previous hash, latest to the new hash).
```

Could you please elaborate?

oshoval:
Before this PR we used `baedb268ac66456018fb30af395bb3d69af5fff3252ff5d549f0231b1ebb6901`
everywhere

This PR did 2 things
1. Changed to use variable to static hash, because it is wrong to use variable which exists only at 
test/release/99.0.0.go but might be different between versions.
2. Now that [1] is fixed, manually bump on the latest and components.go to the new hash.

oshoval:
Updated PR desc with the above info

oshoval:
rebased

oshoval:
It seems some quay flaky issues, but didn't investigate deeper, and it is not just on this PR
/retest

PR #1708, opened 2023-11-19
CVE-2021-38561: Bump golang.org/x/text

oshoval:
https://github.com/advisories/GHSA-ppp9-7jff-5vj2

Just had to remove the replace that blocked the fixed version.

**Release note**:

```release-note
None
```


RamLavi (commented):
I'm not sure I understand - how did the library get removed?
The way I understand it - 
if it wasn't needed - then go wouldn't put it there (and `make vendor` would bring it back)
so what triggered its removal?

oshoval:
It is needed, it is [there](https://github.com/kubevirt/cluster-network-addons-operator/blob/12f6852a1d6ad6e96260fc61defb5170f053b402/go.mod#L232)
we just had a replace that blocked it, this what was removed.

Those replaces prevent bumping both manually and also by dependabot, hence prone to present CVEs as happened here.

RamLavi:
> We just had a replace that frozen old version of it, this what was removed. It is needed, it is [here](https://github.com/kubevirt/cluster-network-addons-operator/blob/12f6852a1d6ad6e96260fc61defb5170f053b402/go.mod#L232) The masked one already includes the fix (once vendored of course).
> 
> Those replaces prevent bumping both manually and also by dependabot, hence prone to present CVEs as happened here. Of course sometimes we do need them, but they tend to age.

oh I missed the fact the it was removed from replace section.

PR #1728, opened 2023-11-21
[release-0.89] components: Update multus-dynamic-networks to tagged

oshoval:
It will save manual cherry picking.
Note that multus-dynamic-networks have only main branch.

https://github.com/kubevirt/cluster-network-addons-operator/pull/1725#discussion_r1400410829

**Release note**:

```release-note
None
```

### DIFF
--- a/components.yaml
+++ b/components.yaml
@@ -39,7 +39,7 @@ components:
     url: https://github.com/k8snetworkplumbingwg/multus-dynamic-networks-controller
     commit: f40f85bc20f361aeeac7b4b04aa6d289c8e130f5
     branch: main
-    update-policy: static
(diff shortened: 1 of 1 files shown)


oshoval:
Hi @phoracek 
wdyt about this change please ?
if you are fine with it i will rebase, thanks
same q about release-0.85 https://github.com/kubevirt/cluster-network-addons-operator/pull/1729

oshoval:
cc @maiqueb 
As discussed offline
Since there are no stable branches on dynamic networks, we can't use it, because every change on main will be populated 
to stable releases.

maybe worth to consider please branches?

PR #1748, opened 2023-11-27
Avoid to emit duplicates for enum fileds

andreaTP:

The enum "ciphers" contains duplicated entries, since it's Json Schema it should have all unique elements:
https://json-schema.org/understanding-json-schema/reference/enum

The bug has been discovered by @iocanel in https://github.com/fabric8io/kubernetes-client/pull/5611

**Release note**:

```release-note
Fixed incorrect generation of ciphers to prevent duplicated enum values.
```

### DIFF
--- a/pkg/components/components.go
+++ b/pkg/components/components.go
@@ -1036,9 +1036,14 @@ func GetCrd() *extv1.CustomResourceDefinition {
 
 	cipherSuites := func() []extv1.JSON {
 		suites := []extv1.JSON{}
+		m := make(map[string]bool)
 		for _, p := range tlsProfiles(ocpv1.TLSProfiles).sortedKeys() {
 			for _, c := range ocpv1.TLSProfiles[p].Ciphers {
-				suites = append(suites, extv1.JSON{Raw: []byte(fmt.Sprintf("\"%s\"", c))})
+				if m[c] {
+					continue
+				}
+				m[c] = true
+				suites = append(suites, extv1.JSON{Raw: []byte(fmt.Sprintf(`"%s"`, c))})
 			}
 		}
 		return suites
--- a/pkg/components/components_test.go
+++ b/pkg/components/components_test.go
@@ -94,4 +94,19 @@ var _ = Describe("Components", func() {
 			Expect(ris[2].Ref).To(Equal("quay.io/kubevirt/ovs-cni-marker@sha256:0f08d6b1550a90c9f10221f2bb07709d1090e7c675ee1a711981bd429074d620"))
 		})
 	})
+
+	Context("When calculating ciphers", func() {
+		It("should not generate duplicates", func() {
+			var ciphers = GetCrd().Spec.Versions[0].Schema.OpenAPIV3Schema.Properties["spec"].Properties["tlsSecurityProfile"].Properties["custom"].Properties["ciphers"].Items.Schema.Enum
+			var stringCiphers = make([]string, len(ciphers))
+			for i, c := range ciphers {
+				stringCiphers[i] = string(c.Raw[:])
+			}
+			for i, vi := range stringCiphers {
+				for j := i + 1; j < len(stringCiphers); j++ {
+					Expect(vi).ToNot(Equal(stringCiphers[j]))
+				}
+			}
+		})
+	})
 })
--- a/pkg/network/tlsSecurityProfile.go
+++ b/pkg/network/tlsSecurityProfile.go
@@ -11,10 +11,26 @@ func SelectCipherSuitesAndMinTLSVersion(profile *ocpv1.TLSSecurityProfile) ([]st
 			Intermediate: &ocpv1.IntermediateTLSProfile{},
 		}
 	}
+	var ciphers []string
+	var minTlsVersion ocpv1.TLSProtocolVersion
 	if profile.Custom != nil {
-		return profile.Custom.TLSProfileSpec.Ciphers, profile.Custom.TLSProfileSpec.MinTLSVersion
+		ciphers = profile.Custom.TLSProfileSpec.Ciphers
+		minTlsVersion = profile.Custom.TLSProfileSpec.MinTLSVersion
+	} else {
+		ciphers = ocpv1.TLSProfiles[profile.Type].Ciphers
+		minTlsVersion = ocpv1.TLSProfiles[profile.Type].MinTLSVersion
 	}
-	return ocpv1.TLSProfiles[profile.Type].Ciphers, ocpv1.TLSProfiles[profile.Type].MinTLSVersion
+	m := make(map[string]bool)
+	var result []string
+	for _, c := range ciphers {
+		if m[c] {
+			continue
+		}
+		m[c] = true
+		result = append(result, c)
+	}
+
+	return result, minTlsVersion
 }
 
 func TLSVersionToHumanReadable(version ocpv1.TLSProtocolVersion) string {
--- a/pkg/network/tlsSecurityProfile_test.go
+++ b/pkg/network/tlsSecurityProfile_test.go
@@ -63,4 +63,24 @@ var _ = Describe("Testing TLS Security Profile", func() {
 			expectedMinTLSVersion: testCustomTLSProfileSpec.MinTLSVersion,
 		}),
 	)
+
+	Context("When selecting ciphers", func() {
+		It("should not generate duplicates", func() {
+			var profile = &ocpv1.TLSSecurityProfile{
+				Type: ocpv1.TLSProfileCustomType,
+				Custom: &ocpv1.CustomTLSProfile{
+					TLSProfileSpec: ocpv1.TLSProfileSpec{
+						Ciphers: []string{"foo", "foo", "bar"},
+					},
+				},
+				Intermediate: &ocpv1.IntermediateTLSProfile{},
+			}
+			var ciphers, _ = SelectCipherSuitesAndMinTLSVersion(profile)
+			for i, vi := range ciphers {
+				for j := i + 1; j < len(ciphers); j++ {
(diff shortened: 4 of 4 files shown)


pkg/components/components_test.go
```diff
@@ -94,4 +94,25 @@ var _ = Describe("Components", func() {
...
+			var stringCiphers = make([]string, len(ciphers))
+			for i, c := range ciphers {
+				stringCiphers[i] = string(c.Raw[:])
+			}
+
+			for i, vi := range stringCiphers {
+				for j, vj := range stringCiphers {
+					if i == j {
+						continue
+					}
```
oshoval, line 111: We can optimize a lil, wdyt ?
```
for i, vi := range stringCiphers {
    for j := i + 1; j < len(stringCiphers); j++ {
```

oshoval:
We have also this one in the CR
https://github.com/kubevirt/cluster-network-addons-operator/blob/3a2b93b4259772257ead6c28ba748936da28e0bf/pkg/network/kubemacpool.go#L126

maybe worth to also dedup it please ? (and then also to add a test for it `pkg/network/tlsSecurityProfile_test.go`)

andreaTP:
Thanks for the review @oshoval , I'll go through it soon.

pkg/components/components_test.go
andreaTP: ok 👍

andreaTP:
@oshoval should be ready for review again, thanks for the hints 👍

pkg/network/tlsSecurityProfile_test.go
```diff
@@ -63,4 +63,26 @@ var _ = Describe("Testing TLS Security Profile", func() {
...
+				Type: ocpv1.TLSProfileCustomType,
+				Custom: &ocpv1.CustomTLSProfile{
+					TLSProfileSpec: ocpv1.TLSProfileSpec{
+						Ciphers: []string{"foo", "foo", "bar"},
+					},
+				},
+				Intermediate: &ocpv1.IntermediateTLSProfile{},
+			}
+			var ciphers, _ = SelectCipherSuitesAndMinTLSVersion(profile)
+
```
oshoval, line 79: can you please move this line right next to the for ?

pkg/network/tlsSecurityProfile.go
```diff
@@ -11,10 +11,27 @@ func SelectCipherSuitesAndMinTLSVersion(profile *ocpv1.TLSSecurityProfile) ([]st
...
 	if profile.Custom != nil {
-		return profile.Custom.TLSProfileSpec.Ciphers, profile.Custom.TLSProfileSpec.MinTLSVersion
+		ciphers = profile.Custom.TLSProfileSpec.Ciphers
+		minTlsVersion = profile.Custom.TLSProfileSpec.MinTLSVersion
+	} else {
+		ciphers = ocpv1.TLSProfiles[profile.Type].Ciphers
+		minTlsVersion = ocpv1.TLSProfiles[profile.Type].MinTLSVersion
 	}
-	return ocpv1.TLSProfiles[profile.Type].Ciphers, ocpv1.TLSProfiles[profile.Type].MinTLSVersion
+	// deduplicate ciphers
```
oshoval, line 23: no need comments please
they tend to age, and code is self explanatory here

pkg/components/components_test.go
```diff
@@ -94,4 +94,22 @@ var _ = Describe("Components", func() {
...
+
+			var stringCiphers = make([]string, len(ciphers))
+			for i, c := range ciphers {
+				stringCiphers[i] = string(c.Raw[:])
+			}
+
+			for i, vi := range stringCiphers {
+				for j := i + 1; j < len(stringCiphers); j++ {
+					vj := stringCiphers[j]
+					Expect(vi).ToNot(Equal(vj))
```
oshoval, line 110: `Expect(vi).ToNot(Equal(stringCiphers[j]))`
wdyt ?
also on https://github.com/kubevirt/cluster-network-addons-operator/pull/1748/files#diff-7aaa98f5da89ba48b6fd676ea6f71cc8791dfc21c34d61e7ca41631aa9fd3535R82 please if so

oshoval (commented):
Thanks
just little nits

"I hope this is the only place where it should be fixed, but I'm having trouble compiling the entire project on my machine (Mac)." 
can be removed from PR desc

pkg/network/tlsSecurityProfile.go
andreaTP: removed

andreaTP:
I think I have addressed all the notes now, should I squash the commits? any more comments? 🙂

oshoval:
Thanks a lot, sorry for the nit picking
ye squashing sounds great in this case
beside that lgtm :)

andreaTP:
Squashed the commits 👍

oshoval:
Thanks

/lgtm
/retest

@qinqon do you want to take a look ?

oshoval:
those seems to be flakes due to quay or so, anyhow not relevant to this PR
once it is lgtm approve it will be auto restested

andreaTP:
I think that the CI failures are unrelated to the changes in this PR.

oshoval:
You totally right

@qinqon can you please review ?

andreaTP:
Ping, anything should I do on my side to finalize this one?

oshoval:
just need review of @qinqon please
/retest

pkg/components/components.go
```diff
@@ -1036,9 +1036,15 @@ func GetCrd() *extv1.CustomResourceDefinition {
...
+		m := make(map[string]bool)
 		for _, p := range tlsProfiles(ocpv1.TLSProfiles).sortedKeys() {
 			for _, c := range ocpv1.TLSProfiles[p].Ciphers {
-				suites = append(suites, extv1.JSON{Raw: []byte(fmt.Sprintf("\"%s\"", c))})
+				format := fmt.Sprintf("\"%s\"", c)
+				if m[format] {
+					continue
+				}
+				m[format] = true
+				suites = append(suites, extv1.JSON{Raw: []byte(format)})
```
qinqon, line 1047: Do the json "encoding" at the end we don't need to store double quotes at the set

```suggestion
				if m[c] {
					continue
				}
				suites = append(suites, extv1.JSON{Raw: []byte(fmt.Sprintf(`"%s"`, c))})
```

qinqon (changes requested):
Small nit

pkg/network/tlsSecurityProfile_test.go
```diff
@@ -63,4 +63,24 @@ var _ = Describe("Testing TLS Security Profile", func() {
...
+					},
+				},
+				Intermediate: &ocpv1.IntermediateTLSProfile{},
+			}
+			var ciphers, _ = SelectCipherSuitesAndMinTLSVersion(profile)
+			for i, vi := range ciphers {
+				for j := i + 1; j < len(ciphers); j++ {
+					Expect(vi).ToNot(Equal(ciphers[j]))
+				}
+			}
```
qinqon, line 83: We should just compare with the exact slice using gomega matchers, dulication at test is worth it.
```suggestion
			ciphers, _ := SelectCipherSuitesAndMinTLSVersion(profile)
			Expect(ciphers).To(ConsistOf("foo", "bar"))
		
```

pkg/components/components_test.go
```diff
@@ -94,4 +94,19 @@ var _ = Describe("Components", func() {
 			Expect(ris[2].Ref).To(Equal("quay.io/kubevirt/ovs-cni-marker@sha256:0f08d6b1550a90c9f10221f2bb07709d1090e7c675ee1a711981bd429074d620"))
 		})
 	})
+
+	Context("When calculating ciphers", func() {
+		It("should not generate duplicates", func() {
```
qinqon, line 99: Where are the duplicated ciphers ?

qinqon (changes requested):

pkg/network/tlsSecurityProfile_test.go
andreaTP: I differ here, and unrolling loops in tests is not worth it.
In this specific case, the reason is that any time you modify/change/add or remove a cipher, you add a completely redundant task to update the test.

I do believe that this is unnecessary and unneeded maintenance, would you mind expanding on why it would be any better?

pkg/components/components_test.go
andreaTP: Here for example:
https://github.com/kubevirt/cluster-network-addons-operator/blob/3a2b93b4259772257ead6c28ba748936da28e0bf/manifests/cluster-network-addons/0.90.1/network-addons-config.crd.yaml#L1523-L1537

qinqon: What I mean is that we should mock those, we are not really enforcing duplicated on the test.

pkg/network/tlsSecurityProfile_test.go
qinqon: ok, fair enough.

pkg/components/components_test.go
andreaTP: The goal of this test(and this PR overall) is to ensure we don't emit duplicates in the "ciphers" enums, the fact that they do exist or not in the original slice is almost an implementation detail.

If you want to do mock testing on something this small, please go ahead, you can PR my branch, fix on top or do whatever you prefer.

qinqon: Ok

qinqon (approved):
/lgtm
/approve

qinqon:
/hold

@andreaTP please squash the commits.

andreaTP:
@qinqon squashed the commits, rebased on latest `main` ...

oshoval:
Thanks

PR #1749, opened 2023-11-28
[release-0.85] Release v0.85.5 

oshoval:
Based on https://github.com/kubevirt/cluster-network-addons-operator/pull/1747

Remove unsupported releases until 0.69.0 (*)
0.69.0 is also flaky and not supported anymore.
This will also cause lifecycle be shorter than 4h.

(*) Do not remove the ones that `RELEASES_SELECTOR` choose.
There is a weird problem with `RELEASES_SELECTOR`,
it pointed to a version that didn't exists, and still didn't
fail (because it is used as an optional filter of the existing
releases), for now just amended it.
We can consider fixing it on a follow-up [effort](https://github.com/kubevirt/cluster-network-addons-operator/pull/1751).

**Release note**:

```release-note
None
```


oshoval:
it passed, basically we can delete all the releases until it maybe on this bracnh

oshoval:
Those seems to come and go during deployment, maybe quay is slower, or CI machines net is slower?
https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_cluster-network-addons-operator/1749/pull-e2e-cnao-br-marker-functests-release-0.85/1729757510232969216

/retest

oshoval:
/retest

not sure why it doesnt manage to deploy sometimes, we need to add post mortem in case it doesnt go away

PR #1776, opened 2024-03-28
Add operator-observability rules unit test

assafad:
Add [operator-observability rules unit test](https://github.com/machadovilaca/operator-observability/tree/main/pkg/testutil), which ensures that alerts and recording rules definitions are following the monitoring best practices, e.g. alerts have all the required annotations and labels with expected values, recording rules have expressions, etc.

Fixes https://issues.redhat.com/browse/CNV-36763
Other PRs references: https://github.com/kubevirt/hyperconverged-cluster-operator/pull/2827, https://github.com/kubevirt/kubevirt/pull/11454

/cc @machadovilaca, @oshoval

**Release note**:

```release-note
None
```

### DIFF
--- a/pkg/monitoring/rules/rules_suite_test.go
+++ b/pkg/monitoring/rules/rules_suite_test.go
@@ -1,0 +1,13 @@
+package rules_test
+
+import (
+	"testing"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+func TestRules(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Rules Suite")
+}
--- a/pkg/monitoring/rules/rules_test.go
+++ b/pkg/monitoring/rules/rules_test.go
@@ -1,0 +1,37 @@
+package rules_test
+
+import (
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	"github.com/machadovilaca/operator-observability/pkg/testutil"
+
+	"github.com/kubevirt/cluster-network-addons-operator/pkg/monitoring/rules"
+)
+
+var _ = Describe("Rules Validation", func() {
+	var linter *testutil.Linter
+
+	BeforeEach(func() {
+		Expect(rules.SetupRules("")).To(Succeed())
+		linter = testutil.New()
+	})
+
+	It("Should validate alerts", func() {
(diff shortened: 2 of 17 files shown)


pkg/monitoring/rules/rules_test.go
```diff
@@ -0,0 +1,42 @@
...
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	"github.com/machadovilaca/operator-observability/pkg/testutil"
+)
+
+func TestRules(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Rules Suite")
+}
```
machadovilaca, line 15: usually, the convention is to create a `..._suite_test.go` file with this code and a `...test.go` for the actual test code, such as in:

https://github.com/kubevirt/cluster-network-addons-operator/blob/main/pkg/controller/networkaddonsconfig/networkaddonsconfig_controller_suite_test.go

https://github.com/kubevirt/cluster-network-addons-operator/blob/main/pkg/controller/networkaddonsconfig/networkaddonsconfig_controller_test.go

machadovilaca (changes requested):

pkg/monitoring/rules/rules_test.go
assafad: Thanks, done.

pkg/monitoring/rules/rules_suite_test.go
```diff
@@ -0,0 +1,13 @@
+package rules
```
oshoval, line 1: please use package rules_test (both files)
all the functions that you are testing are already external so it is easy

oshoval (commented):
Thanks

Please consider splitting to commits, one that does the vendoring and one for the code changes, here it is easy to review, but when it isn't it is much better.

Do we need all the PR referenced ? it is nice to have, but maybe lets trim the list please to the most important ones?

pkg/monitoring/rules/rules_suite_test.go
assafad: I had to import rules package for using some functions, but done.

assafad:
> Thanks
> 
> Please consider splitting to commits, one that does the vendoring and one for the code changes, here it is easy to review, but when it isn't it is much better.
> 
> Do we need all the PR referenced ? it is nice to have, but maybe lets trim the list please to the most important ones?

@oshoval Done.

PR #1798, opened 2024-04-17
[release 0.79] test/e2e/lifecycle: Update to only supported versions 

RamLavi:
CNAO runs lifecycle e2e test over all supported versions.
Updating versions so that the CI will test more relevant releases.

**Release note**:

```release-note
NONE
```


oshoval (approved):
thanks
i wonder if we can implement a more automatic sliding window on a follow-up effort
that will know to ditch versions when they get too old and not supported (not sure it can be fully automatic, because sometimes we support more than 3 minor versions for a period)

RamLavi:
> thanks i wonder if we can implement a more automatic sliding window on a follow-up effort that will know to ditch versions when they get too old and not supported (not sure it can be fully automatic, because sometimes we support more than 3 minor versions for a period)

+1

/approve

Issue #1813, opened 2024-06-24
Auto bump fails because of multus component


oshoval:
**What happened**:
https://github.com/kubevirt/cluster-network-addons-operator/actions/runs/9642633059

It fails because it can't find the commit associated to the tag.
The reason is we are fetching 100 commits, and atm the required commit is far from HEAD
by 109 commits.
`git rev-list --count f03765681fe81ee1e0633ee1734bf48ab3bccf2b..HEAD`
We should use pagination, because max per page is 100 (`getLatestTaggedFromBranch`)
https://docs.github.com/en/rest/commits/commits?apiVersion=2022-11-28#list-commits

**What you expected to happen**:
Bump should work even if the commit is far away.
Alternatively we can do rev-parse / rev-list by the tag to its hash maybe ?

For example this works from CLI, we can check the code version of it
```
git rev-list -n1 v4.0.2
f03765681fe81ee1e0633ee1734bf48ab3bccf2b
```
instead of (this code otherwise need pagination)
```
		for _, commit := range branchCommits {
			if commit.GetSHA() == commitShaOfTag {
				return tagName, commit.GetSHA(), nil
			}
		}
```

**How to reproduce it (as minimally and precisely as possible)**:
Just check the bump and see multus fails.

**Additional context**:

**Environment**:
- KubeVirt version (use `virtctl version`): N/A
- Kubernetes version (use `kubectl version`): N/A
- VM or VMI specifications: N/A
- Cloud provider or hardware configuration: N/A
- OS (e.g. from /etc/os-release): N/A
- Kernel (e.g. `uname -a`): N/A
- Install tools: N/A
- Others: N/A

oshoval:
Hi @RamLavi 
Any reason why not just use rev-parse / rev-list
given the tag, fetch its commit and compare it to the expected one?
Otherwise we would need pagination and iterate over all the commits, trying to find one matching the expected commit
as we do now but with additional pagination.
The logic seems redundant.

RamLavi:
Hey @oshoval 
Are you offering changing the library function with an "exec" equivalent?

oshoval:
Hi
I suggest a `rev-list` equivalent library call if exists
about exec we can consider at worse case, or to do the pagination

Scope of code change is just this code:
```
		for _, commit := range branchCommits {
			if commit.GetSHA() == commitShaOfTag {
				return tagName, commit.GetSHA(), nil
			}
		}
```

[...]

PR #1815, opened 2024-06-26
bumper: Paginate over commits on tagged update policy

RamLavi:
This PR expands the commits range used to find the latest tag when bumping using "tagged" update policy

Fixes https://github.com/kubevirt/cluster-network-addons-operator/issues/1813

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tools/bumper/component_commands.go
+++ b/tools/bumper/component_commands.go
@@ -202,11 +202,10 @@ func (componentOps *gitComponent) getLatestTaggedFromBranch(repo, owner, branch,
 		return "", "", errors.Wrap(err, "Failed to get release tag refs from github client API")
 	}
 
-	branchCommits, _, err := componentOps.githubInterface.listCommits(owner, repo, &github.CommitsListOptions{ListOptions: github.ListOptions{PerPage: 100}, SHA: branch})
-	if err != nil {
-		return "", "", errors.Wrap(err, "Failed to get release tag refs from github client API")
-	}
-
+	const (
+		maxPagePaginate = 10
+		pageSize        = 100
+	)
 	// look for the first tag that belongs to the chosen branch, going over from last tag
 	for i := len(tagRefs) - 1; i >= 0; i-- {
 		tag := tagRefs[i]
@@ -216,14 +215,24 @@ func (componentOps *gitComponent) getLatestTaggedFromBranch(repo, owner, branch,
 			return "", "", errors.Wrap(err, "Failed to get commit-sha from tag name")
 		}
 
-		for _, commit := range branchCommits {
-			if commit.GetSHA() == commitShaOfTag {
-				return tagName, commit.GetSHA(), nil
+		for pageIdx := 1; pageIdx <= maxPagePaginate; pageIdx++ {
+			branchCommits, resp, err := componentOps.githubInterface.listCommits(owner, repo, &github.CommitsListOptions{ListOptions: github.ListOptions{PerPage: pageSize, Page: pageIdx}, SHA: branch})
+			if err != nil {
+				return "", "", errors.Wrap(err, "Failed to get release tag refs from github client API")
+			}
+			if resp.NextPage == 0 {
+				break
+			}
+
+			for _, commit := range branchCommits {
+				if commit.GetSHA() == commitShaOfTag {
+					return tagName, commit.GetSHA(), nil
+				}
 			}
 		}
 	}
 
-	return "", "", fmt.Errorf("Error: tag not found in branch %s\n tagRefs=\n%v\nbranchCommits=\n%v\n", branch, tagRefs, branchCommits)
+	return "", "", fmt.Errorf("Error: tag not found in branch %s on last %d commits\n", branch, maxPagePaginate*pageSize)
 }
 
 // getLatestFromBranch get the latest HEAD commit-sha under a given branch, using "latest" Update policy
--- a/tools/bumper/git_fakes.go
+++ b/tools/bumper/git_fakes.go
@@ -42,7 +42,7 @@ func (g mockGithubApi) listCommits(owner, repo string, opts *github.CommitsListO
 		return nil, nil, errors.Wrap(err, "failed performing mock git log")
 	}
 
-	return convertLogToRepositoryCommitList(gitCommitObjList), nil, nil
+	return convertLogToRepositoryCommitList(gitCommitObjList), &github.Response{NextPage: 1}, nil
 }
 
 func (g mockGithubApi) getBranchRef(owner string, repo string, ref string) (*github.Reference, *github.Response, error) {


RamLavi:
@oshoval PTAL

tools/bumper/component_commands.go
```diff
@@ -216,14 +215,21 @@ func (componentOps *gitComponent) getLatestTaggedFromBranch(repo, owner, branch,
 			return "", "", errors.Wrap(err, "Failed to get commit-sha from tag name")
 		}
 
-		for _, commit := range branchCommits {
-			if commit.GetSHA() == commitShaOfTag {
-				return tagName, commit.GetSHA(), nil
+		for pageIdx := 0; pageIdx < maxPagePaginate; pageIdx++ {
```
oshoval, line 218: i think first page is 1 maybe according 
https://docs.github.com/en/rest/commits/commits?apiVersion=2022-11-28#list-commits

tools/bumper/component_commands.go
```diff
@@ -202,11 +202,10 @@ func (componentOps *gitComponent) getLatestTaggedFromBranch(repo, owner, branch,
 		return "", "", errors.Wrap(err, "Failed to get release tag refs from github client API")
 	}
 
-	branchCommits, _, err := componentOps.githubInterface.listCommits(owner, repo, &github.CommitsListOptions{ListOptions: github.ListOptions{PerPage: 100}, SHA: branch})
-	if err != nil {
-		return "", "", errors.Wrap(err, "Failed to get release tag refs from github client API")
-	}
-
+	const (
+		maxPagePaginate = 10
```
oshoval, line 206: if we use max, the bug can return if there are 1000 commits between
well i guess we can fix it once it happens if it does (rare)

tools/bumper/component_commands.go
oshoval, line 218: the fact it is in a loop means we reget it for each tag
we can optimize it, question is if we want to

one way is that if it didnt find in the first page, fetch the pages but keep them in memory

oshoval:
test is broken also on main branch, retest wont help
just override it please / we need to fix it

tools/bumper/component_commands.go
oshoval: if you want, you can start without saying the page first time (even have it outside the loop as it was)
and only upon needs have it in the loop (not overriding the outer one)
and use there next_page field to determine the next page

oshoval:
```
tag, _, err := componentOps.githubInterface.GetTag(owner, repo, tagName)
if err != nil {
    return "", errors.Wrap(err, "Failed to get tag from GitHub client API")
}

commitHash := tag.Object.SHA
```

what about something like this ? didnt try 
https://pkg.go.dev/github.com/google/go-GitHub/github

tools/bumper/component_commands.go
RamLavi: yes. We can't look forever, we can only look in the logical range..

tools/bumper/component_commands.go
RamLavi: you're right. Fixed

RamLavi: no, we stop as soon as we find a match. 
Most times the first tag searched is the one that will hit.

RamLavi:
> ```
> tag, _, err := componentOps.githubInterface.GetTag(owner, repo, tagName)
> if err != nil {
>     return "", errors.Wrap(err, "Failed to get tag from GitHub client API")
> }
> 
> commitHash := tag.Object.SHA
> ```
> 
> what about something like this ? didnt try https://pkg.go.dev/github.com/google/go-GitHub/github

IIRC this way only finds annotated tags..

tools/bumper/component_commands.go
oshoval: i know, but if it isnt found for the first ttag we refetch them every loop this what i mean
anyhow we can live with that

tools/bumper/component_commands.go
```diff
@@ -216,14 +215,21 @@ func (componentOps *gitComponent) getLatestTaggedFromBranch(repo, owner, branch,
 			return "", "", errors.Wrap(err, "Failed to get commit-sha from tag name")
 		}
 
-		for _, commit := range branchCommits {
-			if commit.GetSHA() == commitShaOfTag {
-				return tagName, commit.GetSHA(), nil
+		for pageIdx := 1; pageIdx <= maxPagePaginate; pageIdx++ {
+			branchCommits, _, err := componentOps.githubInterface.listCommits(owner, repo, &github.CommitsListOptions{ListOptions: github.ListOptions{PerPage: pageSize, Page: pageIdx}, SHA: branch})
```
oshoval, line 219: if there are only 3 pages, and you try to fetch page 4 it will fail
please make sure the page exists, there is a next_page field for that

RamLavi: I can't find this field. Can you direct me to it?

oshoval:
> > ```
> > tag, _, err := componentOps.githubInterface.GetTag(owner, repo, tagName)
> > if err != nil {
> >     return "", errors.Wrap(err, "Failed to get tag from GitHub client API")
> > }
> > 
> > commitHash := tag.Object.SHA
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > what about something like this ? didnt try https://pkg.go.dev/github.com/google/go-GitHub/github
> 
> IIRC this way only finds annotated tags, but feel free to experiment and send a PR to improve this method

imho worth to check, it will even make you not need to mess with the next page thing
up to you

tools/bumper/component_commands.go
oshoval: I think via the link header
"The URL for the last page is followed by rel="last"."
https://docs.github.com/en/rest/using-the-rest-api/using-pagination-in-the-rest-api?apiVersion=2022-11-28

you can also check if err is page not exists / overflow, but it is less nice of course

oshoval:
```
		if resp.NextPage == 0 {
			break
		}
```

resp is the 2nd return value of listCommits

RamLavi: I checked, it works

oshoval:
Thanks
/lgtm
/approve

oshoval:
/lgtm cancel

panic here 
https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_cluster-network-addons-operator/1815/pull-cluster-network-addons-operator-unit-test/1806269905582952448
if you can take a look please

oshoval:

can we just drop maybe nmstate tests if the oldest version we support doesnt deploy cnao anymore?
else we need to fix it in a different manner

https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_cluster-network-addons-operator/1815/pull-e2e-cluster-network-addons-operator-workflow-k8s/1806289965693276160

oshoval:
I think i know what is the bug, will post fix

RamLavi:
> can we just drop maybe nmstate tests if the oldest version we support doesnt deploy cnao anymore?
> else we need to fix it in a different manner

moved the conversation to slack as it doesn't really relate to this specific PR

PR #1817, opened 2024-06-27
auto bumper: Fix in case there is only one page

oshoval:
We first need to consume the page, then to break if this is the last page, else we would directly fail.

**Release note**:

```release-note

```


oshoval:
@RamLavi ptal

oshoval:

/lgtm
/approve

RamLavi:
+1 
good catch!

PR #1816, opened 2024-06-27
ipam: Adapt to ipam-controller renaming and repo transfer

oshoval:
**What this PR does / why we need it**:
Since ipam component was renamed to `kubevirt-ipam-controller`,
and also the repo URL was changed (to https://github.com/kubevirt/ipam-extensions)
we need to adapt to those changes.

**Special notes for your reviewer**:

**Release note**:

```release-note

```

PR #1819, opened 2024-07-02
KIC: Cleanup of old named resources

oshoval:
Since Kubevirt ipam controller resource names were renamed [1]
since 0.94.0, we introduce the special remove that in case the old named
objects exist, will remove them.
This allows the lifecycle lane to pass, as it checks for leftovers,
which otherwise exists.

[1] https://github.com/kubevirt/cluster-network-addons-operator/pull/1816

**Release note**:

```release-note
None
```

### DIFF
--- a/pkg/network/kubevirt_ipam_controller.go
+++ b/pkg/network/kubevirt_ipam_controller.go
@@ -1,11 +1,18 @@
 package network
 
 import (
+	"context"
+	"fmt"
+	"log"
 	"os"
 	"path/filepath"
 
 	"github.com/pkg/errors"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
+	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
 
 	cnao "github.com/kubevirt/cluster-network-addons-operator/pkg/apis/networkaddonsoperator/shared"
 	"github.com/kubevirt/cluster-network-addons-operator/pkg/render"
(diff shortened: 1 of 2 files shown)


oshoval:
https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_cluster-network-addons-operator/1819/pull-cluster-network-addons-operator-unit-test/1808027843397947392

seems it fails due to something like this
https://github.com/operator-framework/operator-registry/issues/178#issuecomment-759090308

so we would need to use deselector

oshoval:
/cc @RamLavi 
ptal

RamLavi:
Not sure what you mean by "Delete release 0.94.0"
It's a version, U/S user may already consume it. 
Can we just remove support from it? it breaks the entire lifecycle

oshoval:
it is very new and not released beside on upstream main 
i think it is safe to delete it, we did that in the past

if we don't delete it lifecycle fails https://github.com/kubevirt/cluster-network-addons-operator/pull/1818#issuecomment-2200457389

since the kubevirt ipam controller component name was changed it leaves leftovers (details in the link above)
the alternative is to add it to deselector

RamLavi:
> it is very new and not released beside on upstream main i think it is safe to delete it, we did that in the past
> 
> if we don't delete it lifecycle fails [#1818 (comment)](https://github.com/kubevirt/cluster-network-addons-operator/pull/1818#issuecomment-2200457389)
> 
> since the kubevirt ipam controller component name was changed it leaves leftovers (details in the link above) the alternative is to add it to deselector

It is not permitted to remove a supported release. Once it is tagged - you can't go back.
From what I see - CNAO API wasn't changed - am I correct?
Regarding the now "old" manifests - CNAO needs to handle them, IIRC we have a special [place](https://github.com/kubevirt/cluster-network-addons-operator/blob/568025746fb75ef990e34f70012e1781b8c30e7b/pkg/network/network.go#L75) for that. I suggest we go towards that approach
/hold

oshoval:
> > it is very new and not released beside on upstream main i think it is safe to delete it, we did that in the past
> > if we don't delete it lifecycle fails [#1818 (comment)](https://github.com/kubevirt/cluster-network-addons-operator/pull/1818#issuecomment-2200457389)
> > since the kubevirt ipam controller component name was changed it leaves leftovers (details in the link above) the alternative is to add it to deselector
> 
> It is not permitted to remove a supported release. Once it is tagged - you can't go back. From what I see - CNAO API wasn't changed - am I correct? Regarding the now "old" manifests - CNAO needs to handle them, IIRC we have a special [place](https://github.com/kubevirt/cluster-network-addons-operator/blob/568025746fb75ef990e34f70012e1781b8c30e7b/pkg/network/network.go#L75) for that. I suggest we go towards that approach /hold

This is a release on main branch which is unstable
Having a function like `cleanUpMultusOldName` will need to make sure it need to delete it only if the current version is >= 0.94.1 and it will need to delete a list of resources (as they all were changed)
I think it is a total overkill
lets at least do the deselect and say remove support in release notes?
if any user uses 0.94.0 we can just request to clean and reinstall one time
again it is main branch, unstable) and the probability someone already uses it and will have problems is rare
(at most there is just release notes)

RamLavi:
> lets at least do the deselect and say remove support in release notes?

I disagree. Every release should be supported lifecycle-wize. main branch is theoretically unstable but it doesn't mean we get to break the lifecycle. This is precisely the reason the lane is there for.
I am not convinced that this case justify breaking the lane.

oshoval:
> > lets at least do the deselect and say remove support in release notes?
> 
> I disagree. Every release should be supported lifecycle-wize. main branch is theoretically unstable but it doesn't mean we get to break the lifecycle. This is precisely the reason the lane is there for. I am not convinced that this case justify breaking the lane.

we even didn't even finish delivering the feature, the release was created only so HCO can consume
and we renamed the repo as soon as possible, which caused this
adding this removal function in so early stage consumes focus and reduce simplification imho
being robust here is the right thing imho

oshoval:
Done, lets see it pass (didn't try locally)

oshoval:
https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_cluster-network-addons-operator/1819/pull-e2e-cluster-network-addons-operator-lifecycle-k8s/1808100330181758976
passed with the special clean-up changes

will arrange the commit

/hold cancel

oshoval:
@RamLavi ptal

oshoval:
split the commits, this PR will be just the cleanup 
and then we would able to take https://github.com/kubevirt/cluster-network-addons-operator/pull/1818

oshoval:
workflow fix
https://github.com/kubevirt/cluster-network-addons-operator/pull/1820

RamLavi:
> @RamLavi ptal
> 
> if needed i can split this PR to 2 different PRs one is the fix and then comes the release i think it should work as well

why do we need to split? this commit has no place but after a new release was created. Am I missing something?

oshoval:
> > @RamLavi ptal
> > if needed i can split this PR to 2 different PRs one is the fix and then comes the release i think it should work as well
> 
> why do we need to split? this commit has no place but after a new release was created. Am I missing something?

the other way around
the renaming was already merged, so this is the cleanup
and then a new release can be introduced
it is better that new releases are PRs which are automated once possible
the PR can be standalone, it is not that they depend and need to be bundled
they just both need to be merged
the fact the logic is there doesnt check what release are we on, because the renaming is already included
anyhow it is really nits imho

RamLavi:
> > > @RamLavi ptal
> > > if needed i can split this PR to 2 different PRs one is the fix and then comes the release i think it should work as well
> > 
> > 
> > why do we need to split? this commit has no place but after a new release was created. Am I missing something?
> 
> the other way around the renaming was already merged, so this is the cleanup and then a new release can be introduced it is better that new releases are PRs which are automated once possible the PR can be standalone, it is not that they depend and need to be bundled they just both need to be merged the fact the logic is there doesnt check what release are we on, because the renaming is already included anyhow it is really nits imho

oh I see then.

oshoval:

pkg/network/multus.go
```diff
@@ -79,6 +79,95 @@ func cleanUpMultusOldName(ctx context.Context, client k8sclient.Client) []error
 	return []error{}
 }
 
+// cleanUpKubevirtIpamController checks specific kic outdated objects or ones that are no longer compatible and deletes them.
+func cleanUpKubevirtIpamController(conf *cnao.NetworkAddonsConfigSpec, ctx context.Context, client k8sclient.Client) []error {
```
RamLavi, line 83: could you move this code to `pkg/network/kubevirt_ipam_controller.go`?

RamLavi (changes requested):
code looks good! just needs to move to the right location

pkg/network/multus.go
oshoval: oops done

lets see it pass

RamLavi (commented):
missed something: please add the PR that renamed the objects on the commit desc.

oshoval:
done

RamLavi (approved):
/lgtm
/approve

oshoval:
btw all those cleanups (not just this one)
becomes outdated at some stage and can be removed
it also cause API calls, maybe we can consider a boolean that once the flow saw that nothing should be removed (but conf.ipam == true), or all that was to removed, removed successfully, short circuit the function
unless those API calls arent expansive

alternative is to not scan for this type of leftovers (but i guess it is less nice solution)

RamLavi:
> btw all those cleanups (not just this one) becomes outdated at some stage and can be removed it also cause API calls, maybe we can consider a boolean that once the flow saw that nothing should be removed (but conf.ipam == true), or all that was to removed, removed successfully, short circuit the function unless those API calls arent expansive
> 
> alternative is to not scan for this type of leftovers (but i guess it is less nice solution)

The same logic needs to apply like the nmstate tests - if it should no longer be supported then we can remove the special clean.

PR #1828, opened 2024-07-21
Fix runbook_url test

assafad:
What this PR does / why we need it:
Fix runbook_url test to not expect the alert name as the URL suffix, as in some installations the alert name is not the URL suffix.
e.g. in openshift the runbooks are in https://github.com/openshift/runbooks/tree/master/alerts/openshift-virtualization-operator, and the suffix is `<alert-name>.md`.

**Release note**:

```release-note
None
```


RamLavi (approved):
approve

assafad:
@RamLavi can you please `/approve`?

PR #1829, opened 2024-07-21
kubevirt IPAM controller: Deploy Passt CNI binary

oshoval:
As part of kubevirt ipam controller deploy, deploy also
the Passt binding CNI binary to the nodes, via a DaemonSet.

**Release note**:

```release-note
Support deploying Passt binding CNI
```


data/kubevirt-ipam-controller/003-passtbindingcni.yaml
oshoval, line 1: @qinqon 
can we keep this here as is ?
it will reduce the need to grab the DaemonSet from the repo and patch it with the required templation
(but on the other hand every change should be done on both repos)
usually those doesnt change for ages, so it is easier to keep as is

wdyt?
maybe we can consider keeping as is now and later if needed change?
whatever you prefer anyhow

oshoval:
@qinqon 
Can you please review?
Thanks

qinqon (approved):
/lgtm
/approve

oshoval:
Thanks
one test failed, due to flakiness not related to this PR specifically, but a problem i saw every now and then
pushed a commit to address it

PR #1832, opened 2024-07-25
manifest templator: Fix var names

oshoval:
Command line variables of manifest generator should have `_image` suffix for the image related ones.
Also `passtBindingCNIImage` variable was missing "Image" suffix.

In order to be backward compatible, lets first add those variables, use them, and then
we would able to drop the older ones, and the related code that was added for this reason.
The logic is if the old var is set, use that, else use the newer one.

**Release note**:

```release-note
None
```


qinqon (approved):
/lgtm
/approve

oshoval:
Once we use the new names we can remove this code
https://github.com/kubevirt/cluster-network-addons-operator/pull/1834

PR #1842, opened 2024-08-05
hack/deploy-kubevirt: Align wait kv ready to 15m

RamLavi:
kubevirt/kubevirt waits for 15m [0], there is no reason we should wait for less.

[0] https://github.com/kubevirt/kubevirt/blob/e18e89be39b705249ca87afae6e0369ce25dbc25/hack/ci/entrypoint.sh#L90

**Release note**:

```release-note
NONE
```

### DIFF
--- a/hack/deploy-kubevirt.sh
+++ b/hack/deploy-kubevirt.sh
@@ -26,6 +26,6 @@ until ./cluster/kubectl.sh -n kubevirt get kv kubevirt; do
     sleep 1
 done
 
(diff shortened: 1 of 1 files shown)


oshoval (approved):
we can try this, not sure yet if it is the real reason, or the watch files (as it fails only on git actions, not on prow)
on the other hand there is no cache
worth to try to check how much OVN-K waits (EDIT it waits 15m so good to change indeed)

oshoval:
hmm
`error: timed out waiting for the condition on kubevirts/kubevirt`
/lgtm cancel

oshoval:
attempt here https://github.com/kubevirt/cluster-network-addons-operator/pull/1843

RamLavi:
seem to not work, let's close this change for now.

PR #1869, opened 2024-08-18
lifecycle: Update RELEASES_SELECTOR

oshoval:
**What this PR does / why we need it**:
Remove old versions (keep 3 versions + release candid).
Update outdated version to use latest release of each minor.
Add missing releases.

**Special notes for your reviewer**:
Created this helper script, it will make this process less prone for mistakes
https://github.com/kubevirt/cluster-network-addons-operator/pull/1872

**Release note**:

```release-note
None
```

PR #1872, opened 2024-08-18
lifecycle: Add helper script to fetch missing manifests

oshoval:
Also fetches required missing test/releases files.
Helps in creating PRs such as https://github.com/kubevirt/cluster-network-addons-operator/pull/1869

Can be improved to read `RELEASES_SELECTOR` directly from
`automation/check-patch.e2e-lifecycle-k8s.sh`.
Can also be improved to purge non supported versions according
lowest `RELEASES_SELECTOR` element.

Script assumes that if the manifests exists, then the test/releases file also exists
(can be improved later according needs).
Script assumes last element is `99.0.0` so it skips it (it is prerequisite).

We can't make it part of the automation (and not even sanity check due to the following reasons)
because sometimes small changes are needed
for the test/releases files [1], unless those were also fixed on the branch itself.
Anyhow we can consider that later if desired.
Helper scripts are good enough for now.

[1] For example https://github.com/kubevirt/cluster-network-addons-operator/compare/6d41e113ea20005c25b45bd896499f31b4e15711..23c5c735e74e854ed5434b0ccf59e3ddf1b2074a
There was a bug, that was fixed on main but not on branch, as it affected only main iirc.

**Release note**:

```release-note
None
```

### DIFF
--- a/hack/update_releases.sh
+++ b/hack/update_releases.sh
@@ -1,0 +1,34 @@
+#!/bin/bash
+
+# This script pulls release manifests needed for lifecycle tests.
+#
+# Usage example:
+# export RELEASES_SELECTOR="{0.89.3,0.91.1,0.93.0,0.95.0,99.0.0}"
+# ./hack/update_releases.sh
+
+set -ue
+RELEASES=${RELEASES_SELECTOR//[\{\}]}
+
+IFS=',' read -ra RELEASES_ARRAY <<< "$RELEASES"
+
+for ((i = 0; i < ${#RELEASES_ARRAY[@]} - 1; i++)); do
+    VERSION=${RELEASES_ARRAY[i]}
+    if [ -d "manifests/cluster-network-addons/${VERSION}" ]; then
+      continue
+    fi
+    echo "Processing release: ${VERSION}"
+
+    mkdir -p manifests/cluster-network-addons/${VERSION}
+    pushd manifests/cluster-network-addons/${VERSION} > /dev/null
+    curl -sLO https://github.com/kubevirt/cluster-network-addons-operator/releases/download/v${VERSION}/cluster-network-addons-operator.${VERSION}.clusterserviceversion.yaml
+    curl -sLO https://github.com/kubevirt/cluster-network-addons-operator/releases/download/v${VERSION}/namespace.yaml
+    curl -sLO https://github.com/kubevirt/cluster-network-addons-operator/releases/download/v${VERSION}/network-addons-config-example.cr.yaml
+    curl -sLO https://github.com/kubevirt/cluster-network-addons-operator/releases/download/v${VERSION}/network-addons-config.crd.yaml
+    curl -sLO https://github.com/kubevirt/cluster-network-addons-operator/releases/download/v${VERSION}/operator.yaml
+    popd > /dev/null
+
+    pushd test/releases > /dev/null
+    curl -sLO https://raw.githubusercontent.com/kubevirt/cluster-network-addons-operator/release-${VERSION%.*}/test/releases/${VERSION}.go
+    popd > /dev/null
+done
+


hack/update_releases.sh
```diff
@@ -0,0 +1,32 @@
...
+IFS=',' read -ra RELEASES_ARRAY <<< "$RELEASES"
+
+for ((i = 0; i < ${#RELEASES_ARRAY[@]} - 1; i++)); do
+    VERSION=${RELEASES_ARRAY[i]}
+    if [ -d "manifests/cluster-network-addons/${VERSION}" ]; then
+      continue
+    fi
+    echo "Processing release: ${VERSION}"
+
+    mkdir -p manifests/cluster-network-addons/${VERSION}
```
RamLavi, line 19: Please consider Double quote to prevent globbing and word splitting ([SC2086](https://www.shellcheck.net/wiki/SC2086))

```suggestion
    mkdir -p manifests/cluster-network-addons/"${VERSION}"
```

hack/update_releases.sh
```diff
@@ -0,0 +1,32 @@
...
+    pushd manifests/cluster-network-addons/${VERSION} > /dev/null
```
RamLavi, line 20: ditto
```suggestion
    pushd manifests/cluster-network-addons/"${VERSION}" > /dev/null
```

hack/update_releases.sh
```diff
@@ -0,0 +1,32 @@
...
+    curl -sLO https://github.com/kubevirt/cluster-network-addons-operator/releases/download/v${VERSION}/cluster-network-addons-operator.${VERSION}.clusterserviceversion.yaml
```
RamLavi, line 21: ditto 

```suggestion
    curl -sLO https://github.com/kubevirt/cluster-network-addons-operator/releases/download/v"${VERSION}"/cluster-network-addons-operator."${VERSION}".clusterserviceversion.yaml
```

hack/update_releases.sh
```diff
@@ -0,0 +1,32 @@
...
+    curl -sLO https://github.com/kubevirt/cluster-network-addons-operator/releases/download/v${VERSION}/namespace.yaml
```
RamLavi, line 22: ditto
```suggestion
    curl -sLO https://github.com/kubevirt/cluster-network-addons-operator/releases/download/v"${VERSION}"/namespace.yaml
```

hack/update_releases.sh
```diff
@@ -0,0 +1,32 @@
...
+    curl -sLO https://github.com/kubevirt/cluster-network-addons-operator/releases/download/v${VERSION}/network-addons-config-example.cr.yaml
```
RamLavi, line 23: ditto
```suggestion
    curl -sLO https://github.com/kubevirt/cluster-network-addons-operator/releases/download/v"${VERSION}"/network-addons-config-example.cr.yaml
```

hack/update_releases.sh
```diff
@@ -0,0 +1,32 @@
...
+    curl -sLO https://github.com/kubevirt/cluster-network-addons-operator/releases/download/v${VERSION}/network-addons-config.crd.yaml
```
RamLavi, line 24: ditto
```suggestion
    curl -sLO https://github.com/kubevirt/cluster-network-addons-operator/releases/download/v"${VERSION}"/network-addons-config.crd.yaml
```

hack/update_releases.sh
```diff
@@ -0,0 +1,32 @@
...
+    curl -sLO https://github.com/kubevirt/cluster-network-addons-operator/releases/download/v${VERSION}/operator.yaml
```
RamLavi, line 25:
```suggestion
    curl -sLO https://github.com/kubevirt/cluster-network-addons-operator/releases/download/v"${VERSION}"/operator.yaml
```

hack/update_releases.sh
```diff
@@ -0,0 +1,32 @@
...
+    popd > /dev/null
+
+    pushd test/releases > /dev/null
+    curl -sLO https://raw.githubusercontent.com/kubevirt/cluster-network-addons-operator/release-${VERSION%.*}/test/releases/${VERSION}.go
```
RamLavi, line 29: ditto
```suggestion
    curl -sLO https://raw.githubusercontent.com/kubevirt/cluster-network-addons-operator/release-"${VERSION%.*}"/test/releases/"${VERSION}".go
```

RamLavi (changes requested):
+1 Like.

Hey, could you please add a short explanation before the `Usage example`? 
Some thing like :
```
This script pulls release manifests needed for lifecycle tests.
```

RamLavi:
> **Special notes for your reviewer**: Can be improved to read `RELEASES_SELECTOR` directly from `automation/check-patch.e2e-lifecycle-k8s.sh`. Can also be improved to purge non supported versions according lowest `RELEASES_SELECTOR` element.
> 

Throwing another improvement suggestion to the pot: Move the manifests to another place like ghcr.io and have tests download the manifest of each test before it runs.

hack/update_releases.sh
oshoval: There is no risk here in globbing and word splitting as the VERSION is known x.y.z format
Imo it makes the line less readable

hack/update_releases.sh
oshoval: same

hack/update_releases.sh
oshoval: same

hack/update_releases.sh
oshoval: same

hack/update_releases.sh
oshoval: same

hack/update_releases.sh
oshoval: same

hack/update_releases.sh
oshoval: same

hack/update_releases.sh
oshoval: same

oshoval:
> > **Special notes for your reviewer**: Can be improved to read `RELEASES_SELECTOR` directly from `automation/check-patch.e2e-lifecycle-k8s.sh`. Can also be improved to purge non supported versions according lowest `RELEASES_SELECTOR` element.
> 
> Throwing another improvement suggestion to the pot: Move the manifests to another place like ghcr.io and have tests download the manifest of each test before it runs.

Less good imo, because it makes it require network connections and prone for network hiccups

hack/update_releases.sh
RamLavi: It's common practice. IMO we should have all our scripts pass shell check, to prevent issues that may occur in the future (even if they don't manifest now).

oshoval:
Thanks
Addressed / commented

hack/update_releases.sh
oshoval: Atm we don't have shell check nor enforcing it, also shell check sometimes even produce bugs btw.
I am against adding it please.
It makes the code less readable, and not mandatory here, as the VERSION is a "fixed", deterministic string.

RamLavi: I agree that shell check is not perfect at all. I'm not suggesting to add a shell-check lane to check all scripts, nor do I think we should rework all the existing ones. 
However, I do think we should use it as a tool on new scripts, to make sure we are not falling to known issues, and by that strive to have our scripts as good as possible.

And as for this case - like you said it's not a problem right now so I don't consider it a major issue - and as such I am not going to block over it.

oshoval:
Thanks

/approve

PR #1887, opened 2024-09-09
bump kubevirt-ipam-controller to v0.1.7-alpha

oshoval:
Beside the bump, adapt the automation script for this bump:
Since https://github.com/kubevirt/ipam-extensions/pull/60
allows custom kubeconfig, we should set the line before cluster-up,
so it won't take the value set by automation/components-functests.setup.sh
which is correct only for kubevirtci providers, not for external ones.
Hence set it just before cluster-up.

**Release note**:

```release-note
bump kubevirt-ipam-controller to v0.1.7-alpha
```

### DIFF
--- a/automation/check-patch.e2e-kubevirt-ipam-controller-functests.sh
+++ b/automation/check-patch.e2e-kubevirt-ipam-controller-functests.sh
@@ -40,9 +40,9 @@ main() {
     COMPONENT="kubevirt-ipam-controller" source automation/components-functests.setup.sh
 
     cd ${TMP_COMPONENT_PATH}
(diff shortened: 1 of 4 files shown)


oshoval:
/test pull-e2e-cnao-kube-secondary-dns-functests 

i just want to see the base ref of it

oshoval:
IPAM fails here, the cluster is not reachable, something delicate it seems, as on clean main it pass
and the point it fails is way before the changes of this PR effect scope

oshoval:
https://github.com/kubevirt/cluster-network-addons-operator/actions/runs/10769003902/job/29859282715?pr=1887
passed now

PR #1888, opened 2024-09-09
tools/bumper: Fix latest vtag check

RamLavi:
In the bumper logic, in case either currentReleaseVersion or latestReleaseVersion are in virtual-tag format (i.e. v0.44.1-1-g4cd33665) - then the bump should be initiated. However the logic implementing it is wrong.
Fixing logic, adding unit tests.

**Release note**:

```release-note
NONE
```


RamLavi:
Ignoring the Sonar cloud issue. Duplicate lines in DescribeTable Entries are OK by design.

PR #1904, opened 2024-09-17
feat: clean task in Makefile

nestoracunablanco:
This update is beneficial for enabling multiarch support. By default, some tooling is downloaded specifically for amd64. Even with changes in the scripts, the tool does not get updated.

**Release note**:

```release-note NONE

```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,7 +38,8 @@ E2E_SUITES = \
 	test/e2e/workflow \
 	test/e2e/monitoring
 
-BIN_DIR = $(CURDIR)/build/_output/bin/
+OUTPUT_DIR = $(CURDIR)/build/_output/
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -223,6 +224,9 @@ lint-monitoring:
 	GOBIN=$$(pwd)/build/_output/bin/ $(GO) install -mod=mod github.com/kubevirt/monitoring/monitoringlinter/cmd/monitoringlinter@e2be790
 	$(MONITORING_LINTER) ./...
 
+clean:
```
RamLavi, line 227: please also add it to the PHONY list

nestoracunablanco: @RamLavi done

PR #1905, opened 2024-09-19
feat: enable download of go client for host arch

nestoracunablanco:
This commit is a step towards enabling builds for different architectures.

**Release note**:

```release-note NONE

```

### DIFF
--- a/hack/install-go.sh
+++ b/hack/install-go.sh
@@ -2,7 +2,8 @@
 
 destination=$1
 version=$(curl -s https://go.dev/dl/?mode=json | jq -r ".[0].version")
-tarball=$version.linux-amd64.tar.gz
+arch=$(uname -m | sed 's/x86_64/amd64/')
+tarball=$version.linux-$arch.tar.gz
 url=https://dl.google.com/go/
 
 mkdir -p $destination


nestoracunablanco:
@qinqon @RamLavi what do you think about this change?

hack/install-go.sh
```diff
@@ -2,7 +2,8 @@
 
 destination=$1
 version=$(curl -s https://go.dev/dl/?mode=json | jq -r ".[0].version")
-tarball=$version.linux-amd64.tar.gz
+arch=$(uname -m | sed 's/x86_64/amd64/')
```
RamLavi, line 5: I'm not sure I understand. 
If the machine running is `x86_64` then why would I want to substitute it to `amd64`?
Shouldn't it just be something like
```
arch=$(uname -m)
if [ "$arch" != "x86_64" ] && [ "$arch" != "amd64" ]; then
  echo "Unsupported architecture: $arch"
  exit 1
fi
```
?

nestoracunablanco: x86_64" and "amd64" are both valid terms used to refer to the same architecture. However, when you check the architecture via the command line using uname -m, it returns "x86_64". The Go programming language project, on the other hand, prefers the term "amd64." Therefore, if we are working with the amd64 architecture, we need to rename it accordingly.

RamLavi: ah I see. Learned something new :) 
Please add this explanation to the commit desc.

nestoracunablanco: @RamLavi done. Thanks for the feedback.

PR #1914, opened 2024-10-07
The origin-kube-rbac-proxy image is not compatible with the s390x

ashokpariya0:
architecture. This change deploys an image that is compatible with amd64, arm and s390x architecture.

The origin-kube-rbac-proxy image is not compatible with the s390x
architecture. This change deploys an image that is compatible with
amd64, arm and s390x.

**Release note**:

```release-note
This change addresses the compatibility issues with the origin-kube-rbac-proxy image on the s390x architecture. The updated deployment now uses an image that supports the amd64, arm and s390x architectures.
This enhancement ensures broader support and improved functionality across diverse environments.

```

### DIFF
--- a/pkg/components/components.go
+++ b/pkg/components/components.go
@@ -37,7 +37,7 @@ const (
 	KubeMacPoolImageDefault            = "quay.io/kubevirt/kubemacpool@sha256:677971a25ff3ce95d9e6ecc86090a09f0ae691ce67b8349384f1881562feed7b"
 	OvsCniImageDefault                 = "ghcr.io/k8snetworkplumbingwg/ovs-cni-plugin@sha256:54be8fcacee50af64deafa9e99f3fe079033630c00c4ed9f74d17b0d91009f10"
 	MacvtapCniImageDefault             = "quay.io/kubevirt/macvtap-cni@sha256:850b89343ace7c7ea6b18dd8e11964613974e9d1f7377af03854d407fb15230a"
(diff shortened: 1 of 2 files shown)


oshoval:
/hold

thanks but this cant be done this way
we should do it correctly
image cant just be changed

oshoval:
please open an issue instead and we will see what can be done and if we should support it upstream

ashokpariya0:
> please open an issue instead and we will see what can be done and if we should support it upstream
Thanks, done that. Issue link: https://github.com/kubevirt/cluster-network-addons-operator/issues/1916

oshoval:
thanks
lets continue there

/close

oshoval:
maybe meanwhile in case you do cluster-sync you can use this env var locally ?
https://github.com/kubevirt/cluster-network-addons-operator/blob/92344ad417c12899958a298556027e8bdb8883c3/hack/generate-manifests.sh#L28

oshoval:
Checked again, wrt auto bumper, i think in this specific case we dont need to update bumper scripts as this parameter is part of CNAO and not part of the components.
So your PR is fine.

What about all the other images, some of them arent multiarch right ? 
We won't able to maintain it right now, but will prioritize it as a task for all the images once we can.
If you plan to contribute and able to give support if it breaks we might able to take your PR, thanks.

Also please consider that kube-rbac-proxy need to be changed on https://github.com/k8snetworkplumbingwg/kubemacpool so if KMP is installed standalone or via CNAO it will be the same

WRT the proposed image, it seems safe to use it as `https://github.com/brancz/kube-rbac-proxy` is the one 
`https://github.com/openshift/kube-rbac-proxy` is forked from

@phoracek agree ? anything i missed ?
Thanks

Issue #1916, opened 2024-10-08
Enable multi-architecture support, specifically for s390x, in CNAO and the associated CNIs.


ashokpariya0:
**Is your feature request related to a problem? Please describe**:
Provide multi-platform support for CNAO and all CNIs supported by CNAO.

**Describe the solution you'd like**:
We would like to see multi-architecture support for CNAO and all of its supported CNIs, including:

multus
multusDynamicNetworks
linuxBridge
kubeMacPool
ovs
macvtap
kubeSecondaryDNS
kubevirtIpamController

**Describe alternatives you've considered**:

ormergi:
@ashokpariya0 could you please elaborate on the motivation for supporting s390x arch? 
Is there some tracking issue or enhancement proposal for supporting s390x?

ashokpariya0:
We are enabling s390x (IBM Z system) support in all(most) kubevirt ecosystem components. Already [kubevirt](https://github.com/kubevirt/kubevirt/pull/10490) (virt-operator), supports s390x and now we are working on additional components like cnao, cdi(almost done), ssp, etc.

oshoval:
Checked again, wrt auto bumper, i think in this specific case we dont need to update bumper scripts as this parameter is part of CNAO and not part of the components.
So your PR is fine.

What about all the other images, some of them arent multiarch right ?
We won't able to maintain it right now, but will prioritize it as a task for all the images once we can.
If you plan to contribute and able to give support if it breaks we might able to take your PR, thanks.

Also please consider that kube-rbac-proxy need to be changed on https://github.com/k8snetworkplumbingwg/kubemacpool so if KMP is installed standalone or via CNAO it will be the same

WRT the proposed image, it seems safe to use it as https://github.com/brancz/kube-rbac-proxy is the one
https://github.com/openshift/kube-rbac-proxy is forked from

@phoracek agree ? anything i missed ?
Thanks

ashokpariya0:
> Checked again, wrt auto bumper, i think in this specific case we dont need to update bumper scripts as this parameter is part of CNAO and not part of the components. So your PR is fine.
> 
>   @oshoval Thank you for your feedback. 
> What about all the other images, some of them arent multiarch right ? We won't able to maintain it right now, but will prioritize it as a task for all the images once we can. If you plan to contribute and able to give support if it breaks we might able to take your PR, thanks. Maybe we first need to map all the required changes over all repos and just then start by fixing one by one.
> 
[...]

PR #945, opened 2021-07-27
components: Consume CNI Bridge with macspoofchk support

EdDev:
**What this PR does / why we need it**:

As an intermediate and temporary step, consume the changes posted to the
CNI bridge [1].

Once the PR is merged into the original repo, this change will be
reverted.

In case the PR will not be merged into the original repo, we will
proceed with a more formal resolution.

[1] https://github.com/containernetworking/plugins/pull/639

**Special notes for your reviewer**:

**Release note**:

```release-note
Consume CNI Bridge with mac-spoof-check support.
```

PR #1917, opened 2024-10-09
The origin-kube-rbac-proxy image is not compatible with the s390x

ashokpariya0:
architecture. This change deploys an image that is multi arch supported.

The origin-kube-rbac-proxy image is not compatible with the s390x architecture. 
This change deploys an image that is multi arch supported and this will enable
CNAO on s390x architecture to get started and test CNAO supported CNI plugins

proposed image(https://quay.io/repository/brancz/kube-rbac-proxy/manifest/sha256:e6a323504999b2a4d2a6bf94f8580a050378eba0900fd31335cf9df5787d9a9b) is built from parent repo of https://github.com/openshift/kube-rbac-proxy.

**Release note**:

```release-note
None
```

### DIFF
--- a/pkg/components/components.go
+++ b/pkg/components/components.go
@@ -37,7 +37,7 @@ const (
 	KubeMacPoolImageDefault            = "quay.io/kubevirt/kubemacpool@sha256:677971a25ff3ce95d9e6ecc86090a09f0ae691ce67b8349384f1881562feed7b"
 	OvsCniImageDefault                 = "ghcr.io/k8snetworkplumbingwg/ovs-cni-plugin@sha256:54be8fcacee50af64deafa9e99f3fe079033630c00c4ed9f74d17b0d91009f10"
 	MacvtapCniImageDefault             = "quay.io/kubevirt/macvtap-cni@sha256:850b89343ace7c7ea6b18dd8e11964613974e9d1f7377af03854d407fb15230a"
-	KubeRbacProxyImageDefault          = "quay.io/openshift/origin-kube-rbac-proxy@sha256:e2def4213ec0657e72eb790ae8a115511d5b8f164a62d3568d2f1bff189917e8"
+	KubeRbacProxyImageDefault          = "quay.io/brancz/kube-rbac-proxy@sha256:e6a323504999b2a4d2a6bf94f8580a050378eba0900fd31335cf9df5787d9a9b"
 	KubeSecondaryDNSImageDefault       = "ghcr.io/kubevirt/kubesecondarydns@sha256:44489bdf8affb0e82b68a7100bbee7d289151ada54ab2e76a9d3afef769a1f54"
 	CoreDNSImageDefault                = "registry.k8s.io/coredns/coredns@sha256:a0ead06651cf580044aeb0a0feba63591858fb2e43ade8c9dea45a6a89ae7e5e"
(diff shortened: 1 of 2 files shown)


ashokpariya0:
Please check issue (https://github.com/kubevirt/cluster-network-addons-operator/issues/1916) for details.

oshoval:
you need please also the 2nd change that you have here
https://github.com/kubevirt/cluster-network-addons-operator/pull/1914

oshoval:
nit: the word architecture should be part of the title, please consider making the title shorter on PR and commit
and then to update the PR desc accordingly

ashokpariya0:
> Thanks You need please also the 2nd change that you have here #1914
I was thinking to not to change kubemacpool-mac-controller-manager image with proposed image now as this is not needed for now, Will follow up issue with on  https://github.com/openshift/kube-rbac-proxy, Is that okay? What do you think?

oshoval:
Thanks

/test all

ashokpariya0:
> It won't work otherwise CNAO deploy the image both for itself and for KMP so both need to have the same expected image in the test
> 
> another ortogonal PR can change it on KMP, but this PR and KMP PR can be merged in any order they dont affect each other

Alright, It's done.

oshoval:
can you please change also on https://github.com/k8snetworkplumbingwg/kubemacpool? 
so we will have both aligned

oshoval:
@phoracek 
would need your ack here please (once all tests pass)

all reasoning on the issue, for me it looks fine fwiw

oshoval:
might be just a flake 
https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_cluster-network-addons-operator/1917/pull-e2e-cluster-network-addons-operator-workflow-k8s/1843990810543001600

/retest

ashokpariya0:
> can you please change also on https://github.com/k8snetworkplumbingwg/kubemacpool? so we will have both aligned (even that it is orthogonal it is better like that)
>
Done,  Please check [PR](https://github.com/k8snetworkplumbingwg/kubemacpool/pull/440).

oshoval:
/lgtm

imo it is good to start like that, allowing CNAO operator itself to be robust and to already support the ones that are multi
and then to tackle one by one as the issue says

but lets see what @phoracek says

Thanks

phoracek:
@oshoval I understand this new image is the real upstream of the one we used to use. I'm not against this change. I won't LGTM just because I don't know whether there any other changes needed to make the image change fit into our automation.

ashokpariya0:
> @oshoval I understand this new image is the real upstream of the one we used to use. I'm not against this change. I won't LGTM just because I don't know whether there any other changes needed to make the image change fit into our automation.
>
@phoracek Thank you for your feedback! I appreciate your understanding regarding the upstream image change. What steps should we take next to ensure our automation tests setup is compatible? It's important to note that without this change, we won't be able to run the CNAO operator on the s390x architecture. I’d welcome any suggestions you might have for a solution.

phoracek:
@ashokpariya0 looking at an older PR that changed an image source https://github.com/kubevirt/cluster-network-addons-operator/pull/945/files#diff-b62a52052b6d9294695602994bd32db41de0d3c71d85ab711551874740eaf0feL42. Chances are your own PR is enough. I'm not all that familiar with the image/bump pipeline of CNAO, I would need @oshoval / @RamLavi  to ack this

oshoval:
Yep it is enough
Kube-rbac-proxy is used by CNAO itself as well, so it doesn't have auto bumping atm via bump-x files

Thanks
/lgtm

PR #1920, opened 2024-10-14
feat: enable s390x prometheus check

nestoracunablanco:
As of version 2.15.2, there is no available container image for s390x. However, version 2.25.2 does provide support for this architecture. This update enables the checks to run on s390x.

**Release note**:

```release-note NONE

```


oshoval:
Thanks
Maybe worth to add it to kubevirtci (if it is used there ?, otherwise fine here) 

but we can just continue the current flow as is, and do what you suggested
/lgtm

PR #1922, opened 2024-10-25
events: Remove duplicated 'Available' events

qinqon:
CNAO is emitting  config 'Availabe' event after each reconcile even when nothing changes, this add noise to users. This change emits the event only if Available changes from false to true.

Some test suites like openshift tests fail if it sees too many duplicated events during a period of times, so it's fair to fix that by not sending events when the state didn't change.

**Release note**:

```release-note
events: Remove duplicated 'Available' events
```


maiqueb (approved):
Good ! 

Let's be less noisy.

PR #1923, opened 2024-11-08
Enable multi-platform support for operator image builds

ashokpariya0:
These changes enable building and pushing CNAO operator image for multiple platforms (amd64, s390x, arm64) from a single Dockerfile. Enhanced multi-platform support in the build process by adding a PLATFORMS argument in the Makefile for amd64, s390x, and arm64 architectures. Updated the docker-build-operator target to support builds with docker buildx and podman, and added new targets for creating Docker Buildx builders and pushing multi-platform images.

The existing CNAO operator ([link](https://quay.io/repository/kubevirt/cluster-network-addons-operator?tab=tags&tag=latest)) currently supports only the amd64 architecture. These changes extend multi-platform support to include arm64 and s390x architectures, enabling broader compatibility for the CNAO operator image.

Please note that Docker and Podman handle multi-architecture image builds differently. As a result, the changes have been made accordingly. Additionally, Go cross-compilation support has been added to build binaries for the target architecture.
Added a PLATFORMS argument in the Makefile to support building images for multiple architectures (amd64, s390x, arm64).
Updated the Dockerfile to dynamically handle multi-architecture builds.
Added Go cross-compilation support in Dockerfile to build binaries for the target architecture.
Enhanced the docker-build-operator target to enable multi-platform builds using both docker buildx and podman.
Introduced a new build-multiarch-operator-image target to build and push multi-platform images using podman.
Added a create-builder target to generate Docker Buildx builders for multi-platform image support.

**Release note**:
```release-note
None
```

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -12,9 +12,6 @@
 _kubevirtci
 
 # No need for source code is already compiled
-pkg
-cmd
-tools
 vendor
 docs
 
--- a/Makefile
+++ b/Makefile
@@ -13,6 +13,15 @@ OPERATOR_IMAGE ?= cluster-network-addons-operator
 REGISTRY_IMAGE ?= cluster-network-addons-registry
 export OCI_BIN ?= $(shell if podman ps >/dev/null 2>&1; then echo podman; elif docker ps >/dev/null 2>&1; then echo docker; fi)
 TLS_SETTING := $(if $(filter $(OCI_BIN),podman),--tls-verify=false,)
+PLATFORMS ?= linux/amd64
+# Set the platforms for building a multi-platform supported image.
+# Example:
+# PLATFORMS ?= linux/amd64,linux/arm64,linux/s390x
+# Alternatively, you can export the PLATFORMS variable like this:
+# export PLATFORMS=linux/arm64,linux/s390x,linux/amd64
+ARCH := $(shell uname -m | sed 's/x86_64/amd64/')
+DOCKER_BUILDER ?= cnao-docker-builder
+OPERATOR_IMAGE_TAGGED := $(IMAGE_REGISTRY)/$(OPERATOR_IMAGE):$(IMAGE_TAG)
 
 TARGETS = \
 	gen-k8s \
@@ -26,7 +35,6 @@ TARGETS = \
 export GOFLAGS=-mod=vendor
 export GO111MODULE=on
 GO_VERSION = $(shell hack/go-version.sh)
-
 WHAT ?= ./pkg/... ./cmd/... ./tools/...
 
 export E2E_TEST_TIMEOUT ?= 3h
@@ -112,8 +120,14 @@ manifest-templator: $(GO)
 
 docker-build: docker-build-operator docker-build-registry
 
-docker-build-operator: manager manifest-templator
-	$(OCI_BIN) build -f build/operator/Dockerfile -t $(IMAGE_REGISTRY)/$(OPERATOR_IMAGE):$(IMAGE_TAG) .
+docker-build-operator:
+ifeq ($(OCI_BIN),podman)
+	$(MAKE) build-multiarch-operator-podman
+else ifeq ($(OCI_BIN),docker)
+	$(MAKE) build-multiarch-operator-docker
+else
+	$(error Unsupported OCI_BIN value: $(OCI_BIN))
+endif
 
 docker-build-registry:
 	$(OCI_BIN) build -f build/registry/Dockerfile -t $(IMAGE_REGISTRY)/$(REGISTRY_IMAGE):$(IMAGE_TAG) .
@@ -121,7 +135,9 @@ docker-build-registry:
 docker-push: docker-push-operator docker-push-registry
 
 docker-push-operator:
-	$(OCI_BIN) push ${TLS_SETTING} $(IMAGE_REGISTRY)/$(OPERATOR_IMAGE):$(IMAGE_TAG)
+ifeq ($(OCI_BIN),podman)
+	podman manifest push ${TLS_SETTING} ${OPERATOR_IMAGE_TAGGED} ${OPERATOR_IMAGE_TAGGED}
+endif
 
 docker-push-registry:
 	$(OCI_BIN) push $(IMAGE_REGISTRY)/$(REGISTRY_IMAGE):$(IMAGE_TAG)
@@ -227,9 +243,17 @@ lint-monitoring:
 clean:
 	rm -rf $(OUTPUT_DIR)
 
+build-multiarch-operator-docker:
+	ARCH=$(ARCH) PLATFORMS=$(PLATFORMS) OPERATOR_IMAGE_TAGGED=$(OPERATOR_IMAGE_TAGGED) DOCKER_BUILDER=$(DOCKER_BUILDER) ./hack/build-operator-docker.sh
+
+build-multiarch-operator-podman:
+	ARCH=$(ARCH) PLATFORMS=$(PLATFORMS) OPERATOR_IMAGE_TAGGED=$(OPERATOR_IMAGE_TAGGED) ./hack/build-operator-podman.sh
+
 .PHONY: \
 	$(E2E_SUITES) \
 	all \
+	build-multiarch-operator-docker \
+	build-multiarch-operator-podman \
 	check \
 	cluster-clean \
 	cluster-down \
--- a/build/operator/Dockerfile
+++ b/build/operator/Dockerfile
@@ -1,4 +1,24 @@
-FROM quay.io/centos/centos:stream9
+ARG BUILD_ARCH=amd64
+FROM --platform=linux/${BUILD_ARCH} quay.io/centos/centos:stream9 AS builder
+
+RUN dnf install -y tar gzip jq && dnf clean all
+RUN ARCH=$(uname -m | sed 's/x86_64/amd64/') && \
+    GO_VERSION=$(curl -L -s "https://go.dev/dl/?mode=json" | jq -r '.[0].version') && \
+    curl -L "https://go.dev/dl/${GO_VERSION}.linux-${ARCH}.tar.gz" -o go.tar.gz && \
+    tar -C /usr/local -xzf go.tar.gz && \
+    rm go.tar.gz
+ENV PATH=$PATH:/usr/local/go/bin
+WORKDIR /go/src/cluster-network-addons-operator
+COPY . .
+
+ARG TARGETOS
+ARG TARGETARCH
+
+RUN CGO_ENABLED=0 GOOS=${TARGETOS} GOARCH=${TARGETARCH} go build -o build/_output/bin/manager ./cmd/... && \
+    CGO_ENABLED=0 GOOS=${TARGETOS} GOARCH=${TARGETARCH} go build -o build/_output/bin/manifest-templator ./tools/manifest-templator/...
+
+FROM --platform=linux/${TARGETARCH} quay.io/centos/centos:stream9 AS final
+
 ENV ENTRYPOINT=/entrypoint \
     OPERATOR=/cluster-network-addons-operator \
     MANIFEST_TEMPLATOR=/manifest-templator \
@@ -8,13 +28,16 @@ ENV ENTRYPOINT=/entrypoint \
 RUN \
     yum -y update && \
     yum clean all
+
 COPY build/operator/bin/user_setup /user_setup
 COPY build/operator/bin/csv-generator /usr/bin/csv-generator
 COPY templates/cluster-network-addons/VERSION/cluster-network-addons-operator.VERSION.clusterserviceversion.yaml.in cluster-network-addons-operator.VERSION.clusterserviceversion.yaml.in
 RUN /user_setup
 COPY data /data
-COPY build/_output/bin/manager $OPERATOR
-COPY build/_output/bin/manifest-templator $MANIFEST_TEMPLATOR
+
+COPY --from=builder /go/src/cluster-network-addons-operator/build/_output/bin/manager $OPERATOR
+COPY --from=builder /go/src/cluster-network-addons-operator/build/_output/bin/manifest-templator $MANIFEST_TEMPLATOR
 COPY build/operator/bin/entrypoint $ENTRYPOINT
+
 ENTRYPOINT $ENTRYPOINT
 USER $USER_UID
--- a/hack/build-operator-docker.sh
+++ b/hack/build-operator-docker.sh
@@ -1,0 +1,17 @@
+#!/bin/bash
+
+if [ -z "$ARCH" ] || [ -z "$PLATFORMS" ] || [ -z "$OPERATOR_IMAGE_TAGGED" ]; then
+    echo "Error: ARCH, PLATFORMS and OPERATOR_IMAGE_TAGGED must be set."
+    exit 1
+fi
+
+IFS=',' read -r -a PLATFORM_LIST <<< "$PLATFORMS"
+
+BUILD_ARGS="--build-arg BUILD_ARCH=$ARCH -f build/operator/Dockerfile -t $OPERATOR_IMAGE_TAGGED --push ."
+
+if [ ${#PLATFORM_LIST[@]} -eq 1 ]; then
+    docker build --platform "$PLATFORMS" $BUILD_ARGS
+else
+    ./hack/init-buildx.sh "$DOCKER_BUILDER"
+    docker buildx build --platform "$PLATFORMS" $BUILD_ARGS
+fi
--- a/hack/build-operator-podman.sh
+++ b/hack/build-operator-podman.sh
@@ -1,0 +1,24 @@
+#!/bin/bash
+
+if [ -z "$ARCH" ] || [ -z "$PLATFORMS" ] || [ -z "$OPERATOR_IMAGE_TAGGED" ]; then
+    echo "Error: ARCH, PLATFORMS, and OPERATOR_IMAGE_TAGGED must be set."
+    exit 1
+fi
+
+IFS=',' read -r -a PLATFORM_LIST <<< "$PLATFORMS"
+
+# Remove any existing manifest and image
+podman manifest rm "${OPERATOR_IMAGE_TAGGED}" || true
+podman rmi "${OPERATOR_IMAGE_TAGGED}" || true
+
+podman manifest create "${OPERATOR_IMAGE_TAGGED}"
+
+for platform in "${PLATFORM_LIST[@]}"; do
+    podman build \
+        --no-cache \
+        --build-arg BUILD_ARCH="$ARCH" \
+        --platform "$platform" \
+        --manifest "${OPERATOR_IMAGE_TAGGED}" \
+        -f build/operator/Dockerfile \
+        .
+done
--- a/hack/init-buildx.sh
+++ b/hack/init-buildx.sh
@@ -1,0 +1,55 @@
+#!/bin/bash
+
+check_buildx() {
+  export DOCKER_CLI_EXPERIMENTAL=enabled
+
+  if ! docker buildx > /dev/null 2>&1; then
+     mkdir -p ~/.docker/cli-plugins
+     BUILDX_VERSION=$(curl -s https://api.github.com/repos/docker/buildx/releases/latest | jq -r .tag_name)
+     curl -L https://github.com/docker/buildx/releases/download/"${BUILDX_VERSION}"/buildx-"${BUILDX_VERSION}".linux-amd64 --output ~/.docker/cli-plugins/docker-buildx
+     chmod a+x ~/.docker/cli-plugins/docker-buildx
+  fi
+}
+
+create_or_use_buildx_builder() {
+  local builder_name=$1
+  if [ -z "$builder_name" ]; then
+    echo "Error: Builder name is required."
+    exit 1
+  fi
+
+  check_buildx
+
+  current_builder="$(docker buildx inspect "${builder_name}")"
+
+  if ! grep -q "^Driver: docker$" <<<"${current_builder}" && \
+     grep -q "linux/amd64" <<<"${current_builder}" && \
+     grep -q "linux/arm64" <<<"${current_builder}" && \
+     grep -q "linux/s390x" <<<"${current_builder}"; then
+    echo "The current builder already has multi-architecture support (amd64, arm64, s390x)."
+    echo "Skipping setup as the builder is already configured correctly."
+    exit 0
+  fi
+
+  # Check if the builder already exists by parsing the output of `docker buildx ls`
+  # We check if the builder_name appears in the list of active builders
+  existing_builder=$(docker buildx ls | grep -w "$builder_name" | awk '{print $1}')
+
+  if [ -n "$existing_builder" ]; then
+    echo "Builder '$builder_name' already exists."
+    echo "Using existing builder '$builder_name'."
+    docker buildx use "$builder_name"
+  else
+    echo "Creating a new Docker Buildx builder: $builder_name"
+    docker buildx create --driver-opt network=host --use --name "$builder_name"
+    echo "The new builder '$builder_name' has been created and set as active."
+  fi
+}
+
+if [ $# -eq 1 ]; then
+  create_or_use_buildx_builder "$1"
+else
+  echo "Usage: $0 <builder_name>"
+  echo "Example: $0 mybuilder"
+  exit 1
+fi


.dockerignore
```diff
@@ -12,9 +12,9 @@
 _kubevirtci
 
 # No need for source code is already compiled
-pkg
-cmd
-tools
+#pkg
```
oshoval, line 15: leftovers?

Makefile
```diff
@@ -13,6 +13,14 @@ OPERATOR_IMAGE ?= cluster-network-addons-operator
 REGISTRY_IMAGE ?= cluster-network-addons-registry
 export OCI_BIN ?= $(shell if podman ps >/dev/null 2>&1; then echo podman; elif docker ps >/dev/null 2>&1; then echo docker; fi)
 TLS_SETTING := $(if $(filter $(OCI_BIN),podman),--tls-verify=false,)
+PLATFORMS ?= linux/amd64,linux/s390x,linux/arm64
```
oshoval, line 16: lets have just amd64 by default, so it wont take time on each regular use ?
the rest can be in an example comment

moreover if we build just one image, we dont need buildx just docker build, much easier to maintain, usually this will be the case, unless on releases and such once time come

Makefile
```diff
@@ -13,6 +13,14 @@ OPERATOR_IMAGE ?= cluster-network-addons-operator
...
+ARCH := $(shell uname -m | sed 's/x86_64/amd64/')
+DOCKER_BUILDER ?= docker-builder
+OPERATOR_IMAGE_TAGGED := $(IMAGE_REGISTRY)/$(OPERATOR_IMAGE):$(IMAGE_TAG)
+null  :=
+space := $(null) #
+comma := ,
```
oshoval, line 22: can we just inline those please?

Makefile
```diff
@@ -112,16 +120,30 @@ manifest-templator: $(GO)
...
+ifeq ($(OCI_BIN),podman)
+	$(MAKE) build-multiarch-operator-image
+else ifeq ($(OCI_BIN),docker)
+	$(MAKE) create-builder
+	docker buildx build \
+        --build-arg BUILD_ARCH=$(ARCH) \
+        --platform $(PLATFORMS) \
+        -f build/operator/Dockerfile \
+        -t ${OPERATOR_IMAGE_TAGGED} \
+        --push .
```
oshoval, line 133: please consider moving it under create-builder target
unless you think it doesnt fit there?

Makefile
```diff
@@ -227,9 +249,33 @@ lint-monitoring:
 clean:
 	rm -rf $(OUTPUT_DIR)
 
+# Target to create a new Docker Buildx builder
+create-builder:
+	./hack/build-multiarch-operator.sh ${DOCKER_BUILDER}
+
+# Target to build the operator image with support for multiple platforms
+build-multiarch-operator-image:
+	# Remove any existing manifest and image
```
oshoval, line 258: please drop the comments (all over unless it is not trivial)

build/operator/Dockerfile
```diff
@@ -1,20 +1,54 @@
-FROM quay.io/centos/centos:stream9
+# Define build-time arguments
+ARG BUILD_ARCH=amd64            # Default value for architecture is amd64
+ARG GO_VERSION="1.23.3"         # Default Go version is 1.23.3
```
oshoval, line 3: might worth to auto take it from go.mod
there are few network repos that doing something like that

build/operator/Dockerfile
```diff
@@ -1,20 +1,54 @@
...
+
+FROM --platform=linux/${BUILD_ARCH} quay.io/centos/centos:stream9 AS builder
+ARG GO_VERSION
+RUN dnf install -y tar gzip && dnf clean all
+RUN ARCH=$(uname -m | sed 's/x86_64/amd64/') && \
+    curl -L https://go.dev/dl/go${GO_VERSION}.linux-${ARCH}.tar.gz -o go.tar.gz && \
+    tar -C /usr/local -xzf go.tar.gz && \
+    rm go.tar.gz
+
```
oshoval, line 12: maybe using docker ADD can simplify some of this ?

build/operator/Dockerfile
```diff
@@ -1,20 +1,54 @@
...
 COPY build/operator/bin/user_setup /user_setup
 COPY build/operator/bin/csv-generator /usr/bin/csv-generator
 COPY templates/cluster-network-addons/VERSION/cluster-network-addons-operator.VERSION.clusterserviceversion.yaml.in cluster-network-addons-operator.VERSION.clusterserviceversion.yaml.in
 RUN /user_setup
 COPY data /data
-COPY build/_output/bin/manager $OPERATOR
-COPY build/_output/bin/manifest-templator $MANIFEST_TEMPLATOR
+
+# Copy the final built binaries into the image
```
oshoval, line 48: please drop comments all over unless non trivial

hack/build-multiarch-operator.sh
```diff
@@ -0,0 +1,54 @@
+#!/bin/bash
+
+# Function to check if Docker Buildx is available
```
oshoval, line 3: please drop comments unless non trivial

oshoval (commented):
Nice thanks
First pass

hack/build-multiarch-operator.sh
oshoval, line 1: might worth to consider doing as close to 
https://github.com/nmstate/kubernetes-nmstate/blob/2e791f9d650fe3cd4844e3964a2fea49688466b9/hack/init-buildx.sh please

Makefile
ashokpariya0: @oshoval Thanks for the feedback! I agree that simplifying the default platform to just amd64 will streamline the build process and reduce unnecessary overhead. I'll update the configuration accordingly and move the other platforms into an example comment as you suggested.

As for the use of docker build vs. buildx, that approach makes sense for most cases. I'll switch to docker build for simplicity and reserve buildx for when it's needed.

During releases, we'll set PLATFORMS ?= linux/amd64,linux/s390x,linux/arm64, use buildx, and build and push a multi-platform image for the operator. For all other cases, we'll default to using amd64 as the platform. Does that sound good?

build/operator/Dockerfile
ashokpariya0: Good suggestion! I'll look into implementing that.

.dockerignore
ashokpariya0: No, this change is necessary because we need to build the two binaries listed [here](https://github.com/kubevirt/cluster-network-addons-operator/blob/main/Makefile#L107-L111) inside the Dockerfile in order to support Go cross-compilation for building the target binaries.

oshoval: lets remove the lines then completely ?

Makefile
oshoval: sgtm thx

.dockerignore
ashokpariya0: Yes, Removed in the latest commit.

Makefile
ashokpariya0: Updated in the latest commit.

Makefile
ashokpariya0: Done.

build/operator/Dockerfile
ashokpariya0: Done

hack/build-multiarch-operator.sh
ashokpariya0: Done.

hack/build-multiarch-operator.sh
ashokpariya0: Updated in latest commit.
FYI- It would be preferable to pass the builder name at line 30 in the script [init-buildx.sh](https://github.com/nmstate/kubernetes-nmstate/blob/2e791f9d650fe3cd4844e3964a2fea49688466b9/hack/init-buildx.sh#L30).

ashokpariya0:
@oshoval  The job [here](https://github.com/kubevirt/cluster-network-addons-operator/actions/runs/11794285408/job/32851563920?pr=1923) failed with the following error:

make[2]: Leaving directory '/tmp/cnao/cluster-network-addons-operator'
**podman manifest push --tls-verify=false localhost:5000/cluster-network-addons-operator:latest
Error: accepts 2 arg(s), received 1**
make[1]: *** [Makefile:143: docker-push-operator] Error 125
make[1]: Leaving directory '/tmp/cnao/cluster-network-addons-operator'
make: *** [Makefile:168: cluster-operator-push] Error 2
Issue Description:

The error indicates that the podman manifest push command is receiving only one argument (localhost:5000/cluster-network-addons-operator:latest), but it expects two arguments (the manifest and the destination).

The command works as expected in my local environment, where I'm using podman version 4.9.4-rhel.
It seems that the issue arises in the CI pipeline, which may be using a different version of Podman, leading to different behavior. Any thoughts?

oshoval:
@phoracek @RamLavi @0xFelix 
please consider also reviewing if you can, to see i dont miss anything

hack/build-multiarch-operator.sh
oshoval: can we make it even closer to that script ?
will be easier to maintain, it still seems too close to the original version if i am not wrong

Makefile
oshoval: please remove more comments, there are still some trivial ones
we usually add as few as possible
only for non trivial stuff

Makefile
```diff
@@ -112,16 +124,30 @@ manifest-templator: $(GO)
...
 docker-push: docker-push-operator docker-push-registry
 
 docker-push-operator:
-	$(OCI_BIN) push ${TLS_SETTING} $(IMAGE_REGISTRY)/$(OPERATOR_IMAGE):$(IMAGE_TAG)
+ifeq ($(OCI_BIN),podman)
+	podman manifest push ${TLS_SETTING} ${OPERATOR_IMAGE_TAGGED} ${OPERATOR_IMAGE_TAGGED}
+else ifeq ($(OCI_BIN),docker)
+ifeq ($(words $(PLATFORM_LIST)), 1)
+	docker push ${TLS_SETTING} ${OPERATOR_IMAGE_TAGGED}
+endif
```
oshoval, line 147: please try to think about a makefile target structure that can be simpler wrt to multi platform and one platform
the logic is split here and on docker-build-operator, it will be better to have it concentrated if possible (i understand it might be not possible because there are stages)

oshoval (commented):
Thanks,
there are some comments that were not addressed / replied to
please check those as well

Makefile
ashokpariya0: For docker buildx, we need to combine both build and push in a single command because the images are not stored locally by default, unlike with docker build. As a result, we don't need a separate docker push command when working with multi-platform builds. To simplify the logic and maintain consistency, we could consider adding the --push option to docker build as well. What do you think @oshoval

oshoval: might be interesting to try if you think it will simplify it please

hack/build-multiarch-operator.sh
ashokpariya0: Yes, I did that.
Just to let you know, I noticed an issue in [this line](https://github.com/nmstate/kubernetes-nmstate/blob/2e791f9d650fe3cd4844e3964a2fea49688466b9/hack/init-buildx.sh#L30) of init-buildx.sh — we are not performing the inspection based on the builder name.

Makefile
ashokpariya0: Done.

build/operator/Dockerfile
ashokpariya0: I made the changes according to the script [install-go.sh](https://github.com/kubevirt/cluster-network-addons-operator/blob/main/hack/install-go.sh). With the current setup, we are compiling and building the manager binary using the Go version specified above.

Please note that the script [go-version.sh](https://github.com/kubevirt/cluster-network-addons-operator/blob/main/hack/go-version.sh) returns the Go version in major.minor format. Additionally, the command curl -L -s "https://go.dev/dl/?mode=json" does not list Go 1.21, as it's archived. Before Go 1.22, we would need to download it directly from the archive. However, the install-go.sh script currently downloads the latest Go version and proceeds to compile with it.

I’ve made the same update in the Dockerfile to ensure consistency.

hack/build-multiarch-operator.sh
oshoval: thanks, if you think it worth it you can open an issue on nmstate repo
i must work on some other stuff at this very moment, so will take me bit time to review

0xFelix (commented):
Perhaps an unpopular opinion, but why do we need to maintain two builds for Podman and Docker? In https://github.com/kubevirt/ssp-operator we went with Podman only.

See https://github.com/kubevirt/ssp-operator/pull/1091

oshoval:
> Perhaps an unpopular opinion, but why do we need to maintain two builds for Podman and Docker? In https://github.com/kubevirt/ssp-operator we went with Podman only.
> 
> See [kubevirt/ssp-operator#1091](https://github.com/kubevirt/ssp-operator/pull/1091)

personally i am against dropping docker support
it is convenient it exists, and also on kubevirt we still do that for some jobs
moreoever, cnao depends on lot of other component, so it better to be robust in case something happens and is required

While it is a nice question, not sure if the PR is the right place for it (i.e it shouldn't block it)

Makefile
```diff
@@ -13,6 +13,19 @@ OPERATOR_IMAGE ?= cluster-network-addons-operator
...
+# Example:
+# PLATFORMS ?= linux/amd64,linux/arm64,linux/s390x
+# Alternatively, you can export the PLATFORMS variable like this:
+# export PLATFORMS=linux/arm64,linux/s390x,linux/amd64
+ARCH := $(shell uname -m | sed 's/x86_64/amd64/')
+DOCKER_BUILDER ?= cnao-docker-builder
+OPERATOR_IMAGE_TAGGED := $(IMAGE_REGISTRY)/$(OPERATOR_IMAGE):$(IMAGE_TAG)
+NULL  :=
+SPACE := $(NULL) #
+COMMA := ,
```
0xFelix, line 27: Are these vars really required?

Makefile
```diff
@@ -112,16 +124,24 @@ manifest-templator: $(GO)
...
+endif
 
 docker-build-registry:
 	$(OCI_BIN) build -f build/registry/Dockerfile -t $(IMAGE_REGISTRY)/$(REGISTRY_IMAGE):$(IMAGE_TAG) .
 
 docker-push: docker-push-operator docker-push-registry
 
 docker-push-operator:
-	$(OCI_BIN) push ${TLS_SETTING} $(IMAGE_REGISTRY)/$(OPERATOR_IMAGE):$(IMAGE_TAG)
+ifeq ($(OCI_BIN),podman)
```
0xFelix, line 142: This is called docker-push-operator but uses podman?

Makefile
```diff
@@ -227,9 +247,44 @@ lint-monitoring:
 clean:
 	rm -rf $(OUTPUT_DIR)
 
+build-multiarch-operator-docker:
+ifeq ($(words $(PLATFORM_LIST)), 1)
+	docker build \
+            --build-arg BUILD_ARCH=$(ARCH) \
+            --platform $(PLATFORMS) \
+            -f build/operator/Dockerfile \
+            -t ${OPERATOR_IMAGE_TAGGED} --push .
```
0xFelix, line 256: Can you reuse/deduplicate the args?

hack/build-multiarch-operator.sh
```diff
@@ -0,0 +1,61 @@
+#!/bin/bash
+
+check_buildx() {
+  export DOCKER_CLI_EXPERIMENTAL=enabled
+
+  # If there is no buildx let's install it
+  if ! docker buildx > /dev/null 2>&1; then
+     mkdir -p ~/.docker/cli-plugins
+     curl -L https://github.com/docker/buildx/releases/download/v0.6.3/buildx-v0.6.3.linux-amd64 --output ~/.docker/cli-plugins/docker-buildx
```
0xFelix, line 9: Hardcoded buildx version?

hack/build-multiarch-operator.sh
0xFelix, line 1: Did you run shellcheck against this file?

0xFelix (commented):
@oshoval Not to block the PR, but IMO it adds a lot of complexity which I'm not sure is required. (especially looking at `hack/build-multiarch-operator.sh`)

oshoval:
> @oshoval Not to block the PR, but IMO it adds a lot of complexity which I'm not sure is required. (especially looking at `hack/build-multiarch-operator.sh`)

we can indeed request to have a separate script for it instead having it internal as it is now, wdyt?

Makefile
ashokpariya0: It was needed, but removed it latest commit and moved to shell script.

Makefile
ashokpariya0: Updated in latest commit.

hack/build-multiarch-operator.sh
ashokpariya0: good catch, corrected it.

hack/build-multiarch-operator.sh
ashokpariya0: Yes.

ashokpariya0:
@oshoval @0xFelix The review comments have been addressed. Please let me know if there's anything else I need to do.

build/operator/Dockerfile
```diff
@@ -1,4 +1,25 @@
...
+    curl -L "https://go.dev/dl/${GO_VERSION}.linux-${ARCH}.tar.gz" -o go.tar.gz && \
+    tar -C /usr/local -xzf go.tar.gz && \
+    rm go.tar.gz
+ENV PATH=$PATH:/usr/local/go/bin
+WORKDIR /go/src/cluster-network-addons-operator
+COPY . .
+
+ARG TARGETOS
+ARG TARGETARCH
+ARG TARGETPLATFORM
```
0xFelix, line 16: Is this arg used?

hack/build-operator-docker.sh
```diff
@@ -0,0 +1,23 @@
...
+
+if [ ${#PLATFORM_LIST[@]} -eq 1 ]; then
+    # Single platform build
+    docker build --platform "$PLATFORMS" $BUILD_ARGS
+else
+    # Multi-platform build
+    ./hack/init-buildx.sh "$DOCKER_BUILDER"
+    docker buildx build --platform "$PLATFORMS" $BUILD_ARGS
+fi
+
```
0xFelix, line 21: Empty lines at EOF

build/operator/Dockerfile
```diff
@@ -1,4 +1,25 @@
...
+    GO_VERSION=$(curl -L -s "https://go.dev/dl/?mode=json" | jq -r '.[0].version') && \
+    curl -L "https://go.dev/dl/${GO_VERSION}.linux-${ARCH}.tar.gz" -o go.tar.gz && \
+    tar -C /usr/local -xzf go.tar.gz && \
+    rm go.tar.gz
+ENV PATH=$PATH:/usr/local/go/bin
+WORKDIR /go/src/cluster-network-addons-operator
+COPY . .
+
+ARG TARGETOS
+ARG TARGETARCH
```
0xFelix, line 15: Where are these set?

build/operator/Dockerfile
ashokpariya0: Yes

hack/build-operator-docker.sh
ashokpariya0: Removed.

build/operator/Dockerfile
ashokpariya0: TARGETOS and TARGETARCH are set by docker/podman based on the platform you're building for based on (--platform ) flag.

ashokpariya0:
@oshoval @0xFelix  I’ve addressed all the review comments.
Could you please take another look and proceed with the next steps for the PR?
Let me know if there’s anything else you’d like me to update.

Makefile
0xFelix: @ashokpariya0 Can you comment on this?

build/operator/Dockerfile
0xFelix: I may be blind, but where?

Makefile
ashokpariya0: Yes, We may need to modify the section at Makefile lines 113-126(https://github.com/kubevirt/cluster-network-addons-operator/blob/main/Makefile#L113-L126), but this can be addressed in a separate PR, as renaming might impact some build processes or jobs. The existing code already supports both Docker and Podman, as indicated in [Makefile line 14](https://github.com/kubevirt/cluster-network-addons-operator/blob/main/Makefile#L14).

build/operator/Dockerfile
ashokpariya0: we use at https://github.com/kubevirt/cluster-network-addons-operator/blob/106c2b661dff49532a0e2b5738f7c12be41a48f2/build/operator/Dockerfile#L18-L19

ashokpariya0: TARGETPLATFORM can be removed.

Makefile
0xFelix: With your change this target will only push in case `OCI_BIN` is equal to `podman`, IIUC?

ashokpariya0: Yes exactly, with docker buildx, push is combined with image build.

0xFelix: Fine for me, @oshoval wdyt?

oshoval: fine by me to address naming on follow-up
thx

0xFelix (approved):
/lgtm

build/operator/Dockerfile
```diff
@@ -1,4 +1,24 @@
-FROM quay.io/centos/centos:stream9
+ARG BUILD_ARCH=amd64
+FROM --platform=linux/${BUILD_ARCH} quay.io/centos/centos:stream9 AS builder
+
+RUN dnf install -y tar gzip jq && dnf clean all
+RUN ARCH=$(uname -m | sed 's/x86_64/amd64/') && \
+    GO_VERSION=$(curl -L -s "https://go.dev/dl/?mode=json" | jq -r '.[0].version') && \
```
oshoval, line 6: i dont think we want unpinned go version?
stuff can break suddenly

unless we are fine with that or if this was the case already, lets see other maintainers idea

hack/build-operator-docker.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+if [ -z "$ARCH" ] || [ -z "$PLATFORMS" ] || [ -z "$OPERATOR_IMAGE_TAGGED" ]; then
+    echo "Error: ARCH, PLATFORMS and OPERATOR_IMAGE_TAGGED must be set."
+    exit 1
+fi
+
+# Split the comma-separated platforms into an array
```
oshoval, line 8: please drop comments
from all files, unless non trivial

oshoval (commented):
quick pass

oshoval:
please see pull-cluster-network-addons-operator-unit-test

ashokpariya0:
> please see pull-cluster-network-addons-operator-unit-test
>
> Sure.

build/operator/Dockerfile
ashokpariya0: This was already the case; this PR follows the same approach.

Makefile
ashokpariya0: @oshoval  I've already looked into this and found that `docker-*` is being called from various scripts in the automation, hack folder, etc. Additionally, we need to update the README and other related files. Including all of this in a single commit would add unnecessary complexity. It would be better to handle these changes separately.

oshoval: ok thanks

oshoval:
Thanks

@qinqon 
any thing you want to add please ?

Makefile
ashokpariya0: @oshoval I will create an separate issue for this, resolve it, and provide a follow-up.

build/operator/Dockerfile
ashokpariya0: @oshoval For this as well, I will create a separate issue, resolve it, and provide a follow-up.

oshoval: thanks, if this was already the case, no rush to fix it, its ok

ashokpariya0:
Could we please have a release once this PR is merged? There were commits made after the last release (v0.96.0) on October 8th. We would like to perform end-to-end testing on this release.
Also, as discussed earlier, please set PLATFORMS=linux/amd64,linux/s390x before pushing the upstream image, as by default it will only publish amd64-compatible images.

oshoval:
yes thanks but at this very moment i cant work on that
once one of us have more time

ashokpariya0:
> yes thanks but at this very moment i cant work on that once one of us have more time
> 
> you are welcome if you want to check project-infra and suggest a PR
>
> Okay, thanks! For now, merging the PR once it passes the tests would be helpful.

oshoval:
@RamLavi 
if you want to also take a look, before we approve soon

ashokpariya0:
@RamLavi Would like to add anything ??
@oshoval Could you please approve the changes if no further actions are required, so that it can trigger the other tests?

build/operator/Dockerfile
oshoval: Please consider in the follow-up to use ADD if it helps

oshoval:
Very nice thanks
/approve

ashokpariya0:
@oshoval Thanks for approving the PR! How can we trigger all the tests for it?

oshoval:
/test all

thought it auto triggers once lgtm approve

ashokpariya0:
Thanks @oshoval  all tests have passed. There's one pending test job (tide) that requires the LGTM label.

oshoval:
Hi @0xFelix 
Can you please review / lgtm once you are good with it ?
Thanks

0xFelix (approved):
/lgtm

Thanks!

oshoval:
Hi
Please check this error before we advance on any PR

https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_cluster-network-addons-operator/1923/pull-cluster-network-addons-operator-unit-test/1860989134017925120#1:build-log.txt%3A10836

also i dont understand how come some lanes seems they didn't run
I will create a vanilla main PR to see CNAO state

oshoval:
I think it will be the best to revert this PR and recreate it (due to the CI problems above)
I will also need time to update my own machine to see it works locally

hack/build-operator-podman.sh
```diff
@@ -0,0 +1,24 @@
...
+
+# Remove any existing manifest and image
+podman manifest rm "${OPERATOR_IMAGE_TAGGED}" || true
+podman rmi "${OPERATOR_IMAGE_TAGGED}" || true
+
+podman manifest create "${OPERATOR_IMAGE_TAGGED}"
+
+for platform in "${PLATFORM_LIST[@]}"; do
+    podman build \
+        --no-cache \
```
oshoval, line 18: please drop the no-cache on a follow-up

hack/build-operator-podman.sh
```diff
@@ -0,0 +1,24 @@
...
+if [ -z "$ARCH" ] || [ -z "$PLATFORMS" ] || [ -z "$OPERATOR_IMAGE_TAGGED" ]; then
+    echo "Error: ARCH, PLATFORMS, and OPERATOR_IMAGE_TAGGED must be set."
+    exit 1
+fi
+
+IFS=',' read -r -a PLATFORM_LIST <<< "$PLATFORMS"
+
+# Remove any existing manifest and image
+podman manifest rm "${OPERATOR_IMAGE_TAGGED}" || true
+podman rmi "${OPERATOR_IMAGE_TAGGED}" || true
```
oshoval, line 12: please consider adding redir of stderr to /dev/null so it will clean up the logs from marking error on CI

hack/build-operator-podman.sh
ashokpariya0: Done.

hack/build-operator-podman.sh
ashokpariya0: Added in PR: https://github.com/kubevirt/cluster-network-addons-operator/pull/1928

hack/build-operator-podman.sh
oshoval: Thanks

PR #1767, opened 2024-02-06
Update linux-bridge component

ormergi:
**What this PR does / why we need it**:
Update the linux-bridge component to latest in order to bridge the latest changes of bridge CNI that include the new `disableContainerInterface` option https://github.com/containernetworking/plugins/pull/997 

**Special notes for your reviewer**:
Since there is no tagged version of container network plugins yet, consume that latest version from main branch  https://github.com/containernetworking/plugins/commit/14bdce598f9d332303c375c35719c4a158f1e7db

**Release note**:

```release-note
NONE
```

PR #1927, opened 2024-11-28
Enable support for multiple platforms in the Linux Bridge CNI.

ashokpariya0:
These updates allow the building and pushing of Linux Bridge container images for multiple platforms (e.g., amd64, s390x) using a single Dockerfile. Multi-platform build support is provided for both Docker and Podman container runtimes.

"The upstream **cni-default-plugins** image ([[link](https://quay.io/repository/kubevirt/cni-default-plugins?tab=tags&tag=latest)](https://quay.io/repository/kubevirt/cni-default-plugins?tab=tags&tag=latest)) currently supports only the amd64 architecture. The proposed changes in this PR enable multi-platform support for the **linux-bridge** image, broadening its compatibility to include additional architectures like s390x."

Multi-platform build support is enabled for both Docker and Podman container runtimes, and the necessary changes have been made accordingly.

To use this feature, you need to:
For Podman:
No additional dependencies are required to build multi-platform images. Podman supports this feature natively.
and We set the default value of OCI_BIN to podman if it's available, so we are good here.

For Docker:
Have access to docker buildx. More information: [[Docker Buildx Documentation](https://docs.docker.com/build/building/multi-platform/)](https://docs.docker.com/build/building/multi-platform/)
Ensure that BuildKit is enabled. More information: [[Docker BuildKit Enhancements](https://docs.docker.com/develop/develop-images/build_enhancements/)](https://docs.docker.com/develop/develop-images/build_enhancements/)

**Release note**:

```release-note
None
```

### DIFF
--- a/hack/components/bump-linux-bridge.sh
+++ b/hack/components/bump-linux-bridge.sh
@@ -7,7 +7,6 @@ source hack/components/git-utils.sh
 source hack/components/docker-utils.sh
 
 export OCI_BIN=${OCI_BIN:-$(docker-utils::determine_cri_bin)}
-
 echo Bumping linux-bridge
 LINUX_BRIDGE_URL=$(yaml-utils::get_component_url linux-bridge)
 LINUX_BRIDGE_COMMIT=$(yaml-utils::get_component_commit linux-bridge)
@@ -27,19 +26,33 @@ echo 'Build container image with linux-bridge binaries'
 LINUX_BRIDGE_TAR_CONTAINER_DIR=/usr/src/github.com/containernetworking/plugins/bin
 LINUX_BRIDGE_IMAGE=quay.io/kubevirt/cni-default-plugins
 LINUX_BRIDGE_IMAGE_TAGGED=${LINUX_BRIDGE_IMAGE}:${LINUX_BRIDGE_TAG}
-(
-    cd ${LINUX_BRIDGE_PATH}
+DOCKER_BUILDER="${DOCKER_BUILDER:-linux-bridge-docker-builder}"
+# By default, the build will target all supported platforms(i.e PLATFORM_LIST).
+# To build for specific platforms, you can:
+# 1. Specify individual platforms:
+#    export PLATFORMS=linux/amd64
+#    or
+#    export PLATFORMS=linux/amd64,linux/arm64
+PLATFORM_LIST="linux/amd64,linux/s390x,linux/arm64"
+ARCH=$(uname -m | sed 's/x86_64/amd64/;s/aarch64/arm64/')
+PLATFORMS="${PLATFORMS:-all}"
+[ "$PLATFORMS" == "all" ] && PLATFORMS="${PLATFORM_LIST}"
+IFS=',' read -r -a PLATFORM_LIST <<< "$PLATFORMS"
+
+create_dockerfile() {
     cat <<EOF > Dockerfile
-FROM registry.access.redhat.com/ubi8/ubi-minimal AS builder
+FROM --platform=\$BUILDPLATFORM registry.access.redhat.com/ubi8/ubi-minimal AS builder
+ARG TARGETOS
+ARG TARGETARCH
 RUN microdnf install -y golang git
 RUN \
     git clone https://${LINUX_BRIDGE_REPO} ${LINUX_BRIDGE_PATH} && \
     cd ${LINUX_BRIDGE_PATH} && \
     git checkout ${LINUX_BRIDGE_TAG}
 WORKDIR ${LINUX_BRIDGE_PATH}
-RUN GOFLAGS=-mod=vendor ./build_linux.sh
+RUN GOFLAGS=-mod=vendor GOARCH=\${TARGETARCH} GOOS=\${TARGETOS} ./build_linux.sh
 
-FROM registry.access.redhat.com/ubi8/ubi-minimal
+FROM --platform=linux/\$TARGETARCH registry.access.redhat.com/ubi8/ubi-minimal
 LABEL org.opencontainers.image.authors="phoracek@redhat.com"
 ENV SOURCE_DIR=${REMOTE_SOURCE_DIR}/app
 RUN mkdir -p ${LINUX_BRIDGE_TAR_CONTAINER_DIR}
@@ -49,14 +62,60 @@ COPY --from=builder ${LINUX_BRIDGE_PATH}/bin/tuning ${LINUX_BRIDGE_TAR_CONTAINER
 RUN sha256sum ${LINUX_BRIDGE_TAR_CONTAINER_DIR}/bridge >${LINUX_BRIDGE_TAR_CONTAINER_DIR}/bridge.checksum
 RUN sha256sum ${LINUX_BRIDGE_TAR_CONTAINER_DIR}/tuning >${LINUX_BRIDGE_TAR_CONTAINER_DIR}/tuning.checksum
 EOF
-    ${OCI_BIN} build -t ${LINUX_BRIDGE_IMAGE_TAGGED} .
-)
+}
 
-echo 'Push the image to KubeVirt repo'
-(
-    if [ ! -z ${PUSH_IMAGES} ]; then
-        ${OCI_BIN} push "${LINUX_BRIDGE_IMAGE_TAGGED}"
+check_and_create_docker_builder() {
+    existing_builder=$(docker buildx ls | grep -w "$DOCKER_BUILDER" | awk '{print $1}' || true)
+    if [ -n "$existing_builder" ]; then
+        echo "Builder '$DOCKER_BUILDER' already exists. Using existing builder."
+        docker buildx use "$DOCKER_BUILDER"
+    else
+        echo "Creating a new Docker Buildx builder: $DOCKER_BUILDER"
+        docker buildx create --driver-opt network=host --use --name "$DOCKER_BUILDER"
     fi
+}
+
+build_docker_image() {
+    docker buildx build --platform "${PLATFORMS}" -t "${LINUX_BRIDGE_IMAGE_TAGGED}" . --push
+    docker buildx rm "$DOCKER_BUILDER"
+}
+
+build_podman_image() {
+    podman manifest rm "${LINUX_BRIDGE_IMAGE_TAGGED}" 2>/dev/null || true
+    podman rmi "${LINUX_BRIDGE_IMAGE_TAGGED}" 2>/dev/null || true
+    podman rmi $(podman images --filter "dangling=true" -q) 2>/dev/null || true
+    podman manifest create "${LINUX_BRIDGE_IMAGE_TAGGED}"
+
+    for platform in "${PLATFORM_LIST[@]}"; do
+        podman build --platform "$platform" --manifest "${LINUX_BRIDGE_IMAGE_TAGGED}" .
+    done
+}
+
+push_image_to_kubevirt_repo() {
+    echo 'Push the image to KubeVirt repo'
+    if [ "${OCI_BIN}" == "podman" ]; then
+        if [ ! -z "${PUSH_IMAGES}" ]; then
+            podman manifest push "${LINUX_BRIDGE_IMAGE_TAGGED}"
+        fi
+    fi
+}
+
+
+(
+    cd ${LINUX_BRIDGE_PATH}
+    create_dockerfile
+    (
+        if [[ "${OCI_BIN}" == "docker" ]]; then
+            check_and_create_docker_builder
+            build_docker_image
+        elif [[ "${OCI_BIN}" == "podman" ]]; then
+            build_podman_image
+            push_image_to_kubevirt_repo
+        else
+            echo "Invalid OCI_BIN value. It must be either 'docker' or 'podman'."
+            exit 1
+        fi
+    )
 )
 
 if [[ -n "$(docker-utils::check_image_exists "${LINUX_BRIDGE_IMAGE}" "${LINUX_BRIDGE_TAG}")" ]]; then
--- a/hack/build-operator-podman.sh
+++ b/hack/build-operator-podman.sh
@@ -10,6 +10,7 @@ IFS=',' read -r -a PLATFORM_LIST <<< "$PLATFORMS"
 # Remove any existing manifest and image
 podman manifest rm "${OPERATOR_IMAGE_TAGGED}" 2>/dev/null || true
 podman rmi "${OPERATOR_IMAGE_TAGGED}" 2>/dev/null || true
+podman rmi $(podman images --filter "dangling=true" -q) 2>/dev/null || true
 
 podman manifest create "${OPERATOR_IMAGE_TAGGED}"
 


hack/components/bump-linux-bridge.sh
```diff
@@ -27,10 +27,17 @@ echo 'Build container image with linux-bridge binaries'
 LINUX_BRIDGE_TAR_CONTAINER_DIR=/usr/src/github.com/containernetworking/plugins/bin
 LINUX_BRIDGE_IMAGE=quay.io/kubevirt/cni-default-plugins
 LINUX_BRIDGE_IMAGE_TAGGED=${LINUX_BRIDGE_IMAGE}:${LINUX_BRIDGE_TAG}
-(
-    cd ${LINUX_BRIDGE_PATH}
+ARCH=$(uname -m | sed 's/x86_64/amd64/')
+PLATFORMS="linux/amd64,linux/s390x"
```
oshoval, line 31: please add auto detection instead hardcoding both, we should be able to compile as fast as possible

hack/components/bump-linux-bridge.sh
```diff
@@ -39,7 +46,7 @@ RUN \
 WORKDIR ${LINUX_BRIDGE_PATH}
 RUN GOFLAGS=-mod=vendor ./build_linux.sh
 
-FROM registry.access.redhat.com/ubi8/ubi-minimal
+FROM registry.access.redhat.com/ubi8/ubi-minimal AS final
```
oshoval, line 49: please drop AS final, it is not used
prone theoretically for mistakes if someone adds layers below that

hack/components/bump-linux-bridge.sh
```diff
@@ -49,13 +56,65 @@ COPY --from=builder ${LINUX_BRIDGE_PATH}/bin/tuning ${LINUX_BRIDGE_TAR_CONTAINER
...
+    docker buildx rm "$DOCKER_BUILDER"
+}
+
+build_podman_image() {
+    podman manifest rm "${LINUX_BRIDGE_IMAGE_TAGGED}" || true
+    podman rmi "${LINUX_BRIDGE_IMAGE_TAGGED}" || true
+    podman manifest create "${LINUX_BRIDGE_IMAGE_TAGGED}"
+
+    for platform in "${PLATFORM_LIST[@]}"; do
+        podman build --no-cache --build-arg BUILD_ARCH="$ARCH" --platform "$platform" --manifest "${LINUX_BRIDGE_IMAGE_TAGGED}" .
```
oshoval, line 83: why no-cache ?
we do want cache locally (fast build is important)

hack/components/bump-linux-bridge.sh
```diff
@@ -49,13 +56,65 @@ COPY --from=builder ${LINUX_BRIDGE_PATH}/bin/tuning ${LINUX_BRIDGE_TAR_CONTAINER
...
+    done
+}
+
+modify_dockerfile_for_platform_and_architecture() {
+    local dockerfile="$1"
+    # Modify Dockerfile to set platform and architecture
+    sed -i 's|^FROM registry.access.redhat.com/ubi8/ubi-minimal AS builder$|FROM --platform=linux/${BUILD_ARCH} registry.access.redhat.com/ubi8/ubi-minimal AS builder|' "$dockerfile"
+    sed -i 's|RUN GOFLAGS=-mod=vendor ./build_linux.sh|RUN GOFLAGS=-mod=vendor GOARCH=${TARGETARCH} GOOS=${TARGETOS} ./build_linux.sh|' "$dockerfile"
+    sed -i 's/^FROM registry.access.redhat.com\/ubi8\/ubi-minimal AS final$/FROM --platform=linux\/${TARGETARCH} registry.access.redhat.com\/ubi8\/ubi-minimal AS final/' "$dockerfile"
+}
```
oshoval, line 93: i dont understand why we sed -i instead directly building the dockerfile as needed please
after all the file is built in `create_dockerfile` right?
lets refactor this please

hack/components/bump-linux-bridge.sh
```diff
@@ -49,13 +56,65 @@ COPY --from=builder ${LINUX_BRIDGE_PATH}/bin/tuning ${LINUX_BRIDGE_TAR_CONTAINER
...
 )
 
 echo 'Push the image to KubeVirt repo'
 (
-    if [ ! -z ${PUSH_IMAGES} ]; then
-        ${OCI_BIN} push "${LINUX_BRIDGE_IMAGE_TAGGED}"
+    if [ "${OCI_BIN}" == "podman" ]; then
+        if [ ! -z ${PUSH_IMAGES} ]; then
+            podman manifest push "${LINUX_BRIDGE_IMAGE_TAGGED}"
+        fi
```
oshoval, line 117: please move it after `build_podman_image`

oshoval (commented):
Thanks, some comments

hack/components/bump-linux-bridge.sh
ashokpariya0: Done.

hack/components/bump-linux-bridge.sh
ashokpariya0: Done.

hack/components/bump-linux-bridge.sh
ashokpariya0: As discussed no-cache is removed to speed up build.

hack/components/bump-linux-bridge.sh
ashokpariya0: Done

hack/components/bump-linux-bridge.sh
ashokpariya0: We need to use `TARGETARCH` and `TARGETOS` in the Dockerfile, but these values are only available at runtime. Since the Dockerfile is generated using a `cat` command, directly inserting these variables during creation would result in empty values, leading to an incorrect Dockerfile. To address this, we generate the Dockerfile first and then update it using the `sed` command to properly set the values at runtime.

oshoval: Just out of curiosity please (no need to do it, because it means we need a new file)

Cant we do the same you did before, have those in a new actual Dockerfile (not one that is catted),
and supply them externally, exactly as in other places ?
```
ARG TARGETOS
ARG TARGETARCH
```

hack/components/bump-linux-bridge.sh
```diff
@@ -27,10 +27,23 @@ echo 'Build container image with linux-bridge binaries'
...
-(
-    cd ${LINUX_BRIDGE_PATH}
+DOCKER_BUILDER="${DOCKER_BUILDER:-linux-bridge-docker-builder}"
+# By default, the build will be based on the host architecture.
+# To build for other platforms, you can:
+# 1. Export all supported platforms: export PLATFORMS=all
+# 2. Or specify specific platforms: export PLATFORMS=linux/amd64,linux/arm64
+PLATFORM_LIST="linux/amd64,linux/s390x,linux/arm64"
+ARCH=$(uname -m | sed 's/x86_64/amd64/;s/aarch64/arm64/')
+[ -z "$PLATFORMS" ] && PLATFORMS="linux/${ARCH}"
```
oshoval, line 37: please use default value bash syntax, as in line 30

oshoval (commented):
Just a small comment, thanks

hack/components/bump-linux-bridge.sh
oshoval: I am a bit confused why it is different than all the other cases where we have a real Dockerfile (not catted)
with those as args, but we can take it offline, and unblock this PR, (because in any case, not sure we want a standalone Dockerfile for the bump / publish)

hack/components/bump-linux-bridge.sh
ashokpariya0: Done.

hack/components/bump-linux-bridge.sh
ashokpariya0: Fixed this, Removed sed commands and used backslash to avoid substitution while creating file for few specific lines with cat command.

oshoval:
Looks great thanks

/test all
/approve

@0xFelix 
can you please take a look ?

0xFelix (commented):
Can you run `shellcheck` on this file and fix the issues introduced by this PR?

oshoval:
> Can you run `shellcheck` on this file and fix the issues introduced by this PR?

Note that some the warnings are considered annoyance imho, so it is better to have judgement per warning please
(i even saw some delicate ones that can create bugs, but on more sophisticated scripts like we have on kubevirtci sriov)

ashokpariya0:
> Can you run `shellcheck` on this file and fix the issues introduced by this PR?
>
This PR does not introduce any new warnings; I have already verified it using shellcheck

oshoval:
Once this PR is merged, also need to recall / find out please how to trigger its release

ashokpariya0:
> ```
> In /home/felix/Downloads/bump-linux-bridge.sh line 95:
>         if [ ! -z "${PUSH_IMAGES}" ]; then
>              ^-- SC2236 (style): Use -n instead of ! -z.
> 
> 
> In /home/felix/Downloads/bump-linux-bridge.sh line 103:
>     cd ${LINUX_BRIDGE_PATH}
>        ^------------------^ SC2086 (info): Double quote to prevent globbing and word splitting.
> 
> Did you mean:
>     cd "${LINUX_BRIDGE_PATH}"
> ```
> 
> These are relevant.
>



> The 2nd isnt regression of this PR, as long as we don't afraid from word splitting / globbing in our values i prefer to not to enforce it please.
> 
> I remember something vaguely about -n when not using quotes, if the value is empty, it can have undesired output, but value shouldn't be empty here. If needed I can find it.
> 
> More over, the 1st isn't regression of this PR, so imo we don't need to enforce fixing it on this PR (and it is syntactic equivalent)
>
@oshoval I also encountered some issues recently when trying to achieve a 100% result in ShellCheck without any warnings.
Since none of the warnings above were introduced by this PR, can we proceed with it? I can submit a follow-up PR if needed.  @0xFelix

oshoval:
imo we can proceed please,
and even we don't need a follow-up about it from my side
thanks

0xFelix:
I'm not a fan of leaving warnings in place when you change the code anyway. But it up to you to decide.

ashokpariya0:
@0xFelix @oshoval  Is there anything else on your end? If not, could we proceed with merging this PR?

ashokpariya0:
> hack/components/bump-linux-bridge.sh

> btw once this PR is merged Looking on #1767 I think you can just change to "tagged" and the auto bump will do the rest since there is a new tag https://github.com/containernetworking/plugins/releases/tag/v1.6.1 Best if you can verify please with Or Mergi this is correct
> 
I created issue https://github.com/kubevirt/cluster-network-addons-operator/issues/1937 to discuss above.
> you will need to include `PLATFORMS=all` of course in that flow somehow maybe you can set it as default and thats it for simplicity, in case of problem we can rethink
>
Do you mean that we need to set the default PLATFORMS as all in the script hack/components/bump-linux-bridge.sh
?

oshoval:
@0xFelix can you please re-review ?

oshoval:
> I created issue #1937 to discuss above.
> 
> > you will need to include `PLATFORMS=all` of course in that flow somehow maybe you can set it as default and thats it for simplicity, in case of problem we can rethink
> 
> Do you mean that we need to set the default PLATFORMS as all in the script hack/components/bump-linux-bridge.sh ?

Thanks
Either that, or the script need to be smart that if it runs on via git actions it enables it, would you able please to check about it ? if it is complex, we can just set it hardcoded as long as it can still run locally easily

ashokpariya0:
> > > you will need to include `PLATFORMS=all` of course in that flow somehow maybe you can set it as default and thats it for simplicity, in case of problem we can rethink
> > 
> > 
> > Do you mean that we need to set the default PLATFORMS as all in the script hack/components/bump-linux-bridge.sh ?
> 
> Thanks Either that, or the script need to be smart that if it runs on via git actions it enables it, would you able please to check about it ? if it is complex, we can just set it hardcoded as long as it can still run locally easily
>
Let me add this change to the PR itself. I don’t foresee any issues with running it locally, as we use Podman, and there are no additional dependencies required on Podman for the multi-platform image build.

oshoval:
ok, just to make sure, it doesnt break docker right?
we still do support docker unless we decide in the future to stop supporting it

ashokpariya0:
> ok, just to make sure, it doesnt break docker right? we still do support docker unless we decide in the future to stop supporting it
>
Correct, it will work with Docker as well. The only requirement for Docker is the use of docker buildx, while no additional dependencies are needed for Podman.

oshoval:
lgtm

oshoval:
This PR must be merged before the one that update the tag does, else it will be harder to trigger the auto bumper

ashokpariya0:
cc @phoracek

oshoval:
We need Felix review, no need to tag here Petr atm please

ashokpariya0:
@oshoval @0xFelix I added a Podman command to remove dangling images. Please refer to the explanation here:
https://github.com/k8snetworkplumbingwg/kubemacpool/pull/444#discussion_r1883313463

oshoval:
Thanks

if the dangling flag is crucial we might need to add it on all the other repos ?
anyhow no rush, in case there were more nits so we can do them at once

/test all

oshoval:
https://github.com/kubevirt/cluster-network-addons-operator/actions/runs/12310655279/job/34359580353?pr=1927
not related to this PR, seems kubevirt deployment didnt finish in time, first time i see it, so no biggie for now

oshoval:
@RamLavi can you please review ?

RamLavi (approved):
/lgtm

oshoval:
something is broken since latest pushes, please check, you can create a vanilla PR to see main does work if desired

oshoval:
/hold

so it wont be retested

ashokpariya0:
> something is broken since latest pushes, please check, you can create a vanilla PR to see main does work if desired

> I think PR: https://github.com/kubevirt/cluster-network-addons-operator/pull/1943 got merged 3hrs back, May be i need to rebase my pr

ashokpariya0:
> something is broken since latest pushes, please check, you can create a vanilla PR to see main does work if desired
>
 I think PR: #1943 got merged 3hrs back, May be i need to rebase my pr

ashokpariya0:
> > something is broken since latest pushes, please check, you can create a vanilla PR to see main does work if desired
> 
> I think PR: #1943 got merged 3hrs back, May be i need to rebase my pr
>
Done

oshoval:
seems https://github.com/kubevirt/cluster-network-addons-operator/actions/runs/12339360372/job/34435568203?pr=1927 is failing, would you have time to look please ?
it seems it is just because the packages it tries to uninstall arent there,
simple fix it to `|| true` with that
better is to check what https://github.com/ovn-kubernetes/ovn-kubernetes are doing on their lane, because they were the ref

ashokpariya0:
> https://github.com/ovn-kubernetes/ovn-kubernetes



> Thanks
> 
> seems https://github.com/kubevirt/cluster-network-addons-operator/actions/runs/12339360372/job/34435568203?pr=1927 is failing, would you have time to look please ? it seems it is just because the packages it tries to uninstall arent there, simple fix it to `|| true` with that better is to check what https://github.com/ovn-kubernetes/ovn-kubernetes are doing on their lane, because they were the ref
>
I don't see any change w.r.t to free disk space https://github.com/ovn-kubernetes/ovn-kubernetes/blob/eee91422cc619cde68dcb4e3be30ccf33e188867/.github/workflows/test.yml#L295

oshoval:
> > https://github.com/ovn-kubernetes/ovn-kubernetes
> 
> > Thanks
> > seems https://github.com/kubevirt/cluster-network-addons-operator/actions/runs/12339360372/job/34435568203?pr=1927 is failing, would you have time to look please ? it seems it is just because the packages it tries to uninstall arent there, simple fix it to `|| true` with that better is to check what https://github.com/ovn-kubernetes/ovn-kubernetes are doing on their lane, because they were the ref
> 
> I don't see any change w.r.t to free disk space https://github.com/ovn-kubernetes/ovn-kubernetes/blob/eee91422cc619cde68dcb4e3be30ccf33e188867/.github/workflows/test.yml#L295

i believe git actions default container was changed and it will fail there soon as well if we use the same

oshoval:
unit test fail on kmp image mismatch (might be related to the last PR that was merged)
https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_cluster-network-addons-operator/1927/pull-cluster-network-addons-operator-unit-test/1868297019211124736

maybe @RamLavi  knows something about it ?
```
diff --git a/test/releases/99.0.0.go b/test/releases/99.0.0.go
index b3c8ca32..2784c3fa 100644
--- a/test/releases/99.0.0.go
+++ b/test/releases/99.0.0.go
@@ -42,7 +42,7 @@ func init() {
 				ParentName: "kubemacpool-mac-controller-manager",
 				ParentKind: "Deployment",
 				Name:       "manager",
-				Image:      "quay.io/kubevirt/kubemacpool@sha256:eebb65b8a12cbfc20a429bbba399eb5a5c2279f8613c36965957ee7c36cfcbd6",
+				Image:      "quay.io/kubevirt/kubemacpool@sha256:15c7be1e7aad2836237e9a73443192bfa58c044fcc5488b175222f3ae997064f",
 			},
 			{
 				ParentName: "kubemacpool-mac-controller-manager",
@@ -54,7 +54,7 @@ func init() {
 				ParentName: "kubemacpool-cert-manager",
 				ParentKind: "Deployment",
 				Name:       "manager",
-				Image:      "quay.io/kubevirt/kubemacpool@sha256:eebb65b8a12cbfc20a429bbba399eb5a5c2279f8613c36965957ee7c36cfcbd6",
+				Image:      "quay.io/kubevirt/kubemacpool@sha256:15c7be1e7aad2836237e9a73443192bfa58c044fcc5488b175222f3ae997064f",
 			},
 			{
 				ParentName: "ovs-cni-amd64",
+ exit 1
make: *** [Makefile:101: check] Error 1
```

oshoval:
open https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4914 and the one that is mentioned there
ipam-ext one (which has the same code as here) also fails for the same
lets see if OVN as well

oshoval:
still fails after rebase, unit test should be fixed first, and worth to check if master state is broken 
after the KMP bump

oshoval:
opened for that https://github.com/kubevirt/cluster-network-addons-operator/pull/1944

oshoval:
storage problem also hit OVN
https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4914

ashokpariya0:
> opened [ovn-kubernetes/ovn-kubernetes#4914](https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4914) and the one that is mentioned there ipam-ext one (which has the same code as here) also fails for the same lets see if OVN as well
>
The issue also occurs with OVN, as it seems that ubuntu-latest now points to Ubuntu 24.04. This is causing disk cleanup problems across the board. We can address this by either removing only the installed packages or switching to ubuntu-22.04.

Additionally, the following link still doesn't reflect the correct version:

https://github.com/actions/runner-images

Please also review the issue here: https://github.com/actions/runner-images/issues/10636.

oshoval:
Thank you
best to sync please with OVN people with your findings on https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4914 and choose the right direction
I will post that PR on slack once you post your findings there

ashokpariya0:
> Thank you best to sync please with OVN people with your findings on [ovn-kubernetes/ovn-kubernetes#4914](https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4914) and choose the right direction I will post that PR on slack once you post your findings there (or can just link your message, what ever you prefer)
>
Sure

jommeke22f:
> Deze updates maken het mogelijk om Linux Zeecontainers te bouwen en te pushen voor meerdere platformen (bijv. amd64, s390x) met behulp van één Docker file. Ondersteuning voor multiplatform builds wordt geboden voor zowel Docker- als Persconferenties.
> 
> **Wat deze PR doet / waarom we het nodig hebben** :
> 
> "De upstream **calculations** image ([ [link](https://quay.io/repository/kubevirt/cni-default-plugins?tab=tags&tag=latest) ]( https://quay.io/repository/kubevirt/cni-default-plugins?tab=tags&tag=latest )) ondersteunt momenteel alleen de architecture. De voorgestelde wijzigingen in deze PR maken cultuurverandering mogelijk voor de **linux-bridge** image, waardoor de compatibiliteit wordt uitgebreid met extra architecturen zoals s390x."
> 
> **Speciale opmerkingen voor uw reviewer** : Ondersteuning voor multi-platform builds is ingeschakeld voor zowel Docker- als Postmodernisme en de nodige wijzigingen zijn dienovereenkomstig doorgevoerd.
> 
> Om deze functie te gebruiken, moet u het volgende doen: Voor Polman: Er zijn geen extra afhankelijkheden vereist om multi-platform images te bouwen. Podia ondersteunt deze functie standaard. en We stellen de standaardwaarde van OCI_BIN in op podium als deze beschikbaar is, dus we zitten hier goed.
> 
> Voor Docker: Heb toegang tot docker build. Meer informatie: [ [Docker Documentatie](https://docs.docker.com/build/building/multi-platform/) ]( https://docs.docker.com/build/building/multi-platform/ ) Zorg ervoor dat Builds is ingeschakeld. Meer informatie: [ [Docker Kwaliteitsverbetering](https://docs.docker.com/develop/develop-images/build_enhancements/) ]( https://docs.docker.com/develop/develop-images/build_enhancements/ )
> 
> **Release-opmerking** :
> 
> ```
> None
> ```

ashokpariya0:
Before merging this PR, we need to first merge the following:

1. The PR that fixes the unit test issue with the Kubemacpool image: [#1946](https://github.com/kubevirt/cluster-network-addons-operator/pull/1946)
2. The PR that provides a workaround for the cleanup issue with the latest Ubuntu version: [#1945](https://github.com/kubevirt/cluster-network-addons-operator/pull/1945)

oshoval:
can you please rebase to take ipam fix ?

ashokpariya0:
> can you please rebase to take ipam fix ? we will also run the rest
>
I am on it.

ashokpariya0:
> can you please rebase to take ipam fix ? we will also run the rest
>
Done.

ashokpariya0:
@oshoval IPAM controller e2e tests passed on this pr, could you please trigger all other tests?

PR #1928, opened 2024-11-29
Enable dynamic host architecture detection.

ashokpariya0:
This update sets the platform dynamically based on the host architecture, enhancing flexibility for building multi-platform images across different architectures.

Instead of setting the platform value to `amd64`, these changes configure it dynamically based on the host architecture.

**Release note**:

```release-note
None
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,13 +13,13 @@ OPERATOR_IMAGE ?= cluster-network-addons-operator
 REGISTRY_IMAGE ?= cluster-network-addons-registry
 export OCI_BIN ?= $(shell if podman ps >/dev/null 2>&1; then echo podman; elif docker ps >/dev/null 2>&1; then echo docker; fi)
 TLS_SETTING := $(if $(filter $(OCI_BIN),podman),--tls-verify=false,)
-PLATFORMS ?= linux/amd64
+ARCH := $(shell uname -m | sed 's/x86_64/amd64/;s/aarch64/arm64/')
+PLATFORMS ?= linux/${ARCH}
 # Set the platforms for building a multi-platform supported image.
 # Example:
 # PLATFORMS ?= linux/amd64,linux/arm64,linux/s390x
 # Alternatively, you can export the PLATFORMS variable like this:
 # export PLATFORMS=linux/arm64,linux/s390x,linux/amd64
-ARCH := $(shell uname -m | sed 's/x86_64/amd64/')
 DOCKER_BUILDER ?= cnao-docker-builder
 OPERATOR_IMAGE_TAGGED := $(IMAGE_REGISTRY)/$(OPERATOR_IMAGE):$(IMAGE_TAG)
 


oshoval (approved):
Thanks
I assume it works for all the platforms mentioned on the comment

Makefile
```diff
@@ -13,13 +13,13 @@ OPERATOR_IMAGE ?= cluster-network-addons-operator
 REGISTRY_IMAGE ?= cluster-network-addons-registry
 export OCI_BIN ?= $(shell if podman ps >/dev/null 2>&1; then echo podman; elif docker ps >/dev/null 2>&1; then echo docker; fi)
 TLS_SETTING := $(if $(filter $(OCI_BIN),podman),--tls-verify=false,)
-PLATFORMS ?= linux/amd64
+ARCH := $(shell uname -m | sed 's/x86_64/amd64/;s/aarch64/arm64/')
+PLATFORMS ?= linux/${ARCH}
```
oshoval, line 17: wdyt about adding an option to set `export PLATFORMS=all` 
which will be translated internally to `linux/amd64,linux/arm64,linux/s390x` ?
it is less explicit but easier to maintain

ashokpariya0: @oshoval That's a great idea! Adding an `export PLATFORMS=all` option would simplify the configuration and make it easier to maintain, especially as we add more platforms in the future. I'll include this change in a follow-up PR.
All tests have passed, and there is one pending tide test awaiting the approval label.

oshoval:
cc @RamLavi @qinqon 
can you take a look please ?

Makefile
ashokpariya0: Updated PLATFORMS to automatically include all supported platforms when export PLATFORMS=all is used.

oshoval (commented):
Thanks

hack/build-operator-podman.sh
```diff
@@ -8,14 +8,13 @@ fi
 IFS=',' read -r -a PLATFORM_LIST <<< "$PLATFORMS"
 
 # Remove any existing manifest and image
-podman manifest rm "${OPERATOR_IMAGE_TAGGED}" || true
-podman rmi "${OPERATOR_IMAGE_TAGGED}" || true
+podman manifest rm "${OPERATOR_IMAGE_TAGGED}" 2>/dev/null || echo "Manifest ${OPERATOR_IMAGE_TAGGED} not found, skipping removal."
+podman rmi "${OPERATOR_IMAGE_TAGGED}" 2>/dev/null || echo "Image ${OPERATOR_IMAGE_TAGGED} not found, skipping removal."
```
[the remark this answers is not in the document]
ashokpariya0: Done.

hack/build-operator-podman.sh
```diff
@@ -8,14 +8,13 @@ fi
 IFS=',' read -r -a PLATFORM_LIST <<< "$PLATFORMS"
 
 # Remove any existing manifest and image
-podman manifest rm "${OPERATOR_IMAGE_TAGGED}" || true
-podman rmi "${OPERATOR_IMAGE_TAGGED}" || true
+podman manifest rm "${OPERATOR_IMAGE_TAGGED}" 2>/dev/null
+podman rmi "${OPERATOR_IMAGE_TAGGED}" 2>/dev/null
```
oshoval, line 12: you do need `|| true` please, isnt it ?

ashokpariya0: in anyways it will suppress the error but let me add || true.

hack/build-operator-docker.sh
```diff
@@ -14,4 +14,5 @@ if [ ${#PLATFORM_LIST[@]} -eq 1 ]; then
 else
     ./hack/init-buildx.sh "$DOCKER_BUILDER"
     docker buildx build --platform "$PLATFORMS" $BUILD_ARGS
+    docker buildx rm "$DOCKER_BUILDER" 2>/dev/null || echo "Builder ${DOCKER_BUILDER} not found or already removed, skipping."
```
oshoval, line 17: lets just use `2>/dev/null || true` please

hack/build-operator-podman.sh
oshoval: if it suppress the error in case image / manifest not found,  no need to add, my mistake then
thanks

hack/build-operator-docker.sh
ashokpariya0: Done.

hack/init-buildx.sh
```diff
@@ -20,7 +21,7 @@ create_or_use_buildx_builder() {
 
   check_buildx
 
-  current_builder="$(docker buildx inspect "${builder_name}")"
+  current_builder="$(docker buildx inspect "${builder_name}" 2>/dev/null)" || echo "Builder '${builder_name}' not found"
```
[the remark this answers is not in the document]
ashokpariya0: Done.

hack/build-operator-podman.sh
```diff
@@ -8,14 +8,13 @@ fi
 IFS=',' read -r -a PLATFORM_LIST <<< "$PLATFORMS"
 
 # Remove any existing manifest and image
-podman manifest rm "${OPERATOR_IMAGE_TAGGED}" || true
-podman rmi "${OPERATOR_IMAGE_TAGGED}" || true
+podman manifest rm "${OPERATOR_IMAGE_TAGGED}" 2>/dev/null || true
+podman rmi "${OPERATOR_IMAGE_TAGGED}" 2>/dev/null || true
```
oshoval, line 12: if podman suppress the error in case the images dont exists, please drop the `|| true` 

lets keep the code correct and as minimal as possible please
sorry for the back and forth

hack/init-buildx.sh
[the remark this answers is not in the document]
ashokpariya0: Yes, you're right; it will have an empty value. However, we handle errors properly later in the script. I don't see any issues with either the failure or success case, so everything looks good.

hack/build-operator-docker.sh
ashokpariya0: I think it's better to keep this message, as this is not an expected failure. It will help us with debugging in case an issue arises.

hack/build-operator-podman.sh
ashokpariya0: I think it's better to keep || true, Below is the reason:
Without || true: The command suppresses the error output but will still fail if there's an issue (e.g., manifest not found).
With || true: The command suppresses the error and ensures the script continues even if there's an issue.

hack/init-buildx.sh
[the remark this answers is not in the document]
oshoval: please correct me if i am wrong,
this is an expression that must success, hence we should fail fast (i.e remove the `|| echo ...`) ?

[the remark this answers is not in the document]
ashokpariya0: No, by default, `docker buildx inspect <builder-name>` will print an error to the console if the builder is not found. Since we want to suppress this error, it's better to use `||` with an echo statement for debugging purposes.

ashokpariya0:
@RamLavi @qinqon @phoracek 
Could you please take a look at the PR?

oshoval:
@0xFelix 
if you can review this one please, thanks

switching labels

/lgtm cancel
/approve

0xFelix (approved):
/lgtm

PR #1931, opened 2024-12-04
Use OpenShift's required-scc annotation on components

maiqueb:
This PR adds the `required-scc` annotation to the CNAO components.

It's an harmless annotation, and simplifies the code. Given the urgency of this, we do not try to add the annotation only when deploying on an OpenShift cluster; this should/could be done in a follow-up.

**Release note**:

```release-note
Adds the `openshift.io/required-scc` annotation to all CNAO components.
```

### DIFF
--- a/data/kubemacpool/kubemacpool.yaml
+++ b/data/kubemacpool/kubemacpool.yaml
@@ -139,6 +139,8 @@ spec:
     type: Recreate
   template:
     metadata:
+      annotations:
+        openshift.io/required-scc: restricted-v2
       labels:
         app: kubemacpool
         control-plane: cert-manager
@@ -228,6 +230,7 @@ spec:
     metadata:
       annotations:
         description: KubeMacPool manages MAC allocation to Pods and VMs
+        openshift.io/required-scc: restricted-v2
       labels:
         app: kubemacpool
         control-plane: mac-controller-manager
--- a/data/kubevirt-ipam-controller/001-kubevirtipamcontroller.yaml
+++ b/data/kubevirt-ipam-controller/001-kubevirtipamcontroller.yaml
@@ -172,6 +172,9 @@ spec:
     metadata:
       annotations:
         kubectl.kubernetes.io/default-container: manager
+{{ if .IsOpenshift }}
+        openshift.io/required-scc: "restricted-v2"
+{{ end }}
       labels:
         app: ipam-virt-workloads
         control-plane: manager
--- a/data/kubevirt-ipam-controller/003-passtbindingcni.yaml
+++ b/data/kubevirt-ipam-controller/003-passtbindingcni.yaml
@@ -23,6 +23,9 @@ spec:
         app: passt-binding-cni
       annotations:
         description: passt-binding-cni installs passt binding CNI on cluster nodes
+{{ if .EnableSCC }}
+        openshift.io/required-scc: "passt-binding-cni"
+{{ end }}
     spec:
       priorityClassName: system-cluster-critical
 {{ if .EnableSCC }}
--- a/hack/components/bump-kubevirt-ipam-controller.sh
+++ b/hack/components/bump-kubevirt-ipam-controller.sh
@@ -125,6 +125,10 @@ echo 'Adjust kubevirt-ipam-controller to CNAO'
     service.beta.openshift.io/serving-cert-secret-name: kubevirt-ipam-controller-webhook-service\
 {{ end }}' Service_kubevirt-ipam-controller-webhook-service.yaml
 
+  sed -i '/        kubectl.kubernetes.io\/default-container: manager/a\{{ if .IsOpenshift }}\
+        openshift.io/required-scc: "restricted-v2"\
+{{ end }}' Deployment_kubevirt-ipam-controller-manager.yaml
+
   echo 'rejoin sub-manifests to a final manifest'
   cat Namespace_kubevirt-ipam-controller-system.yaml \
       ServiceAccount_kubevirt-ipam-controller-manager.yaml \
@@ -143,6 +147,10 @@ echo 'Adjust kubevirt-ipam-controller to CNAO'
 
   sed -i '/containers:/i\{{ if .EnableSCC }}\
       serviceAccountName: passt-binding-cni\
+{{ end }}' 003-passtbindingcni.yaml
+
+  sed -i '/        description: passt-binding-cni installs passt binding CNI on cluster nodes/a\{{ if .EnableSCC }}\
+        openshift.io/required-scc: "passt-binding-cni"\
 {{ end }}' 003-passtbindingcni.yaml
 )
 
--- a/data/kube-secondary-dns/secondarydns.yaml
+++ b/data/kube-secondary-dns/secondarydns.yaml
@@ -73,6 +73,7 @@ spec:
         k8s-app: secondary-dns
       annotations:
         kubectl.kubernetes.io/default-container: status-monitor
+        openshift.io/required-scc: "restricted-v2"
     spec:
       serviceAccountName: secondary
       securityContext:
--- a/data/linux-bridge/002-linux-bridge.yaml
+++ b/data/linux-bridge/002-linux-bridge.yaml
@@ -22,6 +22,9 @@ spec:
         tier: node
         app: cni-plugins
       annotations:
+{{ if .EnableSCC }}
+        openshift.io/required-scc: "linux-bridge"
+{{ end }}
         description: LinuxBridge installs 'bridge' CNI on cluster nodes, so it can be later used to attach Pods/VMs to Linux bridges
     spec:
 {{ if .EnableSCC }}
--- a/data/linux-bridge/003-bridge-marker.yaml
+++ b/data/linux-bridge/003-bridge-marker.yaml
@@ -23,6 +23,7 @@ spec:
         name: bridge-marker
       annotations:
         description: Bridge marker exposes network bridges available on nodes as node resources
+        openshift.io/required-scc: "bridge-marker"
     spec:
       serviceAccountName: bridge-marker
       hostNetwork: true
--- a/hack/components/bump-bridge-marker.sh
+++ b/hack/components/bump-bridge-marker.sh
@@ -24,6 +24,7 @@ function __parametize_by_object() {
 				yaml-utils::update_param ${f} spec.template.spec.containers[0].imagePullPolicy  '{{ .ImagePullPolicy }}'
 				yaml-utils::update_param ${f} spec.template.spec.nodeSelector '{{ toYaml .Placement.NodeSelector | nindent 8 }}'
 				yaml-utils::set_param ${f} spec.template.spec.affinity '{{ toYaml .Placement.Affinity | nindent 8 }}'
+				yaml-utils::set_param ${f} 'spec.template.metadata.annotations."openshift.io/required-scc"' '"bridge-marker"'
 				yaml-utils::update_param ${f} spec.template.spec.tolerations '{{ toYaml .Placement.Tolerations | nindent 8 }}'
 				yaml-utils::remove_single_quotes_from_yaml ${f}
 				;;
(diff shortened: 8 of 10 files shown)


maiqueb:
@RamLavi I still need to patch up CNAO itself. Could you point me to the correct place ?

data/kubevirt-ipam-controller/001-kubevirtipamcontroller.yaml
```diff
@@ -170,8 +170,14 @@ spec:
       control-plane: manager
   template:
     metadata:
+{{ if .IsOpenshift }}
+      annotations:
+        openshift.io/required-scc: "restricted-v2"
+        kubectl.kubernetes.io/default-container: manager
```
wking, line 176: nit: why include this in the `.IsOpenshift` conditional?  I'd have expected:

```
      annotations:
        kubectl.kubernetes.io/default-container: manager
{{ if .IsOpenshift }}
        openshift.io/required-scc: "restricted-v2"
{{ end }}
```

data/kubemacpool/kubemacpool.yaml
```diff
@@ -143,6 +143,8 @@ spec:
         app: kubemacpool
         control-plane: cert-manager
         controller-tools.k8s.io: "1.0"
+      annotations:
+        openshift.io/required-scc: "restricted-v2"
```
oshoval, line 147: aint it should be with
`{{ if .IsOpenshift }}` or `{{ if .EnableSCC }}` ?

do we need to create this scc ? or it is one that is already created ?
(for example for bridge-marker we create the scc by CNAO upon needs)

data/kubevirt-ipam-controller/003-passtbindingcni.yaml
```diff
@@ -21,8 +21,14 @@ spec:
         name: passt-binding-cni
         tier: node
         app: passt-binding-cni
+{{ if .IsOpenshift }}
       annotations:
+        openshift.io/required-scc: "passt-binding-cni"
```
oshoval, line 26: do we need to create the scc?

data/kubevirt-ipam-controller/001-kubevirtipamcontroller.yaml
oshoval: agree,
also here please
https://github.com/kubevirt/cluster-network-addons-operator/pull/1931/files#diff-5cadb26d0d796a5eff24e582b27e36a926bcaa614a8df5fd54d3b52b4d3addd2R24

data/kubemacpool/kubemacpool.yaml
maiqueb: The script that templates this doesn't know any of those variables.

And kubemacpool - for whatever reason - already always has OpenShift related data set: https://github.com/kubevirt/cluster-network-addons-operator/blob/d5143952e14f7094a1a674ff7d6950aa4e738f37/data/kubemacpool/kubemacpool.yaml#L359

data/kubevirt-ipam-controller/003-passtbindingcni.yaml
maiqueb: yes.

data/kubevirt-ipam-controller/001-kubevirtipamcontroller.yaml
maiqueb: I guess it's possible, I'll try.

data/kubemacpool/kubemacpool.yaml
oshoval: if needed we can add to the script, i can show you if you want where

maiqueb: why is this an issue now, when it wasn't before ?

oshoval:
```
    - key: openshift.io/run-level
      operator: NotIn
      values:
      - "0"
      - "1"
```
for me this seems to have less effect then creating scc
wdty?
i am totally not into being more restrictive by this PR, but imho this might not be the same case here

maiqueb: I'm not creating an SCC here, I'm saying that this pod can only use this SCC. 

On a platform that doesn't use SCCs - e.g. vanilla kubernetes - this is just one more annotation. There's no handler for it.

hack/components/bump-kubevirt-ipam-controller.sh
```diff
@@ -125,6 +125,10 @@ echo 'Adjust kubevirt-ipam-controller to CNAO'
     service.beta.openshift.io/serving-cert-secret-name: kubevirt-ipam-controller-webhook-service\
 {{ end }}' Service_kubevirt-ipam-controller-webhook-service.yaml
 
+  sed -i '/        kubectl.kubernetes.io\/default-container: manager/a\{{ if .IsOpenshift }}\
```
oshoval, line 128: better imo to have please the key as something fixed like "annotations"
not "kubectl.kubernetes.io\/default-container: manager" which can change

(same for the rest when applicable)

data/kubevirt-ipam-controller/003-passtbindingcni.yaml
oshoval: if so, the PR isnt final yet iiuc

data/kubemacpool/kubemacpool.yaml
oshoval: btw, if KMP annotation can always be in the manifest, and no scc need to be created for this specific name
(iiuc it is a system one), then it should be better to not annotate it via CNAO, but add it directly on KMP repo please

data/kubevirt-ipam-controller/003-passtbindingcni.yaml
maiqueb: ?

hack/components/bump-kubevirt-ipam-controller.sh
maiqueb: the deployment also has annotations. I need to put the annotation on the pod template, not on all annotations in the file.

I agree that using sed for this is wrong, we should template this properly using yaml.

data/kubevirt-ipam-controller/003-passtbindingcni.yaml
oshoval: meant that the required SCC might need to be added unless they exists already in the manifests
i see for example this one exists, if all the other already exists / system created we are good
can you please spin it on CRC / kcli to see we are all set please?

```
kind: SecurityContextConstraints
metadata:
  name: passt-binding-cni
```

hack/components/bump-kubevirt-ipam-controller.sh
oshoval: you are right
forget about non sed for now please
you did correct

data/kubevirt-ipam-controller/003-passtbindingcni.yaml
maiqueb: @oshoval it's working:

```sh
oc version
Client Version: 4.17.7
Kustomize Version: v5.0.4-0.20230601165947-6ce0bf390ce3
Server Version: 4.17.7
Kubernetes Version: v1.30.6

cat <<EOF | oc apply -f -              
apiVersion: networkaddonsoperator.network.kubevirt.io/v1
kind: NetworkAddonsConfig
metadata:
  name: cluster
spec:
  multus: {}
  linuxBridge: {}
  kubeMacPool: {}
  ovs: {}
  kubeSecondaryDNS: {}
  kubevirtIpamController: {}
  imagePullPolicy: Always
EOF
networkaddonsconfig.networkaddonsoperator.network.kubevirt.io/cluster configured

oc get pods -ncluster-network-addons -w
NAME                                                  READY   STATUS    RESTARTS   AGE
bridge-marker-44qm4                                   1/1     Running   0          65s
cluster-network-addons-operator-6c96bbb9d9-7dcvr      2/2     Running   0          10m
kube-cni-linux-bridge-plugin-zh6bf                    1/1     Running   0          66s
kubemacpool-cert-manager-65d6b697fb-dqjgv             1/1     Running   0          65s
kubemacpool-mac-controller-manager-5db75d6c4d-hcrvf   2/2     Running   0          65s
kubevirt-ipam-controller-manager-7f588c55df-w9jpk     1/1     Running   0          64s
ovs-cni-amd64-d4tsm                                   1/1     Running   0          65s
passt-binding-cni-wh29m                               1/1     Running   0          64s
secondary-dns-6c89d7864d-rvr94                        2/2     Running   0          64s

oc get pods -oyaml -ncluster-network-addons | grep required-scc
      openshift.io/required-scc: bridge-marker
      openshift.io/required-scc: linux-bridge
      openshift.io/required-scc: restricted-v2
      openshift.io/required-scc: restricted-v2
      openshift.io/required-scc: restricted-v2
      openshift.io/required-scc: passt-binding-cni
```

data/kubemacpool/kubemacpool.yaml
maiqueb: before making me spend a lot of time on this, can I get agreement from the maintainers of those repos that they're OK with it ? 

... like, if they're not, they'll have to do this themselves in 4.19 scope ... 

@phoracek / @RamLavi / @oshoval 
who maintains the linux-bridge thing ?

oshoval: of course you do, thanks
linux-bridge is part of CNAO (all relevant in the bump-linux-bridge.sh iirc)

RamLavi (commented):
@maiqueb thank you for this change. some notes:
- Please avoid adding downstream related detail from the PR Desc: `Starting from 4.19 this is required.`.
- I am missing an explanation in the PR desc as to - why is this needed on specific components and not others ([this](https://docs.openshift.com/container-platform/4.17/authentication/managing-security-context-constraints.html#security-context-constraints-requiring_configuring-internal-oauth) says it applies to manifests with spec.template so what about `secondary-dns`?) Could you add an explanation for each component?.

maiqueb:
> @maiqueb thank you for this change. some notes:
> 
>     * Please avoid adding downstream related detail from the PR Desc: `Starting from 4.19 this is required.`.

OK, will do.

> 
>     * I am missing an explanation in the PR desc as to - why is this needed on specific components and not others ([this](https://docs.openshift.com/container-platform/4.17/authentication/managing-security-context-constraints.html#security-context-constraints-requiring_configuring-internal-oauth) says it applies to manifests with spec.template so what about `secondary-dns`?) Could you add an explanation for each component?.

Right. I'm just patching the components that I've seen reported on the periodic lanes that uncovered this issue.

maiqueb:
> > > ```
> > > * I am missing an explanation in the PR desc as to - why is this needed on specific components and not others ([this](https://docs.openshift.com/container-platform/4.17/authentication/managing-security-context-constraints.html#security-context-constraints-requiring_configuring-internal-oauth) says it applies to manifests with spec.template so what about `secondary-dns`?) Could you add an explanation for each component?.
> > > ```
> > 
> > 
> > Right. I'm just patching the components that I've seen reported on the periodic lanes that uncovered this issue.
> 
> It should be on all components that _can_ be deployed on OCP cluster.

Meaning, we're missing the kube-secondary DNS.

oshoval:
how come secondary-dns appears here as worked on ocp? 
unless something is different, it means ksd might already have whats needed (on repo itself / here) maybe?
https://github.com/kubevirt/cluster-network-addons-operator/pull/1931#discussion_r1888381172

maiqueb:
> how come secondary-dns appears here as worked on ocp? unless something is different, it means ksd might already have whats needed (on repo itself / here) maybe? [#1931 (comment)](https://github.com/kubevirt/cluster-network-addons-operator/pull/1931#discussion_r1888381172)

What it means is that kube secondary DNS is not deployed by default, and as a result is not installed in the cluster openshift conformance testing installs, and thus not shouted at by the monitor tests.

oshoval (approved):
Very nice
Thank you

maiqueb:
@phoracek do we need to do this for the OVS component ? Is anyone still using that ? FWIW, this annotation is only really required starting from 4.19.

phoracek:
@maiqueb in 4.19, I don't think so. It should be safe to ignore.

oshoval:
We might need a release if you want this code to be populated already
Thanks

maiqueb:
> We might need a release if you want this code to be populated already Thanks

For sure we need a release. Could you issue one ? @oshoval 

If not, I'll just wait for @RamLavi

oshoval:
https://github.com/kubevirt/cluster-network-addons-operator/actions/runs/13114397525

PR #1945, opened 2024-12-15
Update workflow to use Ubuntu 22.04

ashokpariya0:
It has been observed that build actions, such as free disk space, are broken with ubuntu-latest, which now points to Ubuntu 24.04. To resolve this, we are switching to the ubuntu-22.04 tag, which was previously used by ubuntu-latest.

ubuntu-latest now uses Ubuntu 24.04, while it used to use 22.04. 
To unblock Github CI in  the short term, we can switch to ubuntu-22.04,
Error log with ubuntu-latest is available at : https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4914

**Release note**:

```release-note
None
```

### DIFF
--- a/.github/workflows/kubevirt-ipam-controller.yaml
+++ b/.github/workflows/kubevirt-ipam-controller.yaml
@@ -2,7 +2,7 @@ name: Kubevirt IPAM controller Tests
 on: [pull_request]
 jobs:
   e2e:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     steps:
 
     - uses: actions/checkout@v3


ashokpariya0:
this change fixes disk cleanup issue 
cc @oshoval

oshoval:
Nice thanks
lets wait to see what OVN people say and do the same

ashokpariya0:
> Nice thanks lets wait to see what OVN people say and do the same
>
Okay, and you may have already noticed that the e2e test failure in this PR is due to the incorrect image digest for the Kubemacpool image.

oshoval:
yep thanks much
Ram is on that

oshoval:
note: we need to do the same soon on ipam-ext repo

PR #1953, opened 2024-12-17
Update Linux bridge Dockerfile to install Golang based on version in …

ashokpariya0:
…go.mod

Modified Dockerfile to dynamically install the correct Go version based on the version specified in the go.mod file.

**Issue:**
The latest tags (e.g., v1.6.1, v1.6.0, etc.) from the [containernetworking/plugins GitHub repository](https://github.com/containernetworking/plugins/tags) require Go version 1.23. However, due to the current configuration in the [cluster-network-addons-operator repository](https://github.com/kubevirt/cluster-network-addons-operator/blob/edd1798706234830cbe91fa44c554292b1f5096c/hack/components/bump-linux-bridge.sh#L47), the Dockerfile by default installs Go version 1.22.

**Solution:**
We need to modify the installation process to dynamically install the Go version based on the version specified in the go.mod file.

If the go.mod file specifies a complete Go version (e.g., major.minor.patch), the exact version will be downloaded and installed.
If the go.mod file specifies only the major and minor version (e.g., major.minor), the latest patch version for that minor version will be downloaded and installed.

**Release note**:

```release-note
None
```

### DIFF
--- a/hack/components/bump-linux-bridge.sh
+++ b/hack/components/bump-linux-bridge.sh
@@ -44,11 +44,28 @@ create_dockerfile() {
 FROM --platform=\$BUILDPLATFORM registry.access.redhat.com/ubi8/ubi-minimal AS builder
 ARG TARGETOS
 ARG TARGETARCH
-RUN microdnf install -y golang git
+ARG BUILDOS
+ARG BUILDARCH
+ENV BUILDOS=${BUILDOS:-linux}
+ENV BUILDARCH=${BUILDARCH:-amd64}
+
+RUN microdnf install -y git tar wget
 RUN \
     git clone https://${LINUX_BRIDGE_REPO} ${LINUX_BRIDGE_PATH} && \
     cd ${LINUX_BRIDGE_PATH} && \
     git checkout ${LINUX_BRIDGE_TAG}
+
+RUN GO_VERSION=$(sed -En 's/^go +(.*)$/\1/p' ${LINUX_BRIDGE_PATH}/go.mod) && \
+    if [[ "\$GO_VERSION" =~ ^1\.[0-9]+\.[0-9]+$ ]]; then \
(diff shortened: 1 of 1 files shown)


oshoval:
Thanks
I would say that once we do this, we just bump version on go.mod, and remove the auto detection
wdyt?

ashokpariya0:
> Thanks I would say that once we do this, we just bump version on go.mod, and remove the auto detection wdyt?
>
sorry, I did not get this, in this pr go.mod is refers to https://github.com/containernetworking/plugins/blob/7f756b411efc3d3730c707e2cc1f2baf1a66e28c/go.mod#L3

oshoval:
I still need to understand deeper, but lets wait with this PR
Better to keep it simple

Also note please that we can't use go 1.23 for now
Once we can, we would able to align all required repos.
For now lets stick please to the "static" method we discussed

ashokpariya0:
> I still need to understand deeper, but lets wait with this PR Better to keep it simple
> 
> Also note please that we can't use go 1.23 for now Once we can, we would able to align all required repos. For now lets stick please to the "static" method we discussed
>
Okay sure,
For the latest tags, [v1.6.0](https://github.com/containernetworking/plugins/releases/tag/v1.6.0) and [v1.6.1](https://github.com/containernetworking/plugins/releases/tag/v1.6.1), we need Go version 1.23. The previous version works with Go 1.22.

We can hold off on this PR for now and proceed with a static approach using the [v1.5.1](https://github.com/containernetworking/plugins/releases/tag/v1.5.1) tag, which is compatible with Go 1.22 ??

oshoval:
Yes please,
but we need to wait to Petr for that in order to create the images i think

ashokpariya0:
> Yes please, but we need to wait to Petr for that in order to create the images i think
>
Okay, Thanks.

oshoval:
closing this PR, we have temporary solution and i dont like dynamic install approach
in any case, upon needs we can back to it and rediscuss
please let me know if you think otherwise
thanks !

PR #1954, opened 2024-12-17
Bump ipam to v0.1.9-alpha and enable managed tap

RamLavi:
Bump kubevirt-ipam-controller to v0.1.9-alpha
Adapt to support IPAM UDN tests with managedTap binding on CNAO.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/automation/check-patch.e2e-kubevirt-ipam-controller-functests.sh
+++ b/automation/check-patch.e2e-kubevirt-ipam-controller-functests.sh
@@ -56,7 +56,7 @@ main() {
 
     cd ${TMP_COMPONENT_PATH}
     export KUBECONFIG=${TMP_COMPONENT_PATH}/.output/kubeconfig
-    export KIND_ARGS="-ic -i6 -mne"
+    export KIND_ARGS="-ic -i6 -mne -nse"
     make cluster-up
 
     trap teardown EXIT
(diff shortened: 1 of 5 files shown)


hack/deploy-kubevirt.sh
```diff
@@ -1,6 +1,19 @@
 #!/bin/bash
 set -exo pipefail
 
+if [ -z "${OCI_BIN}" ];then
```
oshoval, line 4: this is weird, we concluded at the past that we are doing it once on the bootstrap / locally and not here, 
if it is not mandatory please remove it

oshoval: we can't add the -kv, because it installs ipam and we dont want to reinstall it
and also want to control what kubevirt version we install
i would say please add it to the git actions once instead, no need each time
this what also was decided fwiw when we had same discussions on other repos

RamLavi: Moved it to the automation script so it only involves ipam-ext lane.

automation/check-patch.e2e-kubevirt-ipam-controller-functests.sh
```diff
@@ -41,18 +56,21 @@ main() {
 
     cd ${TMP_COMPONENT_PATH}
     export KUBECONFIG=${TMP_COMPONENT_PATH}/.output/kubeconfig
-    export KIND_ARGS="-ic -i6 -mne"
+    export KIND_ARGS="-ic -i6 -mne -nse"
     make cluster-up
 
     trap teardown EXIT
 
+    increase_ulimit
```
oshoval, line 64: much better thx

RamLavi:
@qinqon PTAL

maiqueb (approved):

PR #1955, opened 2024-12-17
Remove non-existent package msbuild from latest ubuntu images.

ashokpariya0:
Build actions, including free disk space, are broken on ubuntu-latest, which now points to Ubuntu 24.04. To resolve this, we have removed the non-existent package from ubuntu-latest.

We are making these changes to align with the updates in OVN-Kubernetes, as seen in https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4920.

**Release note**:

```release-note
None
```


maiqueb (approved):
tks

RamLavi (approved):
/lgtm
/approve

thanks!

PR #1968, opened 2025-01-07
Add bumper script

oshoval:
Ease go package bumping.

**Release note**:

```release-note
None
```

### DIFF
--- a/hack/bumper.sh
+++ b/hack/bumper.sh
@@ -1,0 +1,69 @@
+#!/bin/bash -e
+
+# package bump helper
+# make sure you already have a local repo with remote upstream, and origin branches
+# no untracked files allowed on folder
+# it does reset --hard so beware, also pushes to save time
+
+# ./hack/bumper.sh CVE-2021-38561 golang.org/x/text@v0.3.7 release-0.89
+
+# to skip semver check (for example in case package is dropped from go.mod after updating)
+# export SKIP=true
+# optional: to update go lang level use for example
+# export DESIRED_VERSION=1.18
+
+SKIP=${SKIP:-false}
+
+if [ $# -ne 3 ]; then
+    echo "Syntax: $0 <CVE> <TARGET_PACK> <BR>"
+    exit 1
+fi
+
+CVE=$1
+TARGET_PACK=$2
+BR=$3
+
+PACK="${TARGET_PACK/@*}"
+[ "$(git rev-parse --abbrev-ref HEAD)" != "$BR" ] && git checkout "$BR"
+git reset --hard upstream/"$(git symbolic-ref --short HEAD)" # || true # try removing the suffix, if it doesnt work try git fetch upstream first
+git pull upstream "$(git symbolic-ref --short HEAD)"
+git push -u origin
+go mod edit -dropreplace="${PACK}"
+go mod edit -require="${TARGET_PACK}"
+
+current_version=$(awk '$1 == "go" {print $2}' go.mod)
+desired_version=${DESIRED_VERSION:-$current_version}
+if [[ "$(printf "%s\n" "$current_version" "$desired_version" | sort -V | head -n1)" != "$desired_version" ]]; then
+  echo "Updating Go version in go.mod to $desired_version"
+  go mod edit -go=$desired_version
+else
+  echo "Go version in go.mod is already $desired_version or later. No update needed."
+fi
+
+if grep -q '^vendor:' Makefile; then
+    rm -rf build/_output/
+    make vendor
+else
+    go mod tidy
+    go mod vendor
+fi
+
+if grep -q '^vet:' Makefile; then
+    make vet
+fi
+
+if [ $SKIP != true ]; then
+	required_version="${TARGET_PACK/*@/}"
+	actual_version=$(go list -m -json $PACK | jq -r .Version)
+	if [ "$(printf '%s\n' "$required_version" "$actual_version" | sort -V | head -n1)" != "$required_version" ]; then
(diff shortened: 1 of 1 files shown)


oshoval:
cc @RamLavi 
also please see my comment on the gist, about why i didnt add vet yet, we can add in case needed

hack/bumper.sh
```diff
@@ -0,0 +1,28 @@
...
+CVE=$1
+TARGET_PACK=$2
+BR=$3
+
+PACK="${TARGET_PACK/@*}"
+git checkout "${BR}"
+git pull upstream "$(git symbolic-ref --short HEAD)"
+go mod edit -dropreplace="${PACK}"
+go mod edit -require="${TARGET_PACK}"
+make vendor
```
RamLavi, line 23:
```suggestion
make fmt vet
```
make fmt already does vendor
make vet checks bump didn't brake anything

oshoval: this script is meant to be generic, not just for CNAO, 
i am not sure we have on all our repos fmt / vet
(we dont even have vendor on all repos, i had a better version that handled this case)

so i prefer to not add it now, before we make sure it works all over
whoever needs can just run the additonal commands and amend
more info / options here https://github.com/kubevirt/cluster-network-addons-operator/pull/1968#issuecomment-2574607216

atm i would prefer to not block due to that

hack/bumper.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+
+# Package bump helper
+# Make sure you already have a repository (i.e CNAO) with remote upstream, and origin branches
+# No untracked files are allowed on folder (they will be added and pushed, so make sure folder is clean)
+
+# ./hack/bumper.sh CVE-2021-38561 golang.org/x/text@v0.3.7 release-0.89
+
```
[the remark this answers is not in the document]
oshoval: added

hack/bumper.sh
RamLavi: But we should fail before the PR is created, 
how about adding `go vet $WHAT` where WHAT is defaulted to `./pkg/...`?

oshoval: Even if we don't fail and just run later commit amend it is fine basically (of course we can optimize)
i am against adding something like `go vet $WHAT`, because on some repos we don't run go vet.
What i can do is add make vet if make vet target exists, this way it is generic and robust, as this script is intended

oshoval:
found the better script, lets continue from there (will add vet as done already with vendor there)

oshoval:
added the vet
script is bit more advance than the previous version, changes are based on cases i saw required once used it

hack/bumper.sh
```diff
@@ -0,0 +1,69 @@
...
+  go mod edit -go=$desired_version
+else
+  echo "Go version in go.mod is already $desired_version or later. No update needed."
+fi
+
+if grep -q '^vendor:' Makefile; then
+    rm -rf build/_output/
+    make vendor
+else
+    go mod tidy
```
RamLavi, line 47: should we do `-compat=${desired_version}`?

hack/bumper.sh
```diff
@@ -0,0 +1,69 @@
+#!/bin/bash -e
+
+# package bump helper
+# make sure you already have a local repo with remote upstream, and origin branches
+# no untracked files allowed on folder
+# it does reset --hard so beware, also pushes to save time
+
+# ./hack/bumper.sh CVE-2021-38561 golang.org/x/text@v0.3.7 release-0.89
+
+# to skip semver check (for example in case package is dropped from go.mod after updating)
```
RamLavi, line 10: I don't get the scenario where SKIP would be true.
Why would we bump a package, and it would end up getting removed?

RamLavi (commented):
Looks good overall, 
Added some comments, PTAL.

hack/bumper.sh
oshoval: Nice question
for example if there was a static `replace`, which the script removes 
https://github.com/kubevirt/cluster-network-addons-operator/pull/1968/files#diff-958372dc14cde9068a295622a0c8af96ac251e37b948589228f81ac0a7d0da6cR31

and once the script done, the new set doesn't need anymore the old set

hack/bumper.sh
oshoval: i would say not yet
because it was meant for OVS, which doesnt do it as well, so they should be the same
https://github.com/k8snetworkplumbingwg/ovs-cni/blob/main/Makefile#L91

btw what we can do for OVS, is to add `make format` if exists
https://github.com/k8snetworkplumbingwg/ovs-cni/blob/main/Makefile#L55
which does fmt / vet

RamLavi: I would say on long term align, and on short term just add all options? wdyt?

hack/bumper.sh
RamLavi: ah you mean there was a lonely replace without a require?

hack/bumper.sh
oshoval: i would suggest please to start simple, as is
and adapt OVS / make sure all the other repos are aligned, and finally change this one, wdyt?

RamLavi: yeah I'm good with that

hack/bumper.sh
oshoval: there were both

RamLavi: OK I guess. I just never encountered this scenario before.

oshoval:
Thanks
/approve

PR #1983, opened 2025-01-15
[Release 0.97] bump multus-dynamic-networks v0.3.6

RamLavi:
**What this PR does / why we need it**:
This PR is a manual bump of multus-dynamic-networks to v0.3.6

**Special notes for your reviewer**:

**Release note**:
```release-note
NONE
```

PR #1984, opened 2025-01-15
set default platform based on host-architecture for linux bridge image build.

ashokpariya0:
We encountered an unexpected failure in the job pull-cluster-network-addons-operator-unit-test-release-0.97(in pr: https://github.com/kubevirt/cluster-network-addons-operator/pull/1983) while performing a multi-architecture build for the Linux bridge. As a result, to unblock, we have adjusted the default build process to target the host architecture rather than building for all architectures by default.

**Release note**:

```release-note
None
```


oshoval:
lets do it on release-0.97 first please ?
not on main
if the same fails on main we can also do it on main

oshoval:
I wonder if same bug happens on main
worth to open an empty PR to check

PR #1985, opened 2025-01-15
set default platform based on host-architecture for linux bridge build.

ashokpariya0:
We encountered an unexpected failure in the job pull-cluster-network-addons-operator-unit-test-release-0.97(in pr: https://github.com/kubevirt/cluster-network-addons-operator/pull/1983) while performing a multi-architecture build for the Linux bridge. As a result, to unblock, we have adjusted the default build process to target the host architecture rather than building for all architectures by default.

**Release note**:

```release-note
None
```


ashokpariya0:
cc @oshoval

oshoval:
Thanks
/lgtm

cc @RamLavi

PR #2009, opened 2025-02-26
Align placement with HCO components

qinqon:
The components installed by CNAO should follow the kind of placement configuration done by HCO components by default so features like node drain work similar.

In this case CNAO was configuring deployments that are part of the controlplane as part of the Workload this means that they will run at workers and that they will not evict after drain (they should not since they are not daemonsets), this PR move those to Infra placemnt and also remove invalid tolerationsn from Infra.

Closes https://issues.redhat.com/browse/CNV-56947

**Release note**:
```release-note
Move control plane deployments to infra placement
```

### DIFF
--- a/data/kubevirt-ipam-controller/003-passtbindingcni.yaml
+++ b/data/kubevirt-ipam-controller/003-passtbindingcni.yaml
@@ -55,6 +55,6 @@ spec:
         - name: cnibin
           hostPath:
             path: {{ .CNIBinDir }}
-      nodeSelector: {{ toYaml .Placement.NodeSelector | nindent 8 }}
-      affinity: {{ toYaml .Placement.Affinity | nindent 8 }}
-      tolerations: {{ toYaml .Placement.Tolerations | nindent 8 }}
+      nodeSelector: {{ toYaml .PlacementPasst.NodeSelector | nindent 8 }}
+      affinity: {{ toYaml .PlacementPasst.Affinity | nindent 8 }}
+      tolerations: {{ toYaml .PlacementPasst.Tolerations | nindent 8 }}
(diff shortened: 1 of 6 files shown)


pkg/network/multus_dynamic_networks_controller.go
```diff
@@ -32,7 +32,7 @@ func renderMultusDynamicNetworks(conf *cnao.NetworkAddonsConfigSpec, manifestDir
 	} else {
 		data.Data["CniMountPath"] = cni.BinDir
 	}
-	data.Data["Placement"] = conf.PlacementConfiguration.Workloads
+	data.Data["Placement"] = conf.PlacementConfiguration.Infra
```
RamLavi, line 35: multus-dynamic-networks-controller deploys a DaemonSet. So it should be Workloads, no?

qinqon: Make sense, also the passt installer is a daemonset, fixing that too.

data/kubevirt-ipam-controller/003-passtbindingcni.yaml
```diff
@@ -55,6 +55,6 @@ spec:
         - name: cnibin
           hostPath:
             path: {{ .CNIBinDir }}
-      nodeSelector: {{ toYaml .Placement.NodeSelector | nindent 8 }}
-      affinity: {{ toYaml .Placement.Affinity | nindent 8 }}
-      tolerations: {{ toYaml .Placement.Tolerations | nindent 8 }}
+      nodeSelector: {{ toYaml .PlacementPasst.NodeSelector | nindent 8 }}
```
RamLavi, line 58: you also need to change it on the [bump script](https://github.com/kubevirt/cluster-network-addons-operator/blob/f1504bd54fc83f371010d8fbd858d51fea4686cc/hack/components/bump-kubevirt-ipam-controller.sh#L78-L80)

RamLavi (changes requested):

RamLavi (approved):
/lgtm

maiqueb (approved):
tks

qinqon:
/retest

Test was expecting Progressing: true but it was already Available: true

PR #2022, opened 2025-03-07
feat: add renovate config

ksimon1:
feat: add renovate config

Renovate will be used for security patches

**Release note**:
```release-note
Add renovate config
```

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,0 +1,30 @@
+{
+  "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+  "baseBranches": ["release-0.97", "release-0.95", "release-0.93", "release-0.91", "release-0.89", "release-0.79", "main"],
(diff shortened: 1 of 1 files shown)


RamLavi (approved):
/lgtm
/approve

oshoval:
Thanks
Seems that once the bot open lots of PRs at the same time, CI struggle to fetch quay images and gets timeouts
maybe we should consider please add some throttling / delay for PRs creation rate, i.e X in Y minutes most ?

cc @brianmcarey

oshoval:
@RamLavi 
i think we should revert it, wdyt?
it adds tons of noise, and we should focus on better triage and not just take whatever the bot tells
(i.e if it is based on go.mod scanning instead of deeper binary inspection as Trivy does, it will add false positive)
it also touch old releases where it is more problematic to bump packages, as some arent backward compatible

PR #2095, opened 2025-03-09
renovate: Enable only for main branch

oshoval:
We saw that enabling for all branches at once,
is harder to maintain.
Lets enable it only for main.
It will allow us to do stuff in controlled manner.

**Release note**:

```release-note
None
```

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,6 +1,6 @@
 {
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
-  "baseBranches": ["release-0.97", "release-0.95", "release-0.93", "release-0.91", "release-0.89", "release-0.79", "main"],
+  "baseBranches": ["main"],
   "prConcurrentLimit": 3,
   "lockFileMaintenance": {
     "enabled": false


oshoval:
cc @RamLavi @ksimon1 
propose to revert it please
based on what we see, imho it is more noise than good for CNAO at least

ksimon1:
I vote for keeping it at least in the main branch and fix the main branch at first. Then we can enable the release branches

oshoval:
Possible, but i would prefer first to focus on the crucial ones please
atm, the current state cause me personally more harm than good
another reason is that when we do it manually, we have the freedom to both select the important ones, and also take one issue and do it over all the required branches, this way common problems are solved horizontally.
We created few tools to assist us with bumping, so it is almost automatic even when doing manually (there are helpers i mean).

One question please, lets say we disable branches, and close the PRs, once we reenable branches
would it know to reopen the PRs ?

oshoval:
anyhow relaxed it to main only
ptal

oshoval:
but there are lot of PRs that are still opened and i would like to close them
what should we do then about them ?
i dont like dangling PRs please

/approve

oshoval:
might be real new problem / flake
but not due to this PR

/override pull-e2e-cluster-network-addons-operator-monitoring-k8s
https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_cluster-network-addons-operator/2095/pull-e2e-cluster-network-addons-operator-monitoring-k8s/1899013775378354176

PR #2118, opened 2025-03-11
chore(deps): update module golang.org/x/oauth2 to v0.27.0 [security] (release-0.95)

RamLavi:
This PR contains the following updates:

| Package | Type | Update | Change |
|---|---|---|---|
| golang.org/x/oauth2 | require | minor | `v0.21.0` -> `v0.27.0` |

---

**Release note**:

```release-note
NONE
```


go.mod
```diff
@@ -1,8 +1,8 @@
 module github.com/kubevirt/cluster-network-addons-operator
 
-go 1.22
+go 1.23.0
```
oshoval, line 3: we cant do it on release-0.95

oshoval:
we should add automatic guards that protect from those
because we all prone to forget either for automatic PRs or for manual ones
https://github.com/kubevirt/cluster-network-addons-operator/pull/2118#discussion_r1989132298

go.mod
oshoval: suggestion
https://github.com/kubevirt/cluster-network-addons-operator/pull/2120

oshoval:
used only on tools/bumper/component_commands.go
so if it is used only for non production, we can simply ignore it, and also worth to add it to ignore folders all over

PR #2121, opened 2025-03-11
checkmod: Update supported releases

oshoval:

**Release note**:

```release-note
None
```

### DIFF
--- a/hack/checkmod.sh
+++ b/hack/checkmod.sh
@@ -14,7 +14,7 @@ RELEASES=()
 GOMOD=()
 
 REPO+=("https://raw.githubusercontent.com/kubevirt/cluster-network-addons-operator")
-RELEASES+=("main release-0.89 release-0.85 release-0.79")
+RELEASES+=("main release-0.97 release-0.95 release-0.93 release-0.91 release-0.89 release-0.79")
 GOMOD+=("go.mod")
 
 REPO+=("https://raw.githubusercontent.com/containernetworking/plugins")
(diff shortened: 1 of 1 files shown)


hack/checkmod.sh
```diff
@@ -26,7 +26,7 @@ RELEASES+=("main")
 GOMOD+=("go.mod")
 
 REPO+=("https://raw.githubusercontent.com/k8snetworkplumbingwg/ovs-cni")
-RELEASES+=("main release-0.31 release-0.29")
+RELEASES+=("main release-0.36 release-0.34 release-0.32 release-0.31 release-0.29")
 GOMOD+=("go.mod")
 
 REPO+=("https://raw.githubusercontent.com/k8snetworkplumbingwg/kubemacpool")
```
oshoval, line 32: @RamLavi can you suggest please what are the branches we need to monitor for KMP so i will update it ?

hack/checkmod.sh
```diff
@@ -26,7 +26,7 @@ RELEASES+=("main")
 GOMOD+=("go.mod")
 
 REPO+=("https://raw.githubusercontent.com/k8snetworkplumbingwg/ovs-cni")
-RELEASES+=("main release-0.31 release-0.29")
+RELEASES+=("main release-0.36 release-0.34 release-0.32 release-0.31 release-0.29")
```
oshoval, line 29: @phoracek  can you suggest please what are the branches we need to monitor for KMP so i will update it ?
(if need i will check, just if you know please by heart maybe)

oshoval:
found the answer based on Renovate config, sorry for the noise

@RamLavi @qinqon can you please review ?

@ksimon1 you might be interested in this script, it helps to scan all the current packages all over the repos / branches
in order to easily know leftovers of CVEs that need to be fixed, package wise

brianmcarey:
> @brianmcarey lately we have lots of quay hiccups such as https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_cluster-network-addons-operator/2121/pull-e2e-cluster-network-addons-operator-ovs-cni-functests/1899452493977358336
> 
> anything we can do about it maybe please? some short circuit maybe if the images are in cache and match the desired hash ? (lets assume we use hashes and not tags, if needed we can adapt to that maybe at least on some cases) or any other solution ? i.e change pull policy and make sure the images are available, for example by fetching them from a mirror and tagging them as required)

Are you using the latest kubevirtci tag from the main branch of kubevirt? These images should be available locally to the test pod so you wouldn't hit these issues.

oshoval:
i bet not, especially not on other branches, will check, thanks !

RamLavi (approved):
/lgtm
All the versions are good, 

not now but a good follow up could be fetching the relevant stable branches from the actual components.yaml on each stable branch..

oshoval:
> not now but a good follow up could be fetching the relevant stable branches from the actual components.yaml on each stable branch..

cant be done because you need oracle to tell you what is the supported stable branches that are still important
i.e no script can tell you that 0.85 is EOL while 0.79 isnt
what can be done is to fetch from renovate config, and that one can fetch from other centralized place

oshoval:
/retest

maybe the BM did some regression, lets see

PR #2150, opened 2025-03-17
[release-0.97] components/kmp: Follow tagged releases

RamLavi:
**What this PR does / why we need it**:
Following the increase in traffic, limiting kubmacpool bumps to only tagged versions on the stable branch

**Special notes for your reviewer**:

**Release note**:

```release-note
NONE
```

PR #2152, opened 2025-03-17
components/kmp: Follow latest main branch

RamLavi:
CNAO main branch should follow the latest changes on kubemacpool, and if necessary - fail fast in order to fix fast.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/components.yaml
+++ b/components.yaml
@@ -15,7 +15,7 @@ components:
     url: https://github.com/k8snetworkplumbingwg/kubemacpool
     commit: 9be6f3e0b82b8b9796fe039d501668daa22d261f
     branch: main
-    update-policy: tagged
+    update-policy: latest
     metadata: v0.46.0
   kubevirt-ipam-controller:
     url: https://github.com/kubevirt/ipam-extensions


RamLavi:
/override pull-e2e-cluster-network-addons-operator-br-marker-functests
irrelevant for PR

RamLavi:
/approve

@oshoval can you take a look? this will incorporate the fix to the KMP tier1 tests to CNAO main branch and should give us more insight on the KMP flakes we're seeing.

oshoval:
Why latest ?
it can add noise of PRs,  especially because of renovate
lets just create tag please imho, i prefer to keep current state

RamLavi:
> Why latest ? it can add noise of PRs, especially because of renovate lets just create tag please imho, i prefer to keep current state

Kubemacpool does not create main branch releases - all major/minor releases are starting stable branches. but bumping then can reveal issues on the last minute which is not good - we shouldn't wait until we branch to know we have a problem. Nor does it make sense to issue "main branch minor releases" just to keep CNAO up to date, it's against semver [conventions](https://semver.org/#spec-item-7) (if no big feature was added, then no reason to issue minor) and poses too much manual labor on our part.

main branch is by definition not stable - it is meant for bleeding changes. We do this because we want to fail fast, and fix fast accordingly. Same goes to CVEs - we want it in asap. CNAO also is protected by tier1 tests so we should be good stability wise.
 As for noise - I don't think that renovate is going to cause many more PR storms.. I propose we try and see if renovate on main branch is creating so much noise as you suggest, then suggest alternatives (like increasing the bumper intervals).

oshoval:
> As for noise - I don't think that renovate is going to cause many more PR storms.. I propose we try and see if renovate on main branch is creating so much noise as you suggest, then suggest alternatives (like increasing the bumper intervals).

I am fine to try, but if CNAO gets unstable (it already has flakes, we can't get it worse please, unless you handle you know you handle them once they come by this change) i think we should revert, not just have other intervals.
Another option is to add git actions that sync kmp upon request and then once needed you sync it, like now, personally i think it is even better then the current PR / can be a fallback in case of problems

RamLavi:
> > As for noise - I don't think that renovate is going to cause many more PR storms.. I propose we try and see if renovate on main branch is creating so much noise as you suggest, then suggest alternatives (like increasing the bumper intervals).
> 
> Theoretically I am fine to try, but if CNAO gets unstable (it already has flakes, we can't get it worse please, unless you handle you know you handle them once they come by this change) i think we should revert, not just have other intervals.

I really don't think that KMP PRs introduce new instability. We haven't touched the core code in the last 2 years, so I'm not sure what you are basing this statement on.

> Another option is to add git actions that sync kmp upon request and then once needed you sync it, like now, personally i think it is even better then the current PR / can be a fallback in case of problems, wdyt?

You are talking about setting KMP statically to the top of main branch right? We can do that, but  unless we see 10 KMP bump PRs a week as a result of following main branch'es HEAD, I don't think it's worth shutting down the bumper for it. But let's try and see where it takes us. We can always set the bumps to static as you suggest.

> 
> Moreover since usually we don't have cutting edge problems and KMP is tested fine already (correct me if needed please), then just for the logs, maybe worth to keep the current state

oshoval:
> > > As for noise - I don't think that renovate is going to cause many more PR storms.. I propose we try and see if renovate on main branch is creating so much noise as you suggest, then suggest alternatives (like increasing the bumper intervals).
> > 
> > 
> > Theoretically I am fine to try, but if CNAO gets unstable (it already has flakes, we can't get it worse please, unless you handle you know you handle them once they come by this change) i think we should revert, not just have other intervals.
> 
> I really don't think that KMP PRs introduce new instability. We haven't touched the core code in the last 2 years, so I'm not sure what you are basing this statement on.

I meant CNAO has flakes, the ones we see from time to time, deploy all components and such

> 
> > Another option is to add git actions that sync kmp upon request and then once needed you sync it, like now, personally i think it is even better then the current PR / can be a fallback in case of problems, wdyt?
> 
> You are talking about setting KMP statically to the top of main branch right? We can do that, but unless we see 10 KMP bump PRs a week as a result of following main branch'es HEAD, I don't think it's worth shutting down the bumper for it. But let's try and see where it takes us. We can always set the bumps to static as you suggest.

I meant to revert to the current state in case we see it is needed, and to add a git actions that allow to bump ad hoc easily
(maybe the way to do that is as you just said indeed)

> 
> > Moreover since usually we don't have cutting edge problems and KMP is tested fine already (correct me if needed please), then just for the logs, maybe worth to keep the current state

RamLavi:
> > > > As for noise - I don't think that renovate is going to cause many more PR storms.. I propose we try and see if renovate on main branch is creating so much noise as you suggest, then suggest alternatives (like increasing the bumper intervals).
> > > 
> > > 
> > > Theoretically I am fine to try, but if CNAO gets unstable (it already has flakes, we can't get it worse please, unless you handle you know you handle them once they come by this change) i think we should revert, not just have other intervals.
> > 
> > 
> > I really don't think that KMP PRs introduce new instability. We haven't touched the core code in the last 2 years, so I'm not sure what you are basing this statement on.
> 
> I meant CNAO has flakes, the ones we see from time to time, deploy all components and such

This is a sour subject indeed, but it has nothing to do with this change. I suggest we talk about it on a different thread. I'll open a task about it.

> 
> > > Another option is to add git actions that sync kmp upon request and then once needed you sync it, like now, personally i think it is even better then the current PR / can be a fallback in case of problems, wdyt?
> > 
> > 
> > You are talking about setting KMP statically to the top of main branch right? We can do that, but unless we see 10 KMP bump PRs a week as a result of following main branch'es HEAD, I don't think it's worth shutting down the bumper for it. But let's try and see where it takes us. We can always set the bumps to static as you suggest.
> 
> I meant to revert to the current state in case we see it is needed, and to add a git actions that allow to bump ad hoc easily (maybe the way to do that is as you just said indeed)

This is the current way to do so. I think it's easy enough, but having CNAO track the latest is the best approach for main branch. If we see we are bombarded with bumps then we should reconsider. Note that this is exactly what we did on #2151, #2150. 

> 
> > > Moreover since usually we don't have cutting edge problems and KMP is tested fine already (correct me if needed please), then just for the logs, maybe worth to keep the current state

oshoval:
https://github.com/kubevirt/cluster-network-addons-operator/pull/2166

i think i dont like it at all, but lets see

PR #2156, opened 2025-03-18
Revert "bump bridge-marker to 0.12.1 (#2128)"

oshoval:
This reverts commit 2a7253c4152b16f8497ee2545bd387629753783b.

**Release note**:

```release-note
None
```

### DIFF
--- a/components.yaml
+++ b/components.yaml
@@ -1,10 +1,10 @@
 components:
   bridge-marker:
     url: https://github.com/kubevirt/bridge-marker
-    commit: 5fb005a5aa5e2d9d5bec2814e680545d3ad67332
+    commit: a2c347b149a4a6e72f43f415325c2af40c34cb0b
     branch: main
     update-policy: tagged
-    metadata: 0.12.1
+    metadata: 0.12.0
   kube-secondary-dns:
     url: https://github.com/kubevirt/kubesecondarydns
     commit: dcfa3bf3a97ae62976f23c0a9a21c186907c9d97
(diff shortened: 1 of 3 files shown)


oshoval:
ok seems the revert doesnt help

cc @phoracek @RamLavi

oshoval:
/hold
doesnt seem related


/retest

ashokpariya0:
Is there any correlation with the kubevirtci tag? We see more consistent failures with the bridge marker test when using the kubevirtci tag `2503111657-1a9278ce`.
https://github.com/kubevirt/cluster-network-addons-operator/pull/2131/files

However, the last merged PR passed(2days back) with that same tag, so we're not entirely sure what's causing the issue:  
https://github.com/kubevirt/cluster-network-addons-operator/pull/2149

Maybe we should try using the latest tag instead?

oshoval:
interesting, might be indeed the kubevirtci , or any PR of the last days
even that it passed it might be just "good path" of a bad flake

lets first rerun

thanks

ashokpariya0:
> we should bi sect, would you like to try please ? for example to revert the kubevirtci as you said

Sure, can we do following ?
1) first try with revert https://github.com/kubevirt/cluster-network-addons-operator/pull/2131
2) may be we can also try using latest tag(https://github.com/kubevirt/kubevirtci/tags) from kubevirtci ?

oshoval:
> > we should bi sect, would you like to try please ? for example to revert the kubevirtci as you said
> 
> Sure, can we do following ?
> 
> 1. first try with revert [Bump kubevirtci #2131](https://github.com/kubevirt/cluster-network-addons-operator/pull/2131)
> 2. may be we can also try using latest tag(https://github.com/kubevirt/kubevirtci/tags) from kubevirtci ?

sure thanks, just note please the that tag we use is pretty new, so it is marginal that [2] would help
there is also https://github.com/kubevirt/cluster-network-addons-operator/pull/2149 that can be the one, but very low chance,
i think it worth to run it locally, if it happens locally, it is much easier to bi sect

oshoval:
closing this PR
thx

PR #2157, opened 2025-03-19
Revert "Bump kubevirtci"

ashokpariya0:
Reverts kubevirt/cluster-network-addons-operator#2131

**What this PR does / why we need it:**
We observed flakiness in the bridge marker test with the latest kubevirtci tag, so we're reverting to the previous tag to see if that resolves the issue.

**Special notes for your reviewer:**

**Release note:**
```
None
```


oshoval:
same, thx for trying
maybe we need to find the earliest failure and go before that ?

PR #2195, opened 2025-03-30
e2e/workflow: Fix race waiting for components

### LINKED ISSUE #626: [Flaky CI] NetworkAddonsConfig when there is no pre-existing Config should be able to deploy all components one by one
**What happened**:
prow [link](https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_cluster-network-addons-operator/625/pull-e2e-cluster-network-addons-operator-workflow-k8s/1319183344935636992)

**Anything else we need to know?**:

**Environment**:

RamLavi:
Currently checkConfigChange waiting on the transient status condition of
the CNAO CR, in order to make sure that the component is properly
deployed.
This is raceful by design, as if the component deployment moved "too
fast" to ConditionProgressing=TRUE then the test will fail.
Also, there is another race where the Available condition might change before the operator changed the status to in progress. 

This PR is changing the test to fix these two races.

Fix #626

**Release note**:

```release-note
NONE
```

### DIFF
--- a/test/check/check.go
+++ b/test/check/check.go
@@ -145,6 +145,76 @@ func CheckConfigCondition(gvk schema.GroupVersionKind, conditionType ConditionTy
 	}
 }
 
+func CheckConfigConditionChangedAfter(
+	gvk schema.GroupVersionKind,
+	conditionType ConditionType,
+	expectedStatus ConditionStatus,
+	checkCondTimestampMap map[conditionsv1.ConditionType]time.Time,
+	timeout, duration time.Duration,
+) {
+	By(fmt.Sprintf("Checking that condition %q status is %q and changed after previous transition time", conditionType, expectedStatus))
+
+	getAndCheck := func() error {
+		return checkConfigConditionChangedAfter(gvk, conditionsv1.ConditionType(conditionType), corev1.ConditionStatus(expectedStatus), checkCondTimestampMap)
+	}
+
+	if timeout != CheckImmediately {
+		Eventually(getAndCheck, timeout, 10*time.Millisecond).ShouldNot(HaveOccurred(), fmt.Sprintf("Timed out waiting for condition %q to change", conditionType))
+	} else {
+		Expect(getAndCheck()).NotTo(HaveOccurred(), fmt.Sprintf("Condition %q is not in expected state or hasn't changed", conditionType))
+	}
+
+	if duration != CheckDoNotRepeat {
+		Consistently(getAndCheck, duration, 10*time.Millisecond).ShouldNot(HaveOccurred(), fmt.Sprintf("Condition %q prematurely changed again", conditionType))
+	}
+}
+
+func checkConfigConditionChangedAfter(
+	gvk schema.GroupVersionKind,
+	conditionType conditionsv1.ConditionType,
+	expectedStatus corev1.ConditionStatus,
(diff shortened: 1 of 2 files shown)


RamLavi:
unrelated flake, opened an issue.
/retest

test/check/check.go
```diff
@@ -75,6 +75,61 @@ func CheckComponentsRemoval(components []Component) {
...
+	}
+	if configV1.Spec.Multus != nil {
+		existingComponentsMap[MultusComponent.ComponentName] = struct{}{}
+	}
+	if configV1.Spec.MultusDynamicNetworks != nil {
+		existingComponentsMap[MultusDynamicNetworks.ComponentName] = struct{}{}
+	}
+	if configV1.Spec.MultusDynamicNetworks != nil {
+		existingComponentsMap[MultusDynamicNetworks.ComponentName] = struct{}{}
+	}
```
qinqon, line 114: dup

test/check/check.go
```diff
@@ -75,6 +75,61 @@ func CheckComponentsRemoval(components []Component) {
 	}
 }
 
+func getConfigComponentsMap(gvk schema.GroupVersionKind) (map[string]struct{}, error) {
```
qinqon, line 78: at golang sets are usually expressed as `map[string]bool`

qinqon (changes requested):

test/check/check.go
RamLavi: fixed! thanks

test/check/check.go
RamLavi: There is no need for a bool, I only want to check if the component is "set" or not. 
IIUC `map[string]struct{}` is idiomatic and preferred in Go when you don't need to store a value, only the presence of a key.

test/check/check.go
oshoval: interesting if there is a tool that can detect copy paste dups like this
sometimes it makes bugs, here it was just cosmetics

test/check/check.go
oshoval: i also think bool is better

test/check/check.go
RamLavi: perhaps introducing golint will be a good idea for CNAO, but it's not on the near future roadmap as it will take time to refactor the code.

oshoval: ye no rush about it imho as well

test/check/check.go
oshoval:
```
if _, exist := deployedComponents[component.ComponentName]; !exist {
   foo()
}
```

semantic bool is easier to read imho

qinqon (approved):
/lgtm
/approve

test/check/check.go
RamLavi: yeah the boolValue wan't really needed I just wanted to make it more obvious..

Well then I guess it's personal preference. I like struct{} cause it saves memory space, but again it's not super important.

RamLavi:
This fix seems safe to me, and it looks like it fixed the flake. 
backporting to the stable branches

/cherry-pick release-0.79
/cherry-pick release-0.89
/cherry-pick release-0.91
/cherry-pick release-0.93
/cherry-pick release-0.95
/cherry-pick release-0.97

FYI @oshoval

PR #2196, opened 2025-03-30
[release-0.97] components, ovs-cni: Follow tagged

RamLavi:
OVS-CNI is to follow tagged releases on stable branch.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/components.yaml
+++ b/components.yaml
@@ -51,5 +51,5 @@ components:
     url: https://github.com/k8snetworkplumbingwg/ovs-cni
     commit: 641e6348f9a6ea234e216a84784f0e1d90c80ade
     branch: release-0.36
-    update-policy: latest
+    update-policy: tagged
     metadata: v0.36.0


components.yaml
```diff
@@ -51,5 +51,5 @@ components:
     url: https://github.com/k8snetworkplumbingwg/ovs-cni
     commit: 641e6348f9a6ea234e216a84784f0e1d90c80ade
     branch: release-0.36
-    update-policy: latest
+    update-policy: tagged
```
oshoval, line 54: since it is already on a branch, why do we need tagged?
unless few releases use same branch (which also doesnt make sense to do, because once both follow tagged / latest, they still have same fate)

RamLavi: I want to reduce noise on stable branches, if I'm not the one maintaining them.

oshoval (approved):

PR #2216, opened 2025-04-01
automation: Add workflow nightly script

RamLavi:
This PR adds a nightly script, that will be run by project-infra.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/automation/test-nightly-build.sh
+++ b/automation/test-nightly-build.sh
@@ -1,0 +1,51 @@
+#!/bin/bash -xe
+
+# This script should be able to execute workflow functional tests against
+# Kubernetes cluster on any environment with basic dependencies listed in
+# check-patch.packages installed and docker running.
+
+teardown() {
+    # Don't fail if there is no logs
+    cp ${E2E_LOGS}/workflow/*.log ${ARTIFACTS} || true
+    export REPORT_FLAG="${ARTIFACTS}/junit.xml"
+
+    make cluster-down
+}
+
+main() {
+    source automation/check-patch.setup.sh
+    cd ${TMP_PROJECT_PATH}
+
+    make cluster-down
+    make cluster-up
+    trap teardown EXIT SIGINT SIGTERM
+
+    # create and push bundle image and index image
+    build_date="$(date +%Y%m%d)"
+    export IMAGE_REGISTRY=quay.io/kubevirtci
+    export IMAGE_TAG="nb_${build_date}_$(git show -s --format=%h)"
+
+    # build CNAO operator
+    make cluster-operator-push
+    make cluster-operator-install
+
+    # run functest
+    make E2E_TEST_EXTRA_ARGS="-ginkgo.noColor --ginkgo.junit-report=$ARTIFACTS/junit.xml" test/e2e/workflow
+
+    # functional test passed: publish latest nightly build
+    echo "${build_date}" > build-date
+    gsutil cp ./build-date gs://${cnao_bucket}/latest
+
+    # publish the nightly build manifests
+    CNAO_VERSION=$(hack/version.sh)
+    SRC_DIR="./_out/cluster-network-addons/${CNAO_VERSION}"
+    DEST="gs://${cnao_bucket}/${build_date}"
+    cnao_bucket="kubevirt-prow/devel/nightly/release/kubevirt/cluster-network-addons-operator"
+    gsutil cp "${SRC_DIR}/namespace.yaml" "${DEST}"
+    gsutil cp "${SRC_DIR}/cluster-network-addons-operator.${CNAO_VERSION}.clusterserviceversion.yaml" "${DEST}"
+    gsutil cp "${SRC_DIR}/network-addons-config.crd.yaml" "${DEST}"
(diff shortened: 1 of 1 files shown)


automation/test-nightly-build.sh
```diff
@@ -0,0 +1,50 @@
...
+teardown() {
+    # Don't fail if there is no logs
+    cp ${E2E_LOGS}/workflow/*.log ${ARTIFACTS} || true
+    export REPORT_FLAG="${ARTIFACTS}/junit.xml"
+
+    make cluster-down
+}
+
+main() {
+    export KUBEVIRT_PROVIDER='k8s-1.25'
```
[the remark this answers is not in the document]
RamLavi: short answer is not really.
Long answer is - we want to separate between successful build (==>save to gs filesystem) and successful e2e (==>signal HCO it's ok to use on their nightly), and that logic kinda interrupts the current workflow automation script. And making it more robust for this use case will just make it unnecessarily messy...

Removeing the KUBEVIRT_PROVIDER set. The nightly should use the default..

RamLavi:
/retest
opened issue on failed tests

automation/test-nightly-build.sh
```diff
@@ -0,0 +1,48 @@
...
+    make cluster-operator-push
+    make cluster-operator-install
+
+    # build succeeded: publish the nightly build manifests
+    cnao_bucket="kubevirt-prow/devel/nightly/release/kubevirt/cluster-network-addons-operator"
+    gsutil cp ./_out/cluster-network-addons/99.0.0/namespace.yaml "gs://${cnao_bucket}/${build_date}"
+    gsutil cp ./_out/cluster-network-addons/99.0.0/cluster-network-addons-operator.99.0.0.clusterserviceversion.yaml "gs://${cnao_bucket}/${build_date}"
+    gsutil cp ./_out/cluster-network-addons/99.0.0/network-addons-config.crd.yaml "gs://${cnao_bucket}/${build_date}"
+    gsutil cp ./_out/cluster-network-addons/99.0.0/operator.yaml "gs://${cnao_bucket}/${build_date}"
+    gsutil cp ./_out/cluster-network-addons/99.0.0/network-addons-config-example.cr.yaml "gs://${cnao_bucket}/${build_date}"
```
nunnatsa, line 38: Maybe it would be better to change the version in the source path, into a variable. It would be even better if you could compute it, so you won't need to visit this file each time you start a new version.

also, I think the source dir should be in a variable; e.g.
```suggestion
    SRC_DIR="./_out/cluster-network-addons/${CNAO_VERSION}"
    DEST="gs://${cnao_bucket}/${build_date}"
    gsutil cp "${SRC_DIR}/namespace.yaml" "${DEST}"
    gsutil cp "${SRC_DIR}/cluster-network-addons-operator.${CNAO_VERSION}.clusterserviceversion.yaml" "${DEST}"
    gsutil cp "${SRC_DIR}/network-addons-config.crd.yaml" "${DEST}"
    gsutil cp "${SRC_DIR}/operator.yaml" "${DEST}"
    gsutil cp "${SRC_DIR}/network-addons-config-example.cr.yaml" "${DEST}"
```

automation/test-nightly-build.sh
```diff
@@ -0,0 +1,48 @@
...
+
+    # run functest
+    make E2E_TEST_EXTRA_ARGS="-ginkgo.noColor --ginkgo.junit-report=$ARTIFACTS/junit.xml" test/e2e/workflow
```
nunnatsa, line 41: Maybe move this before the pushing to the remote bucket, so you'll have a new build only if the tests passed?

nunnatsa (commented):
Hi @RamLavi .

Added two comments inline. Sorry for the late response.

automation/test-nightly-build.sh
RamLavi: >Maybe it would be better to change the version in the source path, into a variable. It would be even better if you could compute it, so you won't need to visit this file each time you start a new version.

in CNAO the version is taken from a file named `version`. we can compute it from that.

automation/test-nightly-build.sh
RamLavi: yeah that makes sense.. Will change.

RamLavi (commented):
@nunnatsa thanks for the review! PTAL

automation/test-nightly-build.sh
```diff
@@ -0,0 +1,51 @@
...
+
+    # functional test passed: publish latest nightly build
+    echo "${build_date}" > build-date
+    gsutil cp ./build-date gs://${cnao_bucket}/latest
+
+    # publish the nightly build manifests
+    CNAO_VERSION=$(hack/version.sh)
+    SRC_DIR="./_out/cluster-network-addons/${CNAO_VERSION}"
+    DEST="gs://${cnao_bucket}/${build_date}"
+    cnao_bucket="kubevirt-prow/devel/nightly/release/kubevirt/cluster-network-addons-operator"
```
nunnatsa, line 43: The test failed in line 37. We need to move this line up, above line 37.

PR #2217, opened 2025-04-01
gitActions: Add workflow to bump kubevirtci

RamLavi:
This PR introduces a gitActions that will run the kubevirtci bumper every 3 months, in order to keep CNAO up to date.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/.github/workflows/kubevirtci-bumper.yaml
+++ b/.github/workflows/kubevirtci-bumper.yaml
@@ -1,0 +1,61 @@
+name: Bump KubevirtCI
+
+on:
+  schedule:
+    - cron: '0 0 1 1,4,7,10 *'  # every 3 months (00:00 UTC on the 1st of Jan, Apr, Jul, Oct)
+  workflow_dispatch:
+
+jobs:
+  bump-kubevirtci:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      pull-requests: write
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Set up Git config
+        run: |
+          git config user.name "github-actions[bot]"
+          git config user.email "github-actions[bot]@users.noreply.github.com"
+
+      - name: Bump KubevirtCI to latest tag
+        run: |
+          ./hack/bump-kubevirtci.sh
+
+      - name: Check for changes
+        id: changes
+        run: |
+          if git diff --quiet cluster/cluster.sh; then
+            echo "changed=false" >> $GITHUB_OUTPUT
+          else
+            echo "changed=true" >> $GITHUB_OUTPUT
+          fi
+
+      - name: Extract new tag from cluster.sh
+        if: steps.changes.outputs.changed == 'true'
+        id: extract
+        run: |
+          NEW_TAG=$(grep -oP 'KUBEVIRTCI_TAG:-\K[^}]+' cluster/cluster.sh)
+          echo "Detected new tag: $NEW_TAG"
+          echo "tag=$NEW_TAG" >> $GITHUB_OUTPUT
+
+      - name: Create Pull Request
+        if: steps.changes.outputs.changed == 'true'
+        uses: peter-evans/create-pull-request@v5
+        with:
+          token: ${{ secrets.KUBEVIRT_BOT_TOKEN }}
+          committer: GitHub <noreply@github.com>
+          commit-message: "Bump KubevirtCI"
+          title: "cluster: Bump KubevirtCI"
+          signoff: true
+          branch: bump_kubevirtci_${{ github.run_id }}
+          body: |
+            **What this PR does / why we need it**:
+            This PR updates relevant KubevirtCI components to latest available.
+
+            ```release-note
+            NONE
+            ```


RamLavi:
@oshoval wdyt?

oshoval:
we might want to make sure 
https://github.com/kubevirt/cluster-network-addons-operator/pull/2131/files#diff-030affb1cecb4332ce886f9c3e14ed4740eff8868bfa708e5ae8f83f0bb133aaR4

this change is robust, i just changed it to make it work, not sure if it is consistent 
if it isnt it meant the job will fail and we dont see it

RamLavi:
> Nice thanks lets change to 3M?

np

> 
> One thing to consider please, is the k8s version we use, sometimes the old one is dropped we can introduce auto detection of the latest version, wdyt? otoh sometimes kubevirtci has a newer version then the one we aim to, (and sometimes we use older one than the one we aim to) the best will be if we could query to detect the desired current k8s version automatically

Do you mean the KUBEVIRT_PROVIDER?

oshoval:
> Do you mean the KUBEVIRT_PROVIDER?

yep

RamLavi:
> > Do you mean the KUBEVIRT_PROVIDER?
> 
> yep

We can add another workflow to bump to the maximum KUBEVIRT_PROVIDER (with format of k8s-x.y). wdyt?

RamLavi:
> we might want to make sure https://github.com/kubevirt/cluster-network-addons-operator/pull/2131/files#diff-030affb1cecb4332ce886f9c3e14ed4740eff8868bfa708e5ae8f83f0bb133aaR4
> 
> this change is robust, i just changed it to make it work, not sure if it is consistent if it isnt it meant the job will fail and we wont see it

If we see we get gitAction fails, we'll tackle it then I think..

oshoval:
> We can add another workflow to bump to the maximum KUBEVIRT_PROVIDER (with format of k8s-x.y). wdyt?

it is problematic, because sometimes latest kubevirtci is more advance than the one we are working on, since the upstream CI team adds it before its needed

for sure we can discuss it on follow ups with Brian and Daniel

thanks

.github/workflows/kubevirtci-tag-bumper.yaml
```diff
@@ -0,0 +1,53 @@
...
+      - name: Extract new tag from cluster.sh
+        if: steps.changes.outputs.changed == 'true'
+        id: extract
+        run: |
+          NEW_TAG=$(grep -oP 'KUBEVIRTCI_TAG:-\K[^}]+' cluster/cluster.sh)
+          echo "Detected new tag: $NEW_TAG"
+          echo "tag=$NEW_TAG" >> $GITHUB_OUTPUT
+
+      - name: Create Pull Request
+        if: steps.changes.outputs.changed == 'true'
```
oshoval, line 46: can we compact those steps to only one ?
maybe directly `if git diff ...; do stuff..` ?

ashokpariya0:
Is it possible to check whether a tag supports multi-architecture? For example, the latest tag (2504011456-29ba6f0e) from [KubeVirtCI GitHub tags](https://github.com/kubevirt/kubevirtci/tags) does not support multi-architecture (you can verify this by checking the corresponding Go-CLI tag on [Quay.io](https://quay.io/repository/kubevirtci/gocli?tab=tags&tag=latest)), while the previous tag (2503281142-a291de27) does. Therefore, always using the latest tag might not be reliable

oshoval:
> Is it possible to check whether a tag supports multi-architecture? For example, the latest tag (2504011456-29ba6f0e) from [KubeVirtCI GitHub tags](https://github.com/kubevirt/kubevirtci/tags) does not support multi-architecture (you can verify this by checking the corresponding Go-CLI tag on [Quay.io](https://quay.io/repository/kubevirtci/gocli?tab=tags&tag=latest)), while the previous tag (2503281142-a291de27) does. Therefore, always using the latest tag might not be reliable

either by podman inspect, or on quay.io web page it shows
(also with docker there is a trick, but with podman its easier)

RamLavi:
> > Is it possible to check whether a tag supports multi-architecture? For example, the latest tag (2504011456-29ba6f0e) from [KubeVirtCI GitHub tags](https://github.com/kubevirt/kubevirtci/tags) does not support multi-architecture (you can verify this by checking the corresponding Go-CLI tag on [Quay.io](https://quay.io/repository/kubevirtci/gocli?tab=tags&tag=latest)), while the previous tag (2503281142-a291de27) does. Therefore, always using the latest tag might not be reliable
> 
> either by podman inspect, or on quay.io web page it shows (also with docker there is a trick, but with podman its easier)
> 
> we dont need to add such validation to git actions otoh for that we have the lanes you and us create, if it fails we will see there there was a little hiccup thats all, it doesnt mean we need to be extra safe from now on in this aspect

I agree with @oshoval . If any, a check like this should be inside the bump-kubevirt.sh script itself, and not in this workflow.

oshoval:
> I agree with @oshoval . If any, a check like this should be inside the bump-kubevirt.sh script itself, and not in this workflow.

I would say no check at all, the lanes are already good enough
we can't add validation based on every problem we see, it is expansive to run and maintain
but thanks a lot for caring and contributing

.github/workflows/kubevirtci-tag-bumper.yaml
RamLavi: no, because this step (by peter-evans) is a standalone gitActions building block. If we wanted to bulk it together, we'd need to implement it ourselves like in the auto bumper case. But that's not necessary.

.github/workflows/kubevirtci-tag-bumper.yaml
```diff
@@ -0,0 +1,53 @@
...
+          git config user.name "github-actions[bot]"
+          git config user.email "github-actions[bot]@users.noreply.github.com"
+
+      - name: Run bump-kubevirtci.sh
+        run: |
+          ./hack/bump-kubevirtci.sh
+
+      - name: Check for changes
+        id: changes
+        run: |
```
oshoval, line 30: please consider 
`run: echo "changed=$(! git diff --quiet cluster/cluster.sh && echo true || echo false)" >> "$GITHUB_OUTPUT"`

.github/workflows/kubevirtci-tag-bumper.yaml
```diff
@@ -0,0 +1,53 @@
+name: Bump KubevirtCI Tag
+
+on:
+  schedule:
+    - cron: '0 0 1 1,5,9 *'  # 00:00 on the 1st of Jan, May, Sep
```
oshoval, line 5: please change to 3M, worth to change the comment to say it is 3M

ashokpariya0:
Okay thanks, Yes, I agree with you guys; there's no need to add such checks, We'll see if anything fails.

RamLavi:
> > We can add another workflow to bump to the maximum KUBEVIRT_PROVIDER (with format of k8s-x.y). wdyt?
> 
> it is problematic, because sometimes latest kubevirtci is more advance than the one we are working on, since the upstream CI team adds it before its needed
> 
> for sure we can discuss it on follow ups with Brian and Daniel
> 
> thanks

Then I'll try to add a step that uses the tag chosen to figure out the latest provider, and use that

oshoval:
> > > We can add another workflow to bump to the maximum KUBEVIRT_PROVIDER (with format of k8s-x.y). wdyt?
> > 
> > 
> > it is problematic, because sometimes latest kubevirtci is more advance than the one we are working on, since the upstream CI team adds it before its needed
> > for sure we can discuss it on follow ups with Brian and Daniel
> > thanks
> 
> Then I'll try to add a step that uses the tag chosen to figure out the latest provider, and use that

it isnt good, because latest might be newer than the one we use, because CI team adds the cutting edge, yet used version

oshoval (approved):

.github/workflows/kubevirtci-tag-bumper.yaml
RamLavi: I would prefer to keep as is

.github/workflows/kubevirtci-tag-bumper.yaml
RamLavi: DONE

RamLavi:
@oshoval added a step to also bump kubevirt-provider. PTAL

hack/bump-kubevirt-provider.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash -e
+
+[[ -z "$1" ]] && echo "error: missing required tag argument" >&2 && exit 1
+TAG="$1"
+
+LATEST_K8S=$(curl -s "https://api.github.com/repos/kubevirt/kubevirtci/contents/cluster-up/cluster?ref=${TAG}" \
+  | grep '"name": "k8s-[0-9]\+\.[0-9]\+"' \
+  | sed -E 's/.*"k8s-([0-9]+\.[0-9]+)".*/\1/' \
+  | sort -V \
+  | tail -n1)
```
oshoval, line 10: Thanks
didn't look on all deeply yet but:

we cant take `tail -n1`
even if we wanted be on the bleeding edge (i.e k8s version higher than the one that is recommended at the very moment)
we can't, because when we freeze and create a branch, the branch need to use the recommended one

i.e release-0.97 uses k8s version X, not X+1 even if kubevirtci already has it

what you can do is to count, and always take the 3rd starting from oldest
on most of the cases it will work, it would stop working if the providers will have suddenly different naming,
as happened in the past, where we had
k8s-x-centos9 and such, which might breaks this formula, but we can do it meanwhile, and be proactive according needs

RamLavi: ah I see. 
>what you can do is to count, and always take the 3rd starting from oldest

why third? I don't get the logic

RamLavi: Sorry I don't get it.. How do we know which version is "the one we officially need to support"? can't we just use that?

RamLavi: In the end we need to agree on the logic "which of the 3 will it choose"

oshoval: lets discuss offline and split to 2 PRs ?

oshoval: https://github.com/kubevirt/cluster-network-addons-operator/pull/2216#discussion_r2026834456

> Removing the KUBEVIRT_PROVIDER set. The nightly should use the default.

Food for thought: 
This can be an idea how to know the suggested provider to use,
we can suggest on kubevirtci, that it will know to use the latest by default, it will know the formula
and also will handle rare corner cases like we discussed

oshoval:
/lgtm
would you able to trigger it to test it please?

unless we are already on latest

RamLavi:
@oshoval once it's merged, I'll run it.
/approve

PR #2120, opened 2025-03-11
Makefile: Enforce max allowed version

oshoval:
In order to protect us from bumping either manually or by Renovate bot, 
go toolchain / lang higher than allowed for stable branches, guard was added to Makefile.

Notes:
1. The guard doesn't protect from bumping higher than allowed Z stream, only X.Y.
2. The guard should be added once the go.mod file has the highest allowed, or at least to know that if you add it as is, you need to update both in case you update go.mod.
3. The guard will disregard toolchain line if such doesn't exists.
4. For simplicity the guard protects the that max normalized (without Z) allowed version match both language level and toolchain, but it is allowed basically that the toolchain will be higher than the lang level on go.mod, so this is extra safty that we might want to remove.

Tried to keep it simple for start, we can later improve upon need.
For example to create a generic snippet that covers all cases, but it is bit more complex,
and keeping stuff simple also has advantage.

**What this PR does / why we need it**:

**Special notes for your reviewer**:

**Release note**:

```release-note
None
```

PR #2220, opened 2025-04-02
[release-0.97] Makefile: Enforce max allowed version

oshoval:
See https://github.com/kubevirt/cluster-network-addons-operator/pull/2120 for more info.
Note that in this PR the check was moved to the end of "make vendor",
so it will detect problems before creating the PR.

**Release note**:

```release-note
None
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,6 +7,8 @@ VERSION_REPLACES ?= 0.97.4
 
 DEPLOY_DIR ?= manifests
 
+MAX_GO_VERSION=1.22
+
 IMAGE_REGISTRY ?= quay.io/kubevirt
 IMAGE_TAG ?= latest
 OPERATOR_IMAGE ?= cluster-network-addons-operator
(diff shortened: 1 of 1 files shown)


oshoval:
@RamLavi can you please take a look ?
it moves the cond to the end of vendoring as we spoke

@ksimon1 maybe Renovate can be configured that if make vendor fails, it wont create the PR ?
btw if so, it would be nice if we could see log of failures of Renovate per CVE, to see if it tried to create, but this is not crucial at all

Thanks

RamLavi (commented):
/lgtm

I'm OK with this change, but please consider a slightly different approach suggested by Quique:
- Starting from main branch
- Using a commented "allowed-go" on the go.mod (closer to the value it guards) instead of env var on the script

Please consider this. Also see [KMP example](https://github.com/k8snetworkplumbingwg/kubemacpool/pull/516)

oshoval:
Thanks, yep saw that, imho go.mod should kept clean, so would prefer the current approach
/approve

oshoval:
> Thanks, yep saw that, imho go.mod should kept clean, so would prefer the current approach

btw another reason is that it saves the requirement of parsing, and also it shows you what version you need to change
for example on other branch "Error: go.mod version 1.21 exceeds max allowed version 1.22"
so you just need to open Makefile and nothing more (well on both cases it is just 1 file)

oshoval:
btw changing the go.mod to have the comment with the max allowed version can lead to conflicts when backporting PRs
which i think makes it a no go

PR #2228, opened 2025-04-06
[release-0.79] makefile: Introduce go version enforcement

oshoval:

Basically we can use go v1.19 in the branch, question if we want to change it for this old branch.

**Release note**:

```release-note
None
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,7 @@ VERSION_REPLACES ?= 0.79.6
 
 DEPLOY_DIR ?= manifests
 
-MAX_GO_VERSION=1.18
+MAX_GO_VERSION=1.19
 
 IMAGE_REGISTRY ?= quay.io/kubevirt
 IMAGE_TAG ?= latest
--- a/Makefile
+++ b/Makefile
@@ -7,6 +7,8 @@ VERSION_REPLACES ?= 0.79.6
 
 DEPLOY_DIR ?= manifests
 
+MAX_GO_VERSION=1.18
+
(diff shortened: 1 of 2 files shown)


oshoval:
@RamLavi @phoracek 
Basically we can use go v1.19 in the branch
Do we want to change it ? (v1.18 atm)

Makefile
```diff
@@ -7,6 +7,8 @@ VERSION_REPLACES ?= 0.79.6
 
 DEPLOY_DIR ?= manifests
 
+MAX_GO_VERSION=1.18
```
RamLavi, line 10: needs to be 1.19

oshoval: https://github.com/kubevirt/cluster-network-addons-operator/pull/2228#issuecomment-2781299233
depends on us, atm it is v1.18, question is if we want to just enforce, or also update 
asking because it is very old branch

RamLavi: ah missed that comment.

IMO we should align to the enforced version, it's never too late :) 
If it breaks something we should know by our U/S (and other) CI lanes.

oshoval: updated

oshoval:
/override pull-e2e-cnao-workflow-k8s-release-0.79

not related to this PR

https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_cluster-network-addons-operator/2228/pull-e2e-cnao-workflow-k8s-release-0.79/1908842132433539072

wondering if we can remove nmstate test in this branch as we did on newer ones

RamLavi (approved):
/lgtm

PR #2232, opened 2025-04-09
Update lifecycle release selector for s390x

aditi-sharma-1:

Full support for the s390x architecture is available starting from CNAO v0.98.2. 

This pull request updates the lifecycle CI pipeline and end-to-end test configurations to use v0.98.2 as the starting version for s390x, so that tests can run successfully. These changes help improve compatibility for the s390x architecture.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/automation/check-patch.e2e-lifecycle-k8s.sh
+++ b/automation/check-patch.e2e-lifecycle-k8s.sh
@@ -28,7 +28,14 @@ main() {
         export E2E_TEST_TIMEOUT=4h
     else
         # Don't run all upgrade tests in regular PRs, stick to those released under HCO
-        export RELEASES_SELECTOR="{0.89.3,0.91.1,0.93.0,0.95.0,99.0.0}"
+        case "$(go env GOARCH)" in
+            s390x)
+                export RELEASES_SELECTOR="{0.98.2,99.0.0}"
+                ;;
+            *)
+                export RELEASES_SELECTOR="{0.89.3,0.91.1,0.93.0,0.95.0,99.0.0}"
+                ;;
+        esac
     fi
 
     make cluster-down
--- a/automation/check-patch.e2e-lifecycle-k8s.sh
+++ b/automation/check-patch.e2e-lifecycle-k8s.sh
@@ -30,7 +30,7 @@ main() {
         # Don't run all upgrade tests in regular PRs, stick to those released under HCO
         case "$(go env GOARCH)" in
             s390x)
-                export RELEASES_SELECTOR="{0.98.2,99.0.0}"
+                export RELEASES_SELECTOR="{0.99.0,99.0.0}"
                 ;;
             *)
                 export RELEASES_SELECTOR="{0.89.3,0.91.1,0.93.0,0.95.0,99.0.0}"
--- a/test/e2e/lifecycle/upgrade_test.go
+++ b/test/e2e/lifecycle/upgrade_test.go
@@ -15,7 +15,7 @@ import (
 const podsDeploymentTimeout = 20 * time.Minute
 
 // CNAO supports multiarch from this release
-const multiArchStartRelease = "0.98.2"
+const multiArchRelease = "0.99.0"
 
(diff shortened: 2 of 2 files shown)


automation/check-patch.e2e-lifecycle-k8s.sh
```diff
@@ -28,7 +28,14 @@ main() {
         export E2E_TEST_TIMEOUT=4h
     else
         # Don't run all upgrade tests in regular PRs, stick to those released under HCO
-        export RELEASES_SELECTOR="{0.89.3,0.91.1,0.93.0,0.95.0,99.0.0}"
+        case "$(go env GOARCH)" in
```
oshoval, line 31: Thanks

instead the change please consider exporting the RELEASES_SELECTOR you want in the 390x lane 
i do understand that you will need to update it from time to time, but it can be nicer
wdyt ?

cc @RamLavi wdyt?

ashokpariya0: Good suggestion!

I think we can approach it like this:

1) Add version 0.98.2 to the list in [check-patch.e2e-lifecycle-k8s.sh](https://github.com/kubevirt/cluster-network-addons-operator/blob/ec5c996cfb835d0ffee7f8a0b6325b1215295ea6/automation/check-patch.e2e-lifecycle-k8s.sh#L31).

if we don't want to do that then we can skip lifecycle for s390x as s390x supported version 0.98.2 is not release under HCO yet

if we are okay with 1 then
For the s390x Prow lane, we can use RELEASES_DESELECTOR env instead of RELEASES_SELECTOR:

https://github.com/kubevirt/cluster-network-addons-operator/blob/ec5c996cfb835d0ffee7f8a0b6325b1215295ea6/test/releases/releases.go#L62



However, one issue I noticed is that 0.98.2 is not currently part of the populated release list. not sure why ???
You can confirm that from the output shown here: https://sharepad.io/p/5L7nfm9

It seems that the version isn't added in the releases initialization:

https://github.com/kubevirt/cluster-network-addons-operator/blob/ec5c996cfb835d0ffee7f8a0b6325b1215295ea6/test/releases/releases.go#L46

So before anything else, we’ll need to ensure 0.98.2 is included in the releases definition.

oshoval: 0.98.2 is just a release on main, so HCO can consume it, since it doesnt belong to a stable branch we should not add it to `RELEASES_SELECTOR`

Please just set `RELEASES_SELECTOR` as you wish in the job, but make sure please that someone maintains it

Possible idea for the future - to think on a way that auto select the 3 latest stable branches that are formal, or at least to have one place where we add it manually in the beginning and from there, helper scripts are populating it to the required place, i.e 390x lane

ashokpariya0: just setting export RELEASES_SELECTOR=0.98.2 is not working, it is failing at below check, as explained earlier

https://github.com/kubevirt/cluster-network-addons-operator/blob/ec5c996cfb835d0ffee7f8a0b6325b1215295ea6/test/releases/releases.go#L46

Setting RELEASES_SELECTOR alone won't work until we add 0.98.2 to the populated releases list.

oshoval: we can't add it 
we add only latest patch on stables and this isn't stable

can you please try to think about other way / debug it deeper ?

oshoval: @RamLavi wdyt ?

ashokpariya0: okay, noted.

aditi-sharma-1:
Hi @ashokpariya0, 
Sure, will test once with 0.99.0 version and update here accordingly. Thank you.

aditi-sharma-1:
Hi @ashokpariya0,
I tested with v0.99.0 locally on the s390x cluster, and it’s building successfully. Thank you.

PR #2239, opened 2025-04-23
tests: Check for configMap deploy and removal

### LINKED ISSUE #2202: Flake: [It] NetworkAddonsConfig when kubeMacPool is deployed should modify the MAC range after being redeployed
**What happened**:
test [failed](https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_cluster-network-addons-operator/2195/pull-e2e-cluster-network-addons-operator-workflow-k8s/1906594309021372416).

It looks like kubemacpool's cert-manager pod cannot work properly because kubemacpool-manager-role is not deplopyed ([logs](https://storage.googleapis.com/kubevirt-prow/pr-logs/pull/kubevirt_cluster-network-addons-operator/2195/pull-e2e-cluster-network-addons-operator-workflow-k8s/1906594309021372416/artifacts/1_kubemacpool-cert-manager-6896f74757-cfk6x_manager.log)). 

On the other hand, CNAO seems to redeploy it [repeatedly](https://storage.googleapis.com/kubevirt-prow/pr-logs/pull/kubevirt_cluster-network-addons-operator/2195/pull-e2e-cluster-network-addons-operator-workflow-k8s/1906594309021372416/artifacts/1_cluster-network-addons-operator-5dc85dcc7-c7mqp_cluster-network-addons-operator.log):
```
2025/03/31 06:57:25 reconciling (rbac.authorization.k8s.io/v1, Kind=ClusterRole) /kubemacpool-manager-role
2025/03/31 06:57:25 does not exist, creating (rbac.authorization.k8s.io/v1, Kind=ClusterRole) /kubemacpool-manager-role
```

Steps to reproduce the behavior.

**Additional context**:

**Environment**:
- KubeVirt version (use `virtctl version`): N/A
- Kubernetes version (use `kubectl version`): N/A
- VM or VMI specifications: N/A
- Cloud provider or hardware configuration: N/A
- OS (e.g. from /etc/os-release): N/A
- Kernel (e.g. `uname -a`): N/A
- Install tools: N/A
- Others: N/A

RamLavi:
When running a CNAO CR, CNAO creates a prev config of the last applied configuration, in order to ensure persistent reconfiguration in case a re-install took place on an existing CR. For example, if kubemacpool was redeployed, it will remember the ranges.

This however creates a test race when checking that removing the CNAO CR recreates the kubemacpool ranges when redeployed after CR deletion. 

This PR is adding a check to ensure the configmap was removed after CR deletion, in order to fix this race.

Fix #2202 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/test/check/check.go
+++ b/test/check/check.go
@@ -478,6 +478,10 @@ func checkForComponent(component *Component) error {
 		errsAppend(checkForNetworkAttachmentDefinition(component.NetworkAttachmentDefinition))
 	}
 
+	if component.ConfigMap != "" {
+		errsAppend(checkForConfigMap(component.ConfigMap))
+	}
+
 	return errsToErr(errs)
 }
 
@@ -531,6 +535,10 @@ func checkForComponentRemoval(component *Component) error {
 
 	if component.NetworkAttachmentDefinition != "" {
 		errsAppend(checkForNetworkAttachmentDefinitionRemoval(component.NetworkAttachmentDefinition))
+	}
+
+	if component.ConfigMap != "" {
+		errsAppend(checkForConfigMapRemoval(component.ConfigMap))
 	}
 
 	return errsToErr(errs)
@@ -801,6 +809,21 @@ func checkForNetworkAttachmentDefinition(name string) error {
 	}
 
 	err = checkRelationshipLabels(networkAttachmentDefinition.GetLabels(), "NetworkAttachmentDefinition", name)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func checkForConfigMap(name string) error {
+	configMap := corev1.ConfigMap{}
+	err := testenv.Client.Get(context.Background(), types.NamespacedName{Name: name, Namespace: components.Namespace}, &configMap)
+	if err != nil {
+		return err
+	}
+
+	err = checkRelationshipLabels(configMap.GetLabels(), "ConfigMap", name)
 	if err != nil {
 		return err
 	}
--- a/test/check/check.go
+++ b/test/check/check.go
@@ -73,6 +73,11 @@ func CheckComponentsRemoval(components []Component) {
 			return checkForComponentRemoval(&component)
 		}, 5*time.Minute, time.Second).ShouldNot(HaveOccurred(), fmt.Sprintf("%s component has not been fully removed within the given timeout", component.ComponentName))
 	}
+
+	By(fmt.Sprintf("Checking the %s configmap has been removed", names.AppliedPrefix+names.OperatorConfig))
+	Eventually(func() error {
+		return checkForConfigMapRemoval(names.AppliedPrefix + names.OperatorConfig)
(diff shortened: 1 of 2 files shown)


RamLavi:
Fixes https://github.com/kubevirt/cluster-network-addons-operator/issues/2202

RamLavi:
not working. closing.

RamLavi:
> not working. closing.

Reopening - found another case and fixed it.

RamLavi:
@oshoval can you take a look?

test/check/check.go
```diff
@@ -808,6 +816,21 @@ func checkForNetworkAttachmentDefinition(name string) error {
 	return nil
 }
 
+func checkForConfigMap(name string) error {
```
oshoval, line 819: Since this template repeats, potential follow up (didn't validate)
wdty ?

```
func checkForObject(name, kind, namespace string, obj client.Object) error {
	err := testenv.Client.Get(context.Background(), types.NamespacedName{Name: name, Namespace: namespace}, obj)
	if err != nil {
		return err
	}

	err = checkRelationshipLabels(obj.GetLabels(), kind, name)
	if err != nil {
		return err
	}

	return nil
}
```

test/check/check.go
```diff
@@ -73,6 +73,11 @@ func CheckComponentsRemoval(components []Component) {
 			return checkForComponentRemoval(&component)
 		}, 5*time.Minute, time.Second).ShouldNot(HaveOccurred(), fmt.Sprintf("%s component has not been fully removed within the given timeout", component.ComponentName))
 	}
+
+	By(fmt.Sprintf("Checking the %s configmap has been removed", names.AppliedPrefix+names.OperatorConfig))
```
oshoval, line 77: <aybe worth to have its own function please ?
it isn't a component

Potential FUP (follow-up) to think about
maybe we can generalize `CheckComponentsRemoval`, that it scan according ownerRef / relation labels what should be deleted ? (not crucial atm i guess)

oshoval (commented):
Thanks

Can this case affect real user, where he doesn't have this improvement that now the test has ?
If so, we might want to think how reconcile should make sure the old objs are deleted, before new ones are created / signaling that we done upgrade

test/check/check.go
RamLavi: Let's think about that in  a follow up

RamLavi:
> Thanks
> 
> Can this case affect real user, where he doesn't have this improvement that now the test has ? If so, we might want to think how reconcile should make sure the old objs are deleted, before new ones are created / signaling that we done upgrade

It very unlikely. This scenario is a install+remove CR, not upgrade. 
install & upgrade are the important feature that is important for our customers, and they are handled in lifecycle lane. 
I would not make CNAO deployment more complex if not really needed.

qinqon (approved):
/lgtm
/approve

PR #2240, opened 2025-04-28
pkg: Fix style and formatting issues

RamLavi:
This PR is fixing minor formatting issues:
- export function documentation
- unit test yaml validity
- error formatting style (do not start with capital letter)

**Release note**:
```release-note
NONE
```

### DIFF
--- a/pkg/controller/statusmanager/status_manager.go
+++ b/pkg/controller/statusmanager/status_manager.go
@@ -190,7 +190,7 @@ func (status *StatusManager) set(reachedAvailableLevel bool, conditions ...condi
 	// Patch NetworkAddonsConfig's status
 	err = status.client.Status().Patch(context.TODO(), config, patch)
 	if err != nil {
-		return fmt.Errorf("Failed to patch NetworkAddonsConfig %q Status: %v", config.Name, err)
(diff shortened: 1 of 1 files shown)


pkg/names/names.go
```diff
@@ -17,6 +17,7 @@ const PROMETHEUS_LABEL_KEY = "prometheus.cnao.io"
 const PROMETHEUS_LABEL_VALUE = "true"
 
 // Relationship labels
+
 const COMPONENT_LABEL_KEY = "app.kubernetes.io/component"
 const PART_OF_LABEL_KEY = "app.kubernetes.io/part-of"
 const VERSION_LABEL_KEY = "app.kubernetes.io/version"
```
nunnatsa, line 23: According golang convention, we should use camelCase rather than upper snake case. e.g.
```suggestion
const VersionLabelKey = "app.kubernetes.io/version"
```

nunnatsa:
Added one inline comment, not exactly related to the current PR.

pkg/names/names.go
RamLavi: Yes, I plan this for another PR :)

PR #2243, opened 2025-04-28
Fix vendoring of golang.org/x/tools/cmd/goimports

phoracek:

**Release note**:

```release-note
NONE
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,36 +1,37 @@
 # Temporary generated files
-_out
+/_out
+/tools/_out
 
 # Temporary build files
-build/operator/_output
-build/operator/_test
-build/_output
+/build/operator/_output
+/build/operator/_test
+/build/_output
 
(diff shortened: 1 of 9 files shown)


.gitignore
```diff
@@ -1,36 +1,37 @@
 # Temporary generated files
-_out
+/_out
```
oshoval, line 2: why the whole file is prefixed with `/` ?
best to just remove the lines we don't need related to goimport in this PR please?

phoracek: I did it so we won't unknowingly won't miss any accidentally matched files. Why is it better to keep this limited to the one dependency that we just happened to find broken?

oshoval: looking again, you are right
thanks

oshoval:
Thanks
/lgtm
/approve

PR #2249, opened 2025-04-29
[release-0.89] Fix vendoring of golang.org/x/tools/cmd/goimports

phoracek:

**Release note**:

```release-note
NONE
```


oshoval:
Thanks
/lgtm
/approve

PR #2254, opened 2025-04-29
[release-0.85] Fix vendoring of golang.org/x/tools/cmd/goimports

phoracek:

**Release note**:

```release-note
NONE
```


RamLavi:
/override pull-e2e-cnao-workflow-k8s-release-0.85

this test should not be in this stable branch. Will remove it in different PR

RamLavi:
@phoracek FWIW, this stable branch is not supported.

phoracek:
@RamLavi not by itself, but in order to upgrade from the previous stable to the next, the upgrade path goes through this one. So we need to continue keeping it tidy.

PR #2257, opened 2025-05-02
Fix nightly build

nunnatsa:
At the end of the nightly build, the automation/test-nightly-build.sh script copies the artifact files to a gcp bucket using the gsutil. The script, as currently implemented, copies the files to the a file named the build date, overriding the last copied file, instaed of to a directory with the name of the build date.

e.g.
```shell
curl https://storage.googleapis.com/kubevirt-prow/devel/nightly/release/kubevirt/cluster-network-addons-operator/20250502
```
returns the network-addons-config-example.cr.yaml file, while
```shell
curl https://storage.googleapis.com/kubevirt-prow/devel/nightly/release/kubevirt/cluster-network-addons-operator/20250502/network-addons-config-example.cr.yaml
```
returns 404.

And all the other files are not there, because they were overrode by the last copied file.

This PR fixes the `gsutil cp` command to work as expected.

In addition, the PR overrides the ["latest" file](https://storage.googleapis.com/kubevirt-prow/devel/nightly/release/kubevirt/cluster-network-addons-operator/latest), only if all the artifacts were copied to the new bucket, successfully.

**Release note**:

```release-note
None
```

### DIFF
--- a/automation/test-nightly-build.sh
+++ b/automation/test-nightly-build.sh
@@ -42,11 +42,11 @@ main() {
     CNAO_VERSION=$(hack/version.sh)
     SRC_DIR="./_out/cluster-network-addons/${CNAO_VERSION}"
     DEST="gs://${cnao_bucket}/${build_date}"
-    gsutil cp "${SRC_DIR}/namespace.yaml" "${DEST}"
-    gsutil cp "${SRC_DIR}/cluster-network-addons-operator.${CNAO_VERSION}.clusterserviceversion.yaml" "${DEST}"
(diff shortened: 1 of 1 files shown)


RamLavi:
@nunnatsa why doesn't it work the same as linux when doing it locally?
```
$ mkdir ram
$ touch 1
$ cp $PWD/1 ram
$ ls ram/
1
```

RamLavi (approved):
Well might as well be explicit with the cp when using gsutil then. 
/lgtm
/approve

PR #2263, opened 2025-05-07
Fix the nightly build test

nunnatsa:

The nightly build pushes the wrong manifest files to the remote storage: it pushes the manifests of the latest release instead of the ones from the nightly build.

This PR fixes the issue by using
`make docker-build-operator docker-push-operator` instead of `cluster-operator-push`, that hard codes the image registry to a registry in a local kubevirtci cluster.

Also, the nightly build "version" is now with the value of the image tag, so the manifests will be unique, and visually linked to the image.

```release-note
None
```


RamLavi (approved):
makes sense

/lgtm
/approve

PR #2268, opened 2025-05-08
[WIP] best-practice check

RamLavi:
This PR introduces a best-practice check for CNAO operands
If a component/operand breaks the best practice code - the unit tests lane should fail.
 
**Release note**:

```release-note
NONE
```


RamLavi:
all pods are running with the best practice approach. 
However I don't think this e2e test is worth the CPU time it takes.
Closing this PR

PR #2331, opened 2025-06-03
Linux bridge:  Harden SCC by restricting SELinux context and allowed volume types

RamLavi:
This pull request improves the security posture of the linux-bridge SecurityContextConstraints (SCC) by introducing two changes:

1. **Restrict SELinux context** to `MustRunAs` with `spc_t` type.
2. **Restrict allowed volume types** to only `hostPath`, `configMap`, and `secret`.

To anticipate reviewers request, [commented](https://github.com/kubevirt/cluster-network-addons-operator/pull/2331#issuecomment-2943335474) on the PR why I did NOT remove other SCCs.

**Release note**:

```release-note
reduce linux bridge SCC
```

### DIFF
--- a/data/linux-bridge/002-linux-bridge.yaml
+++ b/data/linux-bridge/002-linux-bridge.yaml
@@ -69,11 +69,16 @@ spec:
               memory: "15Mi"
           securityContext:
             privileged: true
+            readOnlyRootFilesystem: true
           volumeMounts:
             - name: cnibin
               mountPath: /opt/cni/bin
+            - name: tmp
+              mountPath: /tmp
           terminationMessagePolicy: FallbackToLogsOnError
       volumes:
         - name: cnibin
           hostPath:
             path: {{ .CNIBinDir }}
+        - name: tmp
+          emptyDir: { }
--- a/data/linux-bridge/001-rbac.yaml
+++ b/data/linux-bridge/001-rbac.yaml
@@ -16,7 +16,7 @@ allowHostIPC: false
 allowHostNetwork: false
 allowHostPID: false
 allowHostPorts: false
-readOnlyRootFilesystem: false
+readOnlyRootFilesystem: true
 runAsUser:
   type: RunAsAny
 seLinuxContext:
@@ -29,4 +29,5 @@ volumes:
   - hostPath
   - configMap
   - secret
+  - emptyDir
 {{ end }}
--- a/data/linux-bridge/001-rbac.yaml
+++ b/data/linux-bridge/001-rbac.yaml
@@ -20,7 +20,9 @@ readOnlyRootFilesystem: false
 runAsUser:
   type: RunAsAny
 seLinuxContext:
-  type: RunAsAny
+  type: MustRunAs
+  seLinuxOptions:
+    type: spc_t
 users:
 - system:serviceaccount:{{ .Namespace }}:linux-bridge
 volumes:
--- a/data/linux-bridge/001-rbac.yaml
+++ b/data/linux-bridge/001-rbac.yaml
@@ -26,5 +26,7 @@ seLinuxContext:
 users:
 - system:serviceaccount:{{ .Namespace }}:linux-bridge
 volumes:
-- "*"
+  - hostPath
+  - configMap
+  - secret
 {{ end }}


RamLavi:
/hold
I want to check this on kcli locally to make sure it works

RamLavi:
## 🔒 SCC Non-Default Capabilities Review

| **Capability**                  | **Set To**            | **Default**          | **Reason** |
|:----------------------------------|:----------------------|:---------------------|:-----------|
| `allowPrivilegedContainer`       | `true`                | `false`              | Needed to access and write to root-owned hostPath directories (`/opt/cni/bin`, `/var/lib/cni/bin`). Without this, the container would get `permission denied` errors. |
| `allowHostDirVolumePlugin`       | `true`                | `false`              | Needed to mount `hostPath` volumes for CNI plugin installation on the host filesystem. |
| `readOnlyRootFilesystem`         | `true`                | `false`              | Similarly to restricted-v2, harden the container by preventing write access to the root filesystem. Mitigates potential runtime filesystem tampering. |
| `runAsUser`                      | `RunAsAny`             | `MustRunAsRange`     | Container must run as UID 0 (root) to perform `cp` and `ln` operations into hostPath volumes, which are owned by root. |
| `seLinuxContext`                  | `MustRunAs` with `spc_t` | `MustRunAs`         | Privileged containers must use `spc_t` SELinux type in OpenShift to access the host securely while maintaining SELinux enforcement. |
| `volumes`                         | `hostPath`, `configMap`, `secret` | Only safe types (no hostPath) | Allow only necessary volume types. Avoids wildcard (`*`) volume allowance to reduce attack surface. |

---

### 📚 Sources:
- [Kubernetes Pod Security Standards](https://kubernetes.io/docs/concepts/security/pod-security-standards/)
- [OpenShift SCC Best Practices](https://docs.openshift.com/container-platform/4.10/authentication/managing-security-context-constraints.html)
- [Red Hat: Container Security Guide](https://access.redhat.com/documentation/en-us/openshift_container_platform/4.10/html/security_and_compliance/container-security#security-hardening-containers_container-security)
- [CIS Kubernetes Benchmark](https://www.cisecurity.org/benchmark/kubernetes)
- [Red Hat Best Practices: emptyDir Mounts for /tmp](https://redhat-best-practices-for-k8s.github.io/guide/#k8s-best-practices-storage:-emptydir)

data/linux-bridge/002-linux-bridge.yaml
```diff
@@ -69,11 +69,16 @@ spec:
               memory: "15Mi"
           securityContext:
             privileged: true
+            readOnlyRootFilesystem: true
           volumeMounts:
             - name: cnibin
               mountPath: /opt/cni/bin
+            - name: tmp
+              mountPath: /tmp
```
maiqueb, line 77: why do we need access to /tmp ? Seems we didn't need it before. Why add it now ?

RamLavi: From the commit message:
```
Since some system utilities (e.g., bash, cp, sha256sum) may require a
writable /tmp directory, mount an emptyDir at /tmp to provide a writable
scratch space, following container security best practices [0].
Also update the associated SCC to enforce readOnlyRootFilesystem: true
at the policy level.

[0]
https://redhat-best-practices-for-k8s.github.io/guide/#k8s-best-practices-storage:-emptydir
```

RamLavi: IOW, if the files are too big they sometimes write to tmp. To be safe - mount an empty tmp folder, so that it won't explode should some cp decides to use that folder..

RamLavi:
checked on kcli that linux bridge is properly deployed. compared binaries checksum before and after the SCC change in order to verify they are indeed the same.

/hold cancel

RamLavi:
> Once we return the DaemonSet of passt this way or another, we will need to do more or less the same there right ?
> 
> https://github.com/kubevirt/ipam-extensions/pull/50/files#diff-ad4bb703999139ceb1971eca665b98b606f15db349ef5372346405b65a18b458
> 
> as it also does more or less what linux-bridge container does

pretty much, yes. All containers should adhere to scc best practices, and it will probably have similar considerations..

maiqueb:
/hold

@RamLavi did you check this works in kcli ? 
If so / once you do, feel free to remove the hold.

Thank you.

RamLavi:
> /hold
> 
> @RamLavi did you check this works in kcli ? 
> If so / once you do, feel free to remove the hold.
> 
> Thank you.
> 

As [mentioned](https://github.com/kubevirt/cluster-network-addons-operator/pull/2331#issuecomment-2943697230) in the comment above, yes.

/hold cancel

PR #2341, opened 2025-06-10
Opt-in for network-policies provided by hco-bundle for CNAO and its components pods

ormergi:
Enable CNAO operator and its operands pors operate when network restrictions in form of defaultl deny-all network-policy is in place, when managed by HCO.

The hco-bundle should provide network-policies allowing its operands work when network restrictions are in place.
In order to opt-in pods for the provided network-policies pods should be labeled with the following labels:
1. `hco.kubevirt.io/allow-access-cluster-services`
    Allow access the cluster API and DNS.
    This is a fundamental requirement for a k8s controller to oprate, it is requiremd for the following compoents:
    - kubemacpool
    -  ipam-extentions
    - kube-secondary-dns
2. `hco.kubevirt.io/allow-prometheus-access` 
    Allow Prometheus pods ingress the metrics endpoint
    Required by CNAO opreator pods, allowing promethues scrape CNAO metrics.

With this change the project `manifest-templator` and `csv-gen` tool should produce the project Deployment and CSV with the above labels. 
Allowing CNAO to operate under network restrictions when managed by HCO.

In addition CNAO should generate and install kubemacpool, kubevirt-ipam-contoller and kube-secondary-dns with the above labels.
Allowing them to operate under network restrictions when managed by CNAO and HCO.

**Release note**:

```release-note
CNAO operator and its components opt-in for network-policies provided by hco-bundle
```

### DIFF
--- a/hack/components/bump-kubemacpool.sh
+++ b/hack/components/bump-kubemacpool.sh
@@ -59,6 +59,15 @@ patches:
   target:
     version: v1
     kind: Namespace
+- path: add-pod-template-label-allow-access-cluster-services_patch.yaml
+  target:
+    version: v1
+    kind: Deployment
+- path: add-pod-template-label-allow-prometheus-access_patch.yaml
+  target:
+    version: v1
+    kind: Deployment
+    name: mac-controller-manager
 EOF
 
     cat <<EOF > config/cnao/cnao_kubemacpool_manager_patch.yaml
@@ -145,6 +154,16 @@ EOF
   path: /metadata/labels
 EOF
 
+    cat <<EOF > config/cnao/add-pod-template-label-allow-access-cluster-services_patch.yaml
+- op: add
+  path: /spec/template/metadata/labels/hco.kubevirt.io~1allow-access-cluster-services
+  value: ""
+EOF
+    cat <<EOF > config/cnao/add-pod-template-label-allow-prometheus-access_patch.yaml
+- op: add
+  path: /spec/template/metadata/labels/hco.kubevirt.io~1allow-prometheus-access
+  value: ""
+EOF
 
     (
         cd config/cnao
--- a/data/kube-secondary-dns/secondarydns.yaml
+++ b/data/kube-secondary-dns/secondarydns.yaml
@@ -73,6 +73,7 @@ spec:
     metadata:
       labels:
         k8s-app: secondary-dns
+        hco.kubevirt.io/allow-access-cluster-services: ""
       annotations:
         kubectl.kubernetes.io/default-container: status-monitor
         openshift.io/required-scc: "restricted-v2"
--- a/data/kubemacpool/kubemacpool.yaml
+++ b/data/kubemacpool/kubemacpool.yaml
@@ -153,6 +153,7 @@ spec:
         app: kubemacpool
         control-plane: cert-manager
         controller-tools.k8s.io: "1.0"
+        hco.kubevirt.io/allow-access-cluster-services: ""
     spec:
       affinity: {{ toYaml .Placement.Affinity | nindent 8 }}
       containers:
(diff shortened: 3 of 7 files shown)


ormergi:
@nunnatsa could you please confirm this PR adds the correct labels?

oshoval:
Thanks nice
Naive question
what about the other operands ?

If you want to check it upstream at least one time, you can deploy HCO with the logic and the updated CNAO with all the operands, and then run the CNAO lanes for each operand that runs the operand e2e tests
Imho we need this at least once (and to document the steps, preferably that allow to run this on CI, in case we will need it again)

One can argue that those labels should be on each component instead, but i wont insist on it.
Please note that if we ditch CNAO (atm not known yet), this of course will need to be moved for each component,
while not sufficient condition, it just shows another benefit of putting it on the component themselves.

oshoval:
thinking about it more

we need the following:
1. know if pods are missing as Nahshon said
2. know to test it
3. know to have something that in the future tells us if something is missing

maybe you can use the relation labels, i.e for each pod that has a label managed_by = cnao + cnao itself
should have those labels ?
would this solve all the above (assuming of course relation labels are fine, but they are tested pretty good continuously)

hack/components/bump-kubemacpool.sh
```diff
@@ -59,6 +59,15 @@ patches:
...
     kind: Namespace
+- path: add-pod-template-label-allow-access-cluster-services_patch.yaml
+  target:
+    version: v1
+    kind: Deployment
+- path: add-pod-template-label-allow-prometheus-access_patch.yaml
+  target:
+    version: v1
+    kind: Deployment
+    name: mac-controller-manager
```
qinqon, line 70: Are you sure you cannot inline the json patch with kubebuilder things instead of generating files and apply them ?

ormergi: kubebuilder does provide generating NetworkPolicy for a project webhooks and metric endpoint using markers [[1] ](https://github.com/kubernetes-sigs/kubebuilder/pull/3853/files)(the +kubebuilder annotation), although it seems documnation about it lacking, it requires feasibility check.
Unfortunately kubemapool repo doesnt make use of kubebulder, it use plain controller-gen, which doesnt provider this functionality.
Transitioning the project to use kubebuilder has cost, naturally. I hope we can do it later on.

Note that in the context of this repo, the automation takes the released kubemacpool installtion manifest, generates a template form it and embed the template in CNAO.
Such changes as generating the network-policy using other tooling should be done in kubemacpool repo.

ormergi:
@oshoval 
1. This PR and previous work focuses on providing NP for kubemacpool, ipam-ext and kube-secondary-dns, which the provdied NPs covers the necessary pods (and only the pods that require in/egress configurations).
    CNIs and bridge-makers (linux-bridge, macvtap, multus, ovs-cni) runs in the host network namespace, which NetworkPolicies does not support at the moment (by design), thus no-op for these components.
   As for the dynamic-network-controller, I hope we can do in a follow-up PR upon necessity.
2. The project CI does not test CNAO installation by HCO, as expected, thus I think we cannot cover what the labels trigger.
    This functionality is expected to be tested on HCO repo. It already have e2e tests covering CNAO installtion.
3.  In case some new endpoint is exposed by some component, or new component is added, it should provide the necessary NPs so it can operate under network restrictions.
    What I had in mind is to add a default-deny-all NetworkPolicy in this repo CI test jobs, for having test coverage on NPs.
    With the deny-all NP in place in case some component missing a NP, I expect the relevant lane to fail.
    In case the NP is missing and the job pass, tests doesnt cover what they should, or not tested at all. 

> maybe you can use the relation labels, i.e for each pod that has a label managed_by = cnao + cnao itself should have those labels ? would this solve all the above (assuming of course relation labels are fine, but they are tested pretty good continuously)

I am afraid it will result with some pods get affected by some NPs where they shouldn't.
Also, the provided NPs by HCO are note aware of such details as the label that is used by the operand for marking related workloads/resources.

oshoval:
Thanks analysis and [3] sgtm

qinqon (approved):
/lgtm
/approve

PR #2382, opened 2025-07-28
manifest-templator: Use valid string format

ormergi:
The 'make check' runs go vet which fails with the following error:
   tools/manifest-templator/manifest-templator.go:297:14: non-constant format string in call to fmt.Printf

Replace with Print() because no string formatting is needed.

**Release note**:

```release-note
NONE
```


qinqon (approved):
/lgtm
/approve

PR #2398, opened 2025-08-11
network-policy: Install network-policy for operands

ormergi:
Following recent work around network-policies, we planned to rely on network-policies provided by HCO bundle.
Unfortunately, HCO will not provide the expected network-policies in next release.

Change the operator to install the following network-policies instead:
1. Allow egress to cluster API and DNS Opt-in for policy using label "allow-access-cluster-services"
2. Allow ingress to metric endpoints Opt-in for policy using label "allow-access-to-metrics-endpoints"

Please note the affected components remain as the same although the label name changes.

Once HCO provide the mentioned policies as expected, we can drop this PR changes and label the relevant pods with HCO opt-in for policy labels as initially planned.

**Release note**:

```release-note
CNAO install network-policy for operands allowing access to cluster API, DNS and allow ingress to metric endpoints.
```


data/kube-secondary-dns/secondarydns.yaml
```diff
@@ -73,7 +73,7 @@ spec:
     metadata:
       labels:
         k8s-app: secondary-dns
-        hco.kubevirt.io/allow-access-cluster-services: ""
+        allow-access-cluster-services: ""
```
oshoval, line 76: please see here about prefix that is nice to have
https://github.com/kubevirt/hyperconverged-cluster-operator/commit/a77e454da7d1200fa4d8007045cba64dfde66d9d

pkg/network/network-policy.go
```diff
@@ -0,0 +1,33 @@
...
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+
+	"github.com/kubevirt/cluster-network-addons-operator/pkg/render"
+)
+
+func RenderNetworkPolicy(manifestDir string, info ClusterInfo) ([]*unstructured.Unstructured, error) {
+	data := render.MakeRenderData()
+	data.Data["Namespace"] = os.Getenv("OPERAND_NAMESPACE")
+	if info.OpenShift4 {
+		data.Data["ClusterDNSNamespace"] = "openshift-dns"
```
oshoval, line 17: this wont work on OKD
kubevirt-hyperconverged is the ns there

i am also not sure we are allowed to mention this ns here

what about using the ns where CNAO is deployed instead line 17 (under same condition)

pkg/network/network.go
```diff
@@ -178,6 +178,12 @@ func Render(conf *cnao.NetworkAddonsConfigSpec, manifestDir string, openshiftNet
 	}
 	objs = append(objs, o...)
 
+	o, err = RenderNetworkPolicy(manifestDir, *clusterInfo)
```
oshoval, line 181: what about removal please ?
unless thanks to relation labels (need to make sure anyhow they are updated please) it removes those once cnao is undeployed

can you please test different scenarios ?
i.e if HCO deploys it and removes CNAO, this should be removed as well
(there might be more scenarios)

if you find a way to simulate it or to see in the code / created objects that we are fine, it might be good enough

oshoval (commented):
Thanks
first pass

Please consider checking on kcli, unless you feel comfortable

pkg/network/network-policy.go
oshoval: ye corrected myself

anyhow it might be fine in that aspect
https://docs.okd.io/latest/networking/networking_operators/dns-operator.html

data/kube-secondary-dns/secondarydns.yaml
oshoval: i meant of course to use prefix of cnao,
but yes you are right
thanks

pkg/network/network-policy_test.go
```diff
@@ -0,0 +1,129 @@
...
+
+const (
+	optInLabelAllowEgressClusterServices  = "allow-access-cluster-services"
+	optInLabelAllowIngressMetricEndpoints = "allow-access-metrics-endpoints"
+)
+
+var netPolTypeMeta = metav1.TypeMeta{
+	APIVersion: "networking.k8s.io/v1",
+	Kind:       "NetworkPolicy",
+}
```
oshoval, line 56: please move to beginning of the file

ormergi: Note that these are used by the testNetworkPolices helpers and not in the actual test code

oshoval: of course, but it is still nicer to have it in the ut file itself in the beginning and not in the middle please

ormergi: Done

pkg/network/network.go
ormergi: It is done by [[1]](https://github.com/kubevirt/cluster-network-addons-operator/blob/main/pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go#L383)  for all rendered objects (except for the project  Namespace)

oshoval:
What about installing those NPs only on OCP ?
Upstream K8s won't have deny NPs

This was also the behavior when HCO deployed it, since CNAO standalone didn't install those
(i wonder if HCO standalone did, or the plan was that just OLM will)

Just the upstream tests can install it ad hoc as it was before

Wdyt ?
cc @RamLavi

ormergi:
There will be no e2e coverage because this repo test against valina k8s only.
We wont have early feedback on bad changes on CI (e.g.: endpoint port change, new endpoint is added..)

The plan is to have these NPs provided by HCO bundle (not the operator pods), currently its impossible thus defered to next release.

I hope we can continue with the current approach and change later if necessary.

ormergi:
Verified on KCLI cluster, network-policy work as expected all CNAO pods are healthy

PR #2422, opened 2025-09-18
tier1/kubemacpool: Replace ginkgo skip with label filtering

RamLavi:
Currently tier kubemacpool test is skipping the tests according to the opt mode set on the kubemacpool bumper script.

Using programmatic skip is not a good practice, as it is easy to assume tests passed when they were actually skipped, and the current approach of flexibly skipping according to the bumper script makes it worse. We should explicitly state what the tests should run by using ginkgo label filtering.
Following a [PR](https://github.com/k8snetworkplumbingwg/kubemacpool/pull/568) on KMP that adds ginkgo labels for opt-mode cases, this PR is replacing ginkgo label filtering to predictably state the tests expected to run.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/automation/check-patch.e2e-kubemacpool-functests.sh
+++ b/automation/check-patch.e2e-kubemacpool-functests.sh
@@ -9,23 +9,6 @@
 # yum -y install automation/check-patch.packages
 # automation/check-patch.e2e-kubemacpool-functests.sh
 
-function __get_skipped_tests() {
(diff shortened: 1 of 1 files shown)


RamLavi:
/test pull-e2e-cluster-network-addons-operator-monitoring-k8s
maybe a hiccup

PR #2428, opened 2025-10-16
[main] Release v0.101.0-rc0

RamLavi:
Release v0.101.0-rc0

Release is issued so that it could be consumed by [HCO](https://github.com/kubevirt/hyperconverged-cluster-operator) to use kubemacpool range day2 feature.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@ all: fmt check
 VERSION ?= 99.0.0
 export VERSION := $(VERSION)
 # Always keep the last released version here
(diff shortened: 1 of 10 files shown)


oshoval:
not first time (2nd :) ), but i also think we dont need it, 
moreover it requires the next time we release to do it manually right ? or the script were fixed in that aspect ?

RamLavi:
reason of lifecycle lane failure is flannel not having enough IPs per node. 
@oshoval FYI

PR #2455, opened 2025-11-06
vendor: Reduce kubevirt dependencies

RamLavi:
This PR replaces kubevirt.io/kubevirt test library imports with local implementations from test/libvmi and test/libframework.

Virtctl client is replaced by controller runtime client.
test namespace in monitoring lane is changed to `cnao-test`

This removes the dependency with kubevirt/go-client and makes the vendor much more easy to maintain.

**Release note**:
```release-note
NONE
```


ormergi (approved):
Thanks for the cleanup, its funny that kubevirt client was merly used for fetch some service account, good catch!

I a little bit confused by the first two commits, they seem not related to kubevirt client-go removal.
Maybe the title should say something like "reduce kubevirt dependencies".

One last thing, consider leaving kubevirt's libvmi, I think it make more sense to consume it then implement it on this project, although not likely, the implementation may diverge. The original libvmi is widely used in kubevirt/kubevirt and well tested.

RamLavi:
> Thanks for the cleanup, its funny that kubevirt client was merly used for fetch some service account, good catch!
> 
> I a little bit confused by the first two commits, they seem not related to kubevirt client-go removal. Maybe the title should say something like "reduce kubevirt dependencies".

Yeah you're right, I changed the PR title like you suggested.

> 
> One last thing, consider leaving kubevirt's libvmi, I think it make more sense to consume it then implement it on this project, although not likely, the implementation may diverge. The original libvmi is widely used in kubevirt/kubevirt and well tested.

I think we should keep it separated if possible. We don't really need their libvmi.

RamLavi:
@maiqueb this will allow us to bump k8s vendor much faster.

RamLavi:
/test pull-e2e-cluster-network-addons-operator-lifecycle-k8s

Should pass after #2466 is merged

ormergi (approved):
LGTM

Please consider reflect the test namespaces has been changed, at least in the PR description, its not trivial change.
Placing hold to give change for considering it. Feel free to remove once resolved (or not..)
/hold

PR #2465, opened 2025-11-09
[release-0.95] Pin go and kubevirt versions to fit stable branch

RamLavi:
Currently the go used is the latest from google site. However this now
breaks on stable branches.
Also the kubevirt version uses is the latest from kubevirt repo.

This PR changes install-go.sh script logic to:
- use Go version from the toolchain first; if unavailable, fall back to
the regular Go version.
- if the chosen version lacks a patch (.z), append `.0`.

This PR also changes deploy-kubevirt.sh script logic to:
- use the latest patch of the supported x.y of this stable branch.

**Release note**:
```release-note
NONE
```

### DIFF
--- a/hack/deploy-kubevirt.sh
+++ b/hack/deploy-kubevirt.sh
@@ -1,9 +1,14 @@
 #!/bin/bash
 set -exo pipefail
 
+function getLatestPatchVersion {
+  local major_minors=$1
+  curl -s https://api.github.com/repos/kubevirt/kubevirt/releases?per_page=100 | grep .tag_name | grep ${major_minors} | sort -V | tail -1 | awk -F':' '{print $2}' | sed 's/,//' | xargs
+}
+
 if [ -z "${KUBEVIRT_VERSION}" ];then
   # Get latest stable KubeVirt version
-  export KUBEVIRT_VERSION=$(curl -s https://api.github.com/repos/kubevirt/kubevirt/releases | grep tag_name | grep -v -- - | sort -V | tail -1 | awk -F':' '{print $2}' | sed 's/,//' | xargs)
+  export KUBEVIRT_VERSION=$(getLatestPatchVersion v1.3)
 fi
 
 ./cluster/kubectl.sh apply -f https://github.com/kubevirt/kubevirt/releases/download/${KUBEVIRT_VERSION}/kubevirt-operator.yaml


RamLavi:
@oshoval perhaps I should used the go toolchain version first?

oshoval:
```
tar -xf /tmp/cnao/cluster-network-addons-operator/build/_output/bin//go1.22.linux-amd64.tar.gz -C /tmp/cnao/cluster-network-addons-operator/build/_output/bin/
```
it need z stream in the download link,
lets try first to just append `.9` please (aligning both)
https://github.com/kubevirt/cluster-network-addons-operator/blob/release-0.95/go.mod#L3

might need to be done in the component itself / both

depends where the new script should look at
(the right thing might be to look on cnao itself, not on the component)

btw it might be healthier that the component will present the needed hack/go-version.sh
if logic reminds looking on the component (which i am not sure is the right thing to do)

oshoval:
/retest

lets clean false positives first

oshoval:
> It should look at CNAO, as this is what deploys the component.

but, aint it compiles the tests of the component later which should be done according the component ?
anyhow whatever we are doing on main is fine

> So I should fix the original PR on main to first look at toolchain and only then at the other go version?

i think we should not touch main for this, since it is on a release
i would say not to touch toolchain either, it looks on ^go, which just miss the patch,
since you look on CNAO (i will count on you here, since the code is bit confusing), you just need to change the ^go line to match the toolchain, i.e to add `.9`

moreover since it is a release that merely changes, it is safer

RamLavi:
> 
> 
> > It should look at CNAO, as this is what deploys the component.
> 
> but, aint it compiles the tests of the component later which should be done according the component ?
> anyhow whatever we are doing on main is fine
> 
> > So I should fix the original PR on main to first look at toolchain and only then at the other go version?
> 
> i think we should not touch main for this, since it is on a release
> i would say not to touch toolchain either, it looks on ^go, which just miss the patch,
> since you look on CNAO (i will count on you here, since the code is bit confusing), you just need to change the ^go line to match the toolchain, i.e to add `.9`
> 
> moreover since it is a release that merely changes, it is safer
> 

So iow we should make it hard coded, right?

RamLavi:
[Change](https://github.com/kubevirt/cluster-network-addons-operator/compare/4609a4ce44770f455e137bfe467142c0ec9c5fc6..211e75b57f532d5e46b77c1887235452ec7d9eee): Bump go.mod go version to have a patch version. This does not break the go version restriction, as we are only bumping the patch version.

RamLavi:
> https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_cluster-network-addons-operator/2465/pull-e2e-cnao-multus-dynamic-networks-functests-release-0.95/1987779833307860992
> 
> Here for example it takes 1.21, even that release-0.95 need 1.22 so we need to compile CNAO according CNAO, and tests according component (we can start with compiling both according CNAO, and fix according needs / in follow-up)
> 
> ```
> + tar -xf /tmp/deploy.multus-dynamic-networks.qq9O/github.com/k8snetworkplumbingwg/multus-dynamic-networks-controller/build/_output/go1.18/bin//go1.21.linux-amd64.tar.gz -C /tmp/deploy.multus-dynamic-networks.qq9O/github.com/k8snetworkplumbingwg/multus-dynamic-networks-controller/build/_output/go1.18/bin/
> ```

OK I double checked it cause it is confusing to me as well :).
the way it is right now, when we run a tier1 test:
- first CNAO is compiled (`cluster-sync` → `cluster-operator-push` → `docker-build-operator`), we are using the vendored go. This vendored go is [created](https://github.com/kubevirt/cluster-network-addons-operator/blob/7eb8dff6d1172ae4ac27cd42f7ce2087db29558d/build/operator/Dockerfile#L6) from the latest go. (we might want to change that)
- second we download the component repo, and copy the `go-install.sh` to the component repo, then use it to create another vendored go that fits that component's go.mod, and run the test. ---> go is complaint to component.

Now that we know that, I think we (all the maintainers) should agree on how we want it to be, and then change. Are you OK that we set a meeting about it?

RamLavi:
@oshoval PTAL

oshoval:
lets retest, hopefully one timer

https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_cluster-network-addons-operator/2465/pull-e2e-cnao-kube-secondary-dns-functests-release-0.95/1988215027068309504

/retest

oshoval:
https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_cluster-network-addons-operator/2481/pull-e2e-cnao-kube-secondary-dns-functests-release-0.95/1988264574465347584

https://github.com/kubevirt/cluster-network-addons-operator/pull/2465#issuecomment-3517171500

it works,

i will create a PR with just that change so we can merge it -

EDIT done
please check https://github.com/kubevirt/cluster-network-addons-operator/pull/2481

hack/deploy-kubevirt.sh
```diff
@@ -1,9 +1,14 @@
 #!/bin/bash
 set -exo pipefail
 
+function getLatestPatchVersion {
+  local major_minors=$1
+  curl -s https://api.github.com/repos/kubevirt/kubevirt/releases?per_page=100 | grep .tag_name | grep ${major_minors} | sort -V | tail -1 | awk -F':' '{print $2}' | sed 's/,//' | xargs
```
oshoval: Note

after sort -V
```
    "tag_name": "v1.3.0",
    "tag_name": "v1.3.0-alpha.0",
    "tag_name": "v1.3.0-beta.0",
    "tag_name": "v1.3.0-rc.0",
    "tag_name": "v1.3.0-rc.1",
    "tag_name": "v1.3.0-rc.2",
    "tag_name": "v1.3.1",
```

theoretically it can take a RC, not a pure stable, it is fine on old releases, since usually they dont create RC after stable, on main it might select non "stable", but anyhow we need it just for releases

RamLavi: Hey @oshoval I think we can proceed knowing this, right? can you re-approve it?

PR #2486, opened 2025-11-12
[release-0.91] Pin go and kubevirt versions to fit stable branch

RamLavi:
This is an manual cherry-pick of #2482

/assign RamLavi

```release-note
NONE
```

### DIFF
--- a/hack/deploy-kubevirt.sh
+++ b/hack/deploy-kubevirt.sh
@@ -1,9 +1,14 @@
 #!/bin/bash
 set -exo pipefail
 
+function getLatestPatchVersion {
+  local major_minors=$1
+  curl -s https://api.github.com/repos/kubevirt/kubevirt/releases?per_page=100 | grep .tag_name | grep ${major_minors} | sort -V | tail -1 | awk -F':' '{print $2}' | sed 's/,//' | xargs
+}
+
 if [ -z "${KUBEVIRT_VERSION}" ];then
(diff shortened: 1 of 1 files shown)


oshoval:
i think that 1.20 didnt need .z
(1.19 didnt iirc, so maybe 1.20 fails on the same)

RamLavi:
@oshoval PTAL

hack/install-go.sh
```diff
@@ -1,8 +1,28 @@
...
+fi
+
+if [[ "$go_mod_version" =~ ^[0-9]+\.[0-9]+$ ]]; then
+    minor_version=$(echo "$go_mod_version" | cut -d. -f2)
+    # Starting from Go 1.21, the initial release is 1.N.0
+    # For Go 1.20 and earlier, the initial release is 1.N
+    # For more information, see https://go.dev/doc/toolchain#name
+    if [ "$minor_version" -ge 21 ]; then
+        go_mod_version="${go_mod_version}.0"
+    fi
```
oshoval: on this release branch lets do the minimal required, without all the logic, since it just need 1.N

unless you plan something generic that you can put on main and backport, but if we can refrain from it, and keep it simple it would be better please

it seems the logic adds and then remove the `.z`

oshoval: yep

oshoval:
much better
thanks
/lgtm

PR #2497, opened 2025-11-20
bridge marker: Fix readOnlyRootFilesystem

oshoval:
Fixes the following error (the object was still created despite it).

`{"level":"info","ts":"2025-11-19T14:37:42Z","msg":"unknown field \"spec.template.spec.securityContext.readOnlyRootFilesystem\""}`

It should be under container level.

Field purpose is to block writing to all `/`, beside the mounted folders.

Interesting that on KCLI (based on k8s 1.32), it doesn't print the error, and the fields is created under the container.
It is possible that SCC admission controller mutates it.

**Release note**:

```release-note
None
```

### DIFF
--- a/data/linux-bridge/003-bridge-marker.yaml
+++ b/data/linux-bridge/003-bridge-marker.yaml
@@ -50,6 +50,8 @@ spec:
           volumeMounts:
             - name: tmp
               mountPath: /tmp
+          securityContext:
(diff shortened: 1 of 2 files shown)


oshoval:
True
more interesting info on PR desc
even with the error, it was self healing on KCLI (didn't check if it was self healing U/S)

PR #2500, opened 2025-11-23
makefile: Fix verify-codegen.sh narrowing get-k8s scan paths

RamLavi:
Currently, controller-gen scans all directories using paths="./..." which includes build/_output containing the downloaded Go toolchain. This causes vendoring conflicts as the Go source has its own go.mod that conflicts with the project's vendor directory.

This commit fixes this by changing controller-gen to scan only paths="./pkg/...;./cmd/...;./test/...;./tools/..." where the actual API types
are located, excluding the build output directory. It also contains the output of 'make gen-k8s' that fixes the API types.

**Release note**:

```release-note
NONE
```


Makefile
```diff
@@ -191,7 +191,7 @@ gen-manifests: manifest-templator
 		./hack/generate-manifests.sh
 
 gen-k8s: $(CONTROLLER_GEN) $(apis_sources)
-	$(CONTROLLER_GEN) object:headerFile="hack/boilerplate.go.txt" paths="./..."
+	$(CONTROLLER_GEN) object:headerFile="hack/boilerplate.go.txt" paths="./pkg/...;./cmd/...;./test/...;./tools/..."
```
oshoval: can we somehow exclude the problematic ones instead ?
if not, those are all we need ? (it means that in case we will need more in the future we should remember to add them, but i doubt it will happen)

PR #2690, opened 2026-04-26
Fix goimports-check to validate test/ and tools/ directories

### LINKED ISSUE #2680: CI: make goimports-check does not validate test/ and tools/
## Summary
The `goimports` formatting target and the `goimports-check` validation target do not cover the same directories.

Today, `make goimports` formats `./pkg`, `./cmd`, `./test`, and `./tools`, but `make goimports-check` only checks `./pkg` and `./cmd`. This means `make check` can pass even when files under `test/` or `tools/` would be modified by `make goimports`.

## Why this matters
`make check` should fail when import formatting is out of sync anywhere that `make goimports` would modify. Right now there is a gap in CI coverage.

## Current behavior
- `make goimports` runs on:
  - `./pkg`
  - `./cmd`
  - `./test`
  - `./tools`
- `make goimports-check` only checks:
  - `./pkg`
  - `./cmd`

## Expected behavior
`make goimports-check` should validate the same directories that `make goimports` formats.

## Requested fix
Update the validation target so it checks `./test` and `./tools` as well.

## Acceptance criteria
- `make goimports-check` checks the same directory set as `make goimports`
- `make check` therefore catches goimports drift in `test/` and `tools/`
- Keep the fix minimal
- No behavior changes outside this validation gap

## Suggested verification
Please include in the PR description:
- what changed
- the exact command(s) used to verify the fix

Expected local verification:
```bash
make goimports-check
make check

### PR DESCRIPTION
**Pull request by RamLavi**
## Summary

The `goimports-check` Makefile target only validated `./pkg` and `./cmd`, while the `goimports` formatting target also covered `./test/` and `./tools/`. This allowed import formatting drift in those directories to pass `make check` undetected.

## Changes

- Added `./test/` and `./tools/` to the `goimports-check` command in the Makefile (one-line change)

## Root cause

The directory lists in `goimports` (line 90) and `goimports-check` (line 109) were out of sync. `goimports` used `-w ./pkg ./cmd ./test/ ./tools/` but `goimports-check` only used `-d ./pkg ./cmd`.

## Validation

- `make -n goimports-check` confirms the updated command now includes all four directories
- `make -n goimports` confirms both targets now operate on the same directory set
- The only difference between the two commands is the flag: `-w` (write) vs `-d` (diff), which is correct

## Why this PR

Prior PRs #2681, #2684, and #2689 contained the same correct fix but were blocked by missing DCO signoff or release-note formatting. This PR includes both.

Fixes: #2680

```release-note
Fix goimports-check Makefile target to validate test/ and tools/ directories
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -106,7 +106,7 @@ vet: $(GO) $(cmd_sources) $(pkg_sources)
 	touch $@
 
 goimports-check: $(GO) $(cmd_sources) $(pkg_sources)
-	$(GO) tool goimports -d ./pkg ./cmd
+	$(GO) tool goimports -d ./pkg ./cmd ./test/ ./tools/
 	touch $@
 
 test/unit: $(GO)


RamLavi:
@qinqon I think we need to tell your oompa to flag issues it already opened PRs on, so that it won't duplicate it .
closing

PR #2713, opened 2026-05-03
hack: Prefer go toolchain version in go-version.sh

RamLavi:
When a `toolchain` directive is present in go.mod, use its version instead of the `go` directive. This ensures that install-go.sh downloads the correct patch-level Go binary and that Makefile's GO_VERSION reflects the intended toolchain.

This PR makes the repo toolchain aware.

**Release note**:

```release-note
NONE
```


RamLavi:
There is no concesus so closing

PR #2833, opened 2026-06-23
Allow to configure bridge-marker health port

### LINKED ISSUE #2832: Allow configuring bridge-marker health probe port via NetworkAddonsConfig
_Is your feature request related to a problem? Please describe:_
In certain environments, the default health probe port (8081) used by the bridge-marker container conflicts with other workloads or is blocked by network policies. There is currently no way to change this port through the NetworkAddonsConfig API — it is hardcoded in the DaemonSet manifest, requiring manual patching after deployment.

_Describe the solution you'd like:_
Add an optional bridgeMarkerHealthPort field to the linuxBridge section of the NetworkAddonsConfig spec. When set, this value is used as the health probe port for the bridge-marker container (both the -health-probe-port CLI argument and the containerPort declaration). When omitted, the existing default of 8081 is preserved, so the change is fully backwards-compatible.

Example:

```
spec:
  linuxBridge:
    bridgeMarkerHealthPort: 9090
```
_Describe alternatives you've considered:_
- Patching the DaemonSet directly after deployment — this works but is fragile, as the operator will reconcile the manifest back to the hardcoded value on the next sync.
 - Exposing the port as an environment variable on the operator deployment — this would affect all clusters managed by the operator rather than allowing per-cluster configuration.

_Additional context:_
The bridge-marker container accepts a -health-probe-port flag; (commit merged 9fe89af95800813048b6d4c3babd1fedd442dec8)  this feature simply threads that flag through the operator's API so it can be configured without modifying operator internals. Input is validated to be a valid integer in the range [1, 65535].

AKamyshnikova:
Unittests added with Claude

Fixes: #2832 

Adds a `bridgeMarkerHealthPort` field to the `linuxBridge` section of `NetworkAddonsConfig`, allowing users to override the default health probe port (`8081`) of the `bridge-marker` container. This is needed in environments where the default port conflicts with other workloads or is restricted by network policies. The port value is validated to be a valid integer in range [1, 65535]. When not set, the existing default is preserved for full backwards compatibility.

The `bridge-marker` container already supports the `-health-probe-port` flag  (commit merged 9fe89af95800813048b6d4c3babd1fedd442dec8)— this change only exposes it through the operator API. Both the container port declaration and the CLI argument are templated from the same value to keep them in sync.

**Release note**:

```release-note
Added optional `bridgeMarkerHealthPort` field to the `linuxBridge` spec, allowing users to configure a custom health probe port for the bridge-marker container. Defaults to `8081` if not set.
```

### DIFF
--- a/data/linux-bridge/003-bridge-marker.yaml
+++ b/data/linux-bridge/003-bridge-marker.yaml
@@ -37,6 +37,8 @@ spec:
           args:
             - -node-name
             - $(NODE_NAME)
+            - -health-probe-port
+            - "{{ .HealthPort }}"
           resources:
             requests:
               cpu: "10m"
@@ -48,7 +50,7 @@ spec:
                   fieldPath: spec.nodeName
           ports:
             - name: healthport
-              containerPort: 8081
+              containerPort: {{ .HealthPort }}
               protocol: TCP
           terminationMessagePolicy: FallbackToLogsOnError
           readinessProbe:
--- a/pkg/apis/networkaddonsoperator/shared/networkaddonsconfig_types.go
+++ b/pkg/apis/networkaddonsoperator/shared/networkaddonsconfig_types.go
@@ -61,7 +61,9 @@ type MultusDynamicNetworks struct {
 }
 
 // LinuxBridge plugin allows users to create a bridge and add the host and the container to it
-type LinuxBridge struct{}
+type LinuxBridge struct {
+	BridgeMarkerHealthPort *int32 `json:"bridgeMarkerHealthPort,omitempty"`
+}
 
 // Ovs plugin allows users to define Kubernetes networks on top of Open vSwitch bridges available on nodes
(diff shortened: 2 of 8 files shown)


AKamyshnikova:
@phoracek Hi! Could you take a look at this PR? This is a logical continuation for changes in bridge-marker.

phoracek:
@RamLavi @qinqon @oshoval would one of you please take this?

oshoval:
Thanks for the PR, looks good overall
Please create a tag on https://github.com/kubevirt/bridge-marker
it will auto bump, and then we can continue, bumping shouldn't be done manually please

oshoval:
i wonder if there is some port that we can use, which is pretty free, as default

AKamyshnikova:
@oshoval Got it! I will wait for bridge-marker tag.
@phoracek Could you create bridge-marker release already or is there a schedule?

PR #2830, opened 2026-06-23
CI Failure: pull-e2e-cnao-multus-dynamic-networks-functests / The `pull-e2e-cnao-multus-dynamic-networks-functests` CI

RamLavi:
Fixes #2829

**What this PR does / why we need it**:

This PR upgrades the yq tool from version 3.3.4 to version 4 to resolve CI failures caused by the unavailability of the `docker.io/mikefarah/yq:3.3.4` Docker image on Docker Hub.

The `pull-e2e-cnao-multus-dynamic-networks-functests` CI job was failing with a 404 error when attempting to pull the yq:3.3.4 image, which has been removed from Docker Hub after being in use for 5+ years (since December 2020, commit 3bf8f7174).

The fix includes:
- Updating the Docker image reference from `mikefarah/yq:3.3.4` to `mikefarah/yq:4`
- Migrating the command syntax from yq v3 to v4 API (adding dot prefix to paths for jq-style syntax)
- Maintaining backward compatibility with all existing yaml-utils functions

**Special notes for your reviewer**:

The yq v3 to v4 migration is straightforward - we only need to prepend a dot (`.`) to paths to convert them to yq v4's jq-style path syntax:
- `yq r file path` becomes `yq eval '.path' file`
- `yq w -i file path value` becomes `yq eval '.path = "value"' -i file`
- `yq d -i file path` becomes `yq eval 'del(.path)' -i file`

All existing usages of yaml-utils functions (get_param, set_param, delete_param, update_param) remain compatible without changes to calling code.

**Release note**:

```release-note
fix: upgrade yq dependency from v3.3.4 to v4 to resolve Docker Hub image availability issues in CI
```

PR #2845, opened 2026-06-30
CI Failure: pull-cluster-network-addons-operator-unit-test / The `pull-cluster-network-addons-operator-unit-test` job

### LINKED ISSUE #2837: CI Failure: pull-cluster-network-addons-operator-unit-test / The `pull-cluster-network-addons-operator-unit-test` job
Periodic CI job **pull-cluster-network-addons-operator-unit-test** failed in run [2069353076120096769](https://storage.googleapis.com/kubevirt-prow/pr-logs/pull/kubevirt_cluster-network-addons-operator/2830/pull-cluster-network-addons-operator-unit-test/2069353076120096769/build-log.txt).

## Analysis

Based on my investigation, I have identified the root cause of the test failures. Let me provide a structured analysis.

## Summary

The `pull-cluster-network-addons-operator-unit-test` job fails with 5 test failures in `RenderKubevirtIPAMController`. Tests expect TLS configuration flags (`--tls-min-version` and `--tls-cipher-suites`) in the rendered deployment args, but the generated deployment only contains `["--leader-elect", "\"--certificates-dir=\""]`.

## Root Cause

The failure is caused by a mismatch between the kubevirt-ipam-controller bump script (`hack/components/bump-kubevirt-ipam-controller.sh`) and the committed manifest file (`data/kubevirt-ipam-controller/001-kubevirtipamcontroller.yaml`):

1. **CI execution flow**: The CI script (`automation/check-patch.unit-test.sh:16`) runs `make bump-all` before tests, which regenerates all component manifests including kubevirt-ipam-controller.

2. **Bump script sed order**: The bump script (lines 35-37) uses `sed` commands to insert TLS configuration lines after `--certificates-dir`:
   ```bash
   sed -i '/--certificates-dir=/a{{ if index . "TLSSecurityProfileCiphers" }}...'  # Line 35
   sed -i '/--certificates-dir=/a\\            - "--tls-min-version=..."'          # Line 36  
   sed -i '/--certificates-dir=/a{{- if ne .DefaultNetNADNs "" }}...'              # Line 37
   ```

3. **Order mismatch**: These `sed` commands execute sequentially, each inserting AFTER the same `--certificates-dir` line. This produces:
   ```
   --certificates-dir
   TLSSecurityProfileCiphers block (from line 35, inserted last in sequence)
   --tls-min-version (from line 36)
   DefaultNetNADNs block (from line 37, inserted first in sequence)
   ```

4. **Committed manifest has different order**: The manually committed manifest (lines 184-193) has:
   ```
   --certificates-dir
   DefaultNetNADNs block
   --tls-min-version
   TLSSecurityProfileCiphers block
   ```

5. **Test failure mechanism**: When `make bump-all` regenerates the manifest with the incorrect sed-driven ordering, the resulting YAML structure may become malformed or the template rendering fails, causing the deployment args to only contain the first two elements.

**Key evidence**:
- pkg/network/kubevirt_ipam_controller_test.go:42 - Tests expect TLS flags to be present
- CI log shows "Uncommitted changes found after bump check" - indicating `make bump-all` modified files
- Commit a8b31057d added TLS settings in March 2026, updating both bump script and manifest, but introduced this ordering inconsistency

## Classification

**CODE_BUG**

## Suggested Fix

Update `hack/components/bump-kubevirt-ipam-controller.sh` lines 35-37 to generate the manifest in the correct order matching the committed version. The sed commands should be reordered or use different anchor patterns to produce:

1. `--certificates-dir`
2. DefaultNetNADNs conditional block  
3. `--tls-min-version`
4. TLSSecurityProfileCiphers conditional block

Alternatively, regenerate the manifest using the current bump script and commit the result, then update tests if needed. The fix requires ensuring the bump script and committed manifest stay in sync.

---
*This issue was automatically created by oompa based on CI failure analysis.*

RamLavi:
Fixes #2837

This PR fixes the sed command in `hack/components/bump-kubevirt-ipam-controller.sh` that was incorrectly using a single backslash instead of a double backslash when inserting the `--tls-min-version` flag.

In PR #2830 (yq v3 to v4 migration), the sed commands were updated to match the new YAML format without quotes. However, the sed command on line 36 was changed from `a\\` to `a\`, which could cause issues with how sed interprets the command. This PR restores the double backslash to ensure proper escaping and consistent behavior.

The issue was introduced during the yq v3 to v4 migration in PR #2830. The sed command needs `a\\` (double backslash) to properly append a line with leading whitespace in all cases. While a single backslash may work in some contexts, the double backslash is the standard and more portable approach.

This fix ensures that the bump script correctly generates the kubevirt-ipam-controller manifest with TLS configuration flags in the proper order.

**Release note**:

```release-note
fix: correct sed escaping in kubevirt-ipam-controller bump script to ensure TLS flags are properly generated
```

---
*🤖 Generated by [oompa](https://github.com/qinqon/oompa). Use `/oompa` to talk with me.*

### DIFF
--- a/data/kubemacpool/kubemacpool.yaml
+++ b/data/kubemacpool/kubemacpool.yaml
@@ -126,7 +126,7 @@ subjects:
 apiVersion: v1
 data:
   RANGE_END: '{{ .RangeEnd }}'
-  RANGE_START: '{{ .RangeStart }}'
+  RANGE_START: "{{ .RangeStart }}"
 kind: ConfigMap
 metadata:
   labels:
--- a/data/kubevirt-ipam-controller/001-kubevirtipamcontroller.yaml
+++ b/data/kubevirt-ipam-controller/001-kubevirtipamcontroller.yaml
@@ -183,13 +183,13 @@ spec:
       containers:
         - args:
             - --leader-elect
-            - "--certificates-dir={{ .CertDir }}"
+            - --certificates-dir={{ .CertDir }}
 {{- if ne .DefaultNetNADNs "" }}
-            - "--default-network-nad-namespace={{ .DefaultNetNADNs }}"
+            - --default-network-nad-namespace={{ .DefaultNetNADNs }}
 {{ end }}
-            - "--tls-min-version={{ .TLSMinVersion }}"
+            - --tls-min-version={{ .TLSMinVersion }}
 {{ if index . "TLSSecurityProfileCiphers" }}
-            - "--tls-cipher-suites={{ .TLSSecurityProfileCiphers }}"
+            - --tls-cipher-suites={{ .TLSSecurityProfileCiphers }}
 {{ end }}
           command:
             - /manager
--- a/data/multus/001-multus.yaml
+++ b/data/multus/001-multus.yaml
@@ -156,8 +156,8 @@ spec:
             - name: cni
               mountPath: /host/etc/cni/net.d
             # multus-daemon expects that cnibin path must be identical between pod and container host.
-            # e.g. if the cni bin is in /opt/cni/bin on the container host side, then it should be mount to /opt/cni/bin in multus-daemon,
-            # not to any other directory, like /opt/bin or /usr/bin.
+            # e.g. if the cni bin is in '/opt/cni/bin' on the container host side, then it should be mount to '/opt/cni/bin' in multus-daemon,
+            # not to any other directory, like '/opt/bin' or '/usr/bin'.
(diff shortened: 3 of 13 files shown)


RamLavi:
Closing — duplicate fix attempt. PR #2830 is already addressing the same root cause (yq v4 migration). Oompa should not create new PRs when an existing open PR covers the same issue. Filed [qinqon/oompa#264](https://github.com/qinqon/oompa/issues/264).

PR #2909, opened 2026-08-03
fix(ci): pin K8S_VERSION to v1.36.1 for ipam-ext e2e tests

RamLavi:
## What this PR does / why we need it

Fixes the `ipam-ext` E2E job failing on all CNAO PRs with:
```
ERROR: failed to create cluster: failed to pull image "kindest/node:v1.36.2"
Command Output: Error response from daemon: manifest for kindest/node:v1.36.2 not found
```

The kubevirt-ipam-controller repo defaults to `K8S_VERSION=v1.36.2`, whose custom-built kind node image is not published/accessible in CI. This PR overrides it to `v1.36.1`, which is available on `docker.io/kindest/node`.

## Special notes for your reviewer

Stopgap fix to unblock CI. Once ovn-kubernetes publishes their custom `v1.36.2` node images, we can revert this override.

## Release note

```release-note
NONE
```

PR #2918, opened 2026-08-11
Remove KubeSecondaryDNS

oshoval:
Since KSD isn't used anymore since 4.22, remove it from CNAO.

API was kept so we won't break HCO and won't change API.

**Release note**:

```release-note
None
```


oshoval:
https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_cluster-network-addons-operator/2918/pull-e2e-cluster-network-addons-operator-workflow-k8s/2087069117772206080

might be TLS failure unrelated to this PR, i think i saw it on other PRs

oshoval:
lets see if temp or not

https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_cluster-network-addons-operator/2918/pull-e2e-cluster-network-addons-operator-workflow-k8s-s390x/2087076828920418304

/test pull-e2e-cluster-network-addons-operator-workflow-k8s-s390x

we might need the memory extension of
https://github.com/kubevirt/cluster-network-addons-operator/pull/2909/files#diff-426e7a410d92c9ee2ea4e365697ede2f5f829539cdb3b10d623d170a326d6076R26
as we see 
dynamic-networks-controller-ds stayed Pending and the scheduler message was Insufficient memory

oshoval:
failed again
https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_cluster-network-addons-operator/2918/pull-e2e-cluster-network-addons-operator-workflow-k8s-s390x/2087103186606231552

first thing is to try add memory

cc @RamLavi 
can we drop this lane otoh ? (with whom should we speak about it ? Ashok ?)

oshoval:
added memory extension, lets see if it works, if needed will drop it in favor of Ram's PR

RamLavi:
@oshoval just making sure we're aligned - let's make sure to merge this only after we branch the stable branch

oshoval:
> @oshoval just making sure we're aligned - let's make sure to merge this only after we branch the stable branch

np

RamLavi:
> > @oshoval just making sure we're aligned - let's make sure to merge this only after we branch the stable branch
> 
> np

FYI we're now branched

oshoval:
Thanks
ready for review please (rebased)

we might need to drop it also from HCO, 
cc @nunnatsa

oshoval:
looks like unrelated flake 
https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_cluster-network-addons-operator/2918/pull-e2e-cluster-network-addons-operator-kubemacpool-functests/2091815304261799936

oshoval:
Hi @RamLavi can you please review ?

oshoval:
https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_cluster-network-addons-operator/2918/pull-e2e-cnao-multus-dynamic-networks-functests/2096528775884836864

flake here, unrelated to the PR

pkg/components/components.go
```diff
@@ -1417,20 +1397,6 @@ func GetCrd() *extv1.CustomResourceDefinition {
 								},
 							},
 						},
-						"kubeSecondaryDNS": extv1.JSONSchemaProps{
```
RamLavi, line 1420: This contradicts the PR note that the API was kept so HCO would not break.

The Go field is still on `NetworkAddonsConfigSpec`, but `GetCrd()` no longer defines `spec.kubeSecondaryDNS`. For CRDs with a structural OpenAPI schema, unspecified fields are pruned on persist (`preserveUnknownFields: false` is already set on our CRD), and kubectl's client-side OpenAPI check can reject objects that still send the field:

https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#field-pruning

HCO still writes this field when the feature gate is enabled (although it is false by default)

Please put `kubeSecondaryDNS` back on the CRD schema with a deprecated description (ignored / no longer deployed). Then drop the CRD property only after HCO stops setting it.

pkg/apis/networkaddonsoperator/shared/networkaddonsconfig_types.go
```diff
@@ -17,12 +17,14 @@ type NetworkAddonsConfigSpec struct {
...
 	ImagePullPolicy        corev1.PullPolicy         `json:"imagePullPolicy,omitempty"`
 	NMState                *NMState                  `json:"nmstate,omitempty"`
-	KubeSecondaryDNS       *KubeSecondaryDNS         `json:"kubeSecondaryDNS,omitempty"`
 	MacvtapCni             *MacvtapCni               `json:"macvtap,omitempty"`
 	KubevirtIpamController *KubevirtIpamController   `json:"kubevirtIpamController,omitempty"`
 	SelfSignConfiguration  *SelfSignConfiguration    `json:"selfSignConfiguration,omitempty"`
 	PlacementConfiguration *PlacementConfiguration   `json:"placementConfiguration,omitempty"`
 	TLSSecurityProfile     *ocpv1.TLSSecurityProfile `json:"tlsSecurityProfile,omitempty"`
+
+	// Deprecated
```
RamLavi: IIUC `// Deprecated` on the field and on the type does not tell callers that CNAO no longer deploys anything for this spec.

Please use the Go deprecation form so `go doc` / IDEs surface it:

https://go.dev/wiki/Deprecated

Something like:

```go
// Deprecated: KubeSecondaryDNS is no longer deployed by CNAO. The field is
// retained for API compatibility and is ignored.
KubeSecondaryDNS *KubeSecondaryDNS `json:"kubeSecondaryDNS,omitempty"`
```
Same for the KubeSecondaryDNS type, and for the CRD Description if the schema field stays.

pkg/network/network.go
```diff
@@ -82,6 +82,7 @@ func SpecialCleanUp(conf *cnao.NetworkAddonsConfigSpec, client k8sclient.Client,
 
 	errs = append(errs, cleanUpMultus(conf, ctx, client)...)
 	errs = append(errs, cleanUpKubevirtIpamController(conf, ctx, client, clusterInfo.OpenShift4)...)
+	errs = append(errs, cleanUpKubeSecondaryDNS(ctx, client)...)
```
RamLavi, line 85: `cleanUpKubeSecondaryDNS` is the upgrade path, but nothing in this PR exercises it.

Please add a targeted check that after upgrade (or after reconcile with `spec.kubeSecondaryDNS` set), these objects are gone: Deployment `secondary-dns`, ConfigMap `secondary-dns`, SA/ClusterRole/ClusterRoleBinding `secondary`, NetworkPolicy `allow-ingress-to-secondary-dns`.

RamLavi (commented):
Thanks for this change!
I added comments, but I think we should hold until 
- HCO removes the HCO `deployKubeSecondaryDNS` FG to remove it. IIUC it is still possible to ship it both U/S and D/S
- kubevirtci removes it on the vendorred [CNAO](https://github.com/kubevirt/kubevirtci/blob/main/cluster-provision/gocli/opts/cnao/manifests/network-addons-config-example.cr.yaml#L9) 
- project-infra is already fixed which is good.
- did I miss another place?

/hold

pkg/network/network.go
oshoval: i think we don't need this specific corner case, KSD is optional for long time,
no one cared for it, 
we don't need to maintain another case for such simple clean-up

same as we dont have special case for the old ipam entities and such
lets keep it simple imho

btw even operand delete flow isn't tested for any operand

oshoval:
thanks, i dont think we need hold, as long as we don't break HCO, 
and this is done by keeping the API, and the flags stub, even if HCO set it, it will do nothing, which is fine

once i address comments, imho we can merge it without creating other dependencies,
since it will be safe, and then we can drop stuff orthogonal from HCO etc

pkg/network/network.go
RamLavi: fair enough

RamLavi:
> thanks, i dont think we need hold, as long as we don't break HCO, and this is done by keeping the API, and the flags stub, even if HCO set it, it will do nothing, which is fine
> 
> once i address comments, imho we can merge it without creating other dependencies, since it will be safe, and then we can drop stuff orthogonal from HCO etc

I see your point. My concern is mostly about the release flow here.

If we merge this first, we won't really know whether we've missed an HCO dependency until we cut a CNAO release, bump it in HCO, and run the HCO tests. If something still depends on the FG there, we'd need to fix it and potentially cut another CNAO release, which adds some avoidable iteration.

I'd prefer to remove the KSD knob from HCO first. Then we know HCO no longer depends on this API behavior, and the CNAO removal becomes genuinely decoupled and safer to merge independently.

So I'd still prefer keeping the hold until the HCO side is cleaned up.

RamLavi:
> > thanks, i dont think we need hold, as long as we don't break HCO, and this is done by keeping the API, and the flags stub, even if HCO set it, it will do nothing, which is fine
> > once i address comments, imho we can merge it without creating other dependencies, since it will be safe, and then we can drop stuff orthogonal from HCO etc
> 
> I see your point. My concern is mostly about the release flow here.
> 
> If we merge this first, we won't really know whether we've missed an HCO dependency until we cut a CNAO release, bump it in HCO, and run the HCO tests. If something still depends on the FG there, we'd need to fix it and potentially cut another CNAO release, which adds some avoidable iteration.
> 
> I'd prefer to remove the KSD knob from HCO first. Then we know HCO no longer depends on this API behavior, and the CNAO removal becomes genuinely decoupled and safer to merge independently.
> 
> So I'd still prefer keeping the hold until the HCO side is cleaned up.

I talked to @nunnatsa , he says that removing the KSD FG is relatively easy, and also provided with a set of 2DOs for anyone that would want to do that:
```
To drop the deployKubeSecondaryDNS FG and deprecate the KubeSecondaryDNSNameServerIP fields

1. remove its json item from pkg/featuregatedetails/feature-gates.json (27-31)
2. in api/v1beta1/hyperconverged_types.go:
  - remove the default from the FeatureGates field (line 70)
  - remove the default from the spec field (line 630)
  - for the DeployKubeSecondaryDNS field (lines 403 - 407): change the comment to // Deprecated: This feature gate   is ignored  and remove the two default lines
  - deprecate the KubeSecondaryDNSNameServerIP field in the HyperConvergedSpec type. Something like // Deprecated: this field is not in use and ignored

3. in api/v1/hyperconverged_types.go:
  - deprecate the KubeSecondaryDNSNameServerIP filed in the NetworkingConfig type. Something like // Deprecated: this field is not in use and ignored

4. Conversion:
  - in api/v1beta1/conversion.go drop lines 416, 432 and 439. 
  - Fix the fuzz test: api/v1beta1/conversion_fuzz_test.go: drop lines 242-244, 462-467
  - fix the tests in api/v1beta1/conversion_test.go

5. from controllers/handlers/networkAddons.go:
  - drop lines 154 - 165
  - drop the getKSDNameServerIP function

6. Remove all the tests that checks the FG or using it
7. run make build-manifests to re-generate and to update the CRD
8. run make make generate-doc
9. Update docs/cluster-configuration.md
  - drop the FG section (lines 103 - 107)
  - drop the KubeSecondaryDNS Name Server IP section (including the example), lines 1293-1324
10. same in docs/cluster-configuration-v1beta1-api.md
```

oshoval:
Thanks
already started https://github.com/kubevirt/hyperconverged-cluster-operator/pull/4529
will verify it against your list

pkg/components/components.go
oshoval: done

not sure we need to drop it later even after HCO drops it, since API doesnt need to be dropped but deprecated, as done on other repos

pkg/apis/networkaddonsoperator/shared/networkaddonsconfig_types.go
oshoval: done

pkg/network/network.go
oshoval: btw, it does run every reconcile, regardless of upgrade path
and it doesn't check this (maybe i should add to save some cycles)
```
	if conf.KubeSecondaryDNS == nil {
		return []error{}
	}
```

RamLavi: but we do want to remove them, even if they configure it, no?

oshoval: `if conf.KubeSecondaryDNS == nil ` means if they are dont configured, just short circuit that we can add if we want to save cycles

if it is configured it will be cleaned on reconcile once this PR lands, yes

oshoval:
HCO was merged, thx Nahshon, Ram
created https://github.com/kubevirt/kubevirtci/pull/1843

RamLavi:
/hold cancel

I will try to re-review this during this week.

oshoval:
kubevirtci PR was merged (not bumped yet in kubevirt but it shouldn't block this PR)

RamLavi (approved):
/lgtm

PR #2923, opened 2026-08-17
Add support for configuring TLS groups

ormergi:
This PR extend the NetworkAddonsConfig CRD to enable configuring TLS groups.

This change enable improving TLS security posture of CNAO pods and operands. 

- This PR wires CNAO operator pods metric server to the TLSProfice TLSGroups.
- Follow up PRs will wire operands to the TLS profile TLS groups, such as Kubemacpool and kubevirt-ipam-controller.
- This PR extend the e2e test to cover explicit TLS group setting via Custom profile take affect.
- https://redhat.atlassian.net/browse/CNV-84845

**Release note**:

```release-note
Extend NetworkAddonsConfig to support configuring TLS groups via tlsSecurityProfile
```


pkg/network/tlsSecurityProfile.go
```diff
@@ -44,6 +44,7 @@ var ocpTLSProfileOpenSSLCipherNames = map[string]uint16{
 type tlsConfig struct {
 	MinVersion   ocpv1.TLSProtocolVersion
 	CipherSuites []string
+	Groups       []ocpv1.TLSGroup
 }
 
 func SelectTLSSettings(profile *ocpv1.TLSSecurityProfile) tlsConfig {
```
nirdothan, line 50: Commit message: "Extend 'SelectCipherSuitesAndMinTLSVersion'..." now  "SelectTLSSettings".

pkg/network/tlsSecurityProfile_test.go
```diff
@@ -233,4 +251,35 @@ var _ = Describe("Testing TLS Security Profile", func() {
...
+				tls.X25519,
+				tls.CurveP256,
+				tls.CurveP384,
+				tls.CurveP521,
+			}))
+		})
+		It("should skip groups with no known crypto/tls.CurveID", func() {
+			ids := CurveIDs([]ocpv1.TLSGroup{
+				ocpv1.TLSGroupX25519,
+				ocpv1.TLSGroupSecP256r1MLKEM768,  // available on Go 1.26
```
nirdothan: Didn't you just add it in line 137? should it still be skipped?

pkg/network/tlsSecurityProfile.go
ormergi: Fixed

pkg/network/tlsSecurityProfile_test.go
ormergi: They should be remvoed for this test, yes. Good catch, thanks!
Fixed

ormergi:
[push1](https://github.com/kubevirt/cluster-network-addons-operator/compare/be6b02be98e76449bd40f8a533e4588c85f4a7e9..b1ee3023642730f3c0e5d3a58c74cf3ae0da4655) - rabese
[push2](https://github.com/kubevirt/cluster-network-addons-operator/compare/b1ee3023642730f3c0e5d3a58c74cf3ae0da4655..2e3ab2fe54e5e3458dab8726d9b839c576cd6bd1) - address https://github.com/kubevirt/cluster-network-addons-operator/pull/2923#pullrequestreview-5002097548

nirdothan (approved):
Thank you!

ormergi:
[Rebase](https://github.com/kubevirt/cluster-network-addons-operator/compare/2e3ab2fe54e5e3458dab8726d9b839c576cd6bd1..849b54037cc12488ff3762d4621a7fb5e8c5e949)

ormergi:
[push1](https://github.com/kubevirt/cluster-network-addons-operator/compare/849b54037cc12488ff3762d4621a7fb5e8c5e949..d52dccbcfdbefe67581d94ad7f092e29ba0c69e3): update NetworkAddonsConfig CRD following OCP API update
[push2](https://github.com/kubevirt/cluster-network-addons-operator/compare/d52dccbcfdbefe67581d94ad7f092e29ba0c69e3..24e5946966da7d44f356651db52d2d8208ad9e96): refactor and add func test for covering TLS groups

pkg/components/components.go
```diff
@@ -1195,6 +1197,26 @@ func GetCrd() *extv1.CustomResourceDefinition {
 			},
 			Type: "array",
 		},
+		"groups": extv1.JSONSchemaProps{
```
RamLavi, line 1198: openshift/api [marks](https://github.com/openshift/api/blob/70f01b82bb53/config/v1/types_tlssecurityprofile.go#L226-L228) groups as +listType=set (unique, min 1, max 7). MinItems/MaxItems and the enum are here, but uniqueness is not.
Kube-apiserver does not enforce OpenAPI uniqueItems. 

Please set XListType to "set" so duplicates are rejected, matching the field description.
Also a small GetCrd() test like the existing ciphers uniqueness test would help keep this from drifting again.

pkg/network/tlsSecurityProfile.go
```diff
@@ -120,3 +125,29 @@ var tlsVersionID = map[ocpv1.TLSProtocolVersion]uint16{
...
+	ocpv1.TLSGroupX25519MLKEM768:     tls.X25519MLKEM768,
+	ocpv1.TLSGroupSecP256r1MLKEM768:  tls.SecP256r1MLKEM768,
+	ocpv1.TLSGroupSecP384r1MLKEM1024: tls.SecP384r1MLKEM1024,
+}
+
+// CurveIDs converts OpenShift TLS group identifiers to crypto/tls.CurveID values
+// suitable for tls.Config.CurvePreferences.
+// Unknown groups are silently skipped, in case all groups are unknown TLS groups is
+// selected by the runtime.
+func CurveIDs(groups []ocpv1.TLSGroup) []tls.CurveID {
```
RamLavi, line 150: openshift/api [says](https://docs.redhat.com/en/documentation/red_hat_openshift_service_mesh/3.3/html/release_notes/ossm-release-notes#ossm-release-3-3-new-features-enhancements_ossm-release-notes) X25519MLKEM768 is not FIPS-approved and must be omitted by FIPS components.

CurveIDs currently forwards the profile list as-is. Go will drop non-FIPS groups, so the predefined profiles still work. A Custom profile that lists only a PQ hybrid becomes an empty CurvePreferences after FIPS filtering and the handshake fails.

How does HCO handle this? Should we document that Custom-only-PQ is expected to fail there?

pkg/network/tlsSecurityProfile.go
```diff
@@ -75,6 +79,7 @@ func SelectTLSSettings(profile *ocpv1.TLSSecurityProfile) tlsConfig {
 	return tlsConfig{
 		MinVersion:   minTlsVersion,
 		CipherSuites: result,
+		Groups:       groups,
```
RamLavi: `groups` is a slice. we should dedup-copy it same way as we did `result`, so we do not alias TLSProfiles or the live spec.

pkg/network/tlsSecurityProfile_test.go
```diff
@@ -53,6 +58,18 @@ var _ = Describe("Testing TLS Security Profile", func() {
 			},
 			expectedCiphers:       ocpv1.TLSProfiles[ocpv1.TLSProfileIntermediateType].Ciphers,
 			expectedMinTLSVersion: ocpv1.TLSProfiles[ocpv1.TLSProfileIntermediateType].MinTLSVersion,
+			expectedGroups:        ocpv1.TLSProfiles[ocpv1.TLSProfileIntermediateType].Groups,
+		}),
+		Entry("when Modern Security Profile is selected", loadSecurityProfileCase{
```
RamLavi, line 71: Worth adding:
- Modern/Intermediate/Old Groups piped through CurveIDs
- Custom with Groups omitted (nil -> runtime defaults)

test/e2e/compliance/tls_compliance_test.go
```diff
@@ -132,6 +133,38 @@ var _ = Describe("TLS", func() {
...
+		// TODO: remove below slice and assert all tested components once they are all wired to tlsSecurityProfile groups
+		tlsGroupSupportedHosts := []string{cnaoHost}
+
+		By("asserting the specified TLS group is set")
+		expectedNegotiatedCurves := map[string]string{"TLS 1.3": "CurveP521"}
+		Eventually(func(g Gomega) {
+			tlsReports, err := getNamespaceTLSReports(cnaoNamespace)
+			g.Expect(err).NotTo(HaveOccurred())
+			actualReports := filterTLSReportsBySpecHost(tlsReports, tlsGroupSupportedHosts...)
+			Expect(actualReports).To(HaveLen(len(tlsGroupSupportedHosts)))
```
RamLavi:
```suggestion
			g.Expect(actualReports).To(HaveLen(len(tlsGroupSupportedHosts)))
```

test/e2e/compliance/tls_compliance_test.go
```diff
@@ -132,6 +133,38 @@ var _ = Describe("TLS", func() {
 			actualReports := filterTLSReportsBySpecHost(tlsReports, cnaoHost, ipamExtHost, kmpMetricsHost, kmpSvcHost)
 			g.Expect(actualReports).To(WithTransform(normalizeConditions, ConsistOf(expectedReports)))
 		}, 5*time.Minute, 1*time.Second).Should(Succeed())
+
+		By("set tlsSecurityProfile with custom type and explicit TLS group")
+		c := operations.ConvertToConfigV1(operations.GetConfig(operations.GetCnaoV1GroupVersionKind()))
+		c.Spec.TLSSecurityProfile = &ocpv1.TLSSecurityProfile{}
+		c.Spec.TLSSecurityProfile.Type = ocpv1.TLSProfileCustomType
+		c.Spec.TLSSecurityProfile.Custom = &ocpv1.CustomTLSProfile{TLSProfileSpec: ocpv1.TLSProfileSpec{
```
RamLavi, line 141: why do you omit Ciphers?

RamLavi (commented):
Thanks! Looks good overall, 
added comments

test/e2e/compliance/tls_compliance_test.go
ormergi: Min version is 1.3, they will be ignored by the runtime anyways

pkg/network/tlsSecurityProfile.go
ormergi: I dont know how HCO handles it, there is reference for this on OCP API though https://github.com/openshift/api/blob/master/config/v1/types_tlssecurityprofile.go#L6

For our concern we dont need to do anything because the project doesnt support FIPS mode, right?

RamLavi: no idea, that's why I'm asking if HCO is taking care of it for us :)

pkg/components/components.go
ormergi: I rather implement this outside this odd CRD implementation, if you dont mind.

The fact the project generate the CRD programmatically is absurd, is should use controller-gen as 99% of Go projects involve k8s controllers, if it used it we wouldn't not have this conversation or any concern about the CRD implemention.

pkg/network/tlsSecurityProfile.go
ormergi: I dont follow, do you mean groups should not contain duplicates?

pkg/network/tlsSecurityProfile_test.go
ormergi: > Modern/Intermediate/Old Groups piped through CurveIDs

I dont follow, could you re-phrase or give example?

> Custom with Groups omitted (nil -> runtime defaults)

Please note this is covered by other tests (who doesnt specify Groups)

pkg/network/tlsSecurityProfile.go
ormergi: Done

test/e2e/compliance/tls_compliance_test.go
ormergi: Done

ormergi:
[Changes](https://github.com/kubevirt/cluster-network-addons-operator/compare/fcb82bde7f1e741d2f35104af5b8b9a167fd2c48..fd5bb6ff4043acd8464026ccefb275858614eb03): resolve https://github.com/kubevirt/cluster-network-addons-operator/pull/2923#discussion_r3850665111 https://github.com/kubevirt/cluster-network-addons-operator/pull/2923#discussion_r3850555708 https://github.com/kubevirt/cluster-network-addons-operator/pull/2923#discussion_r3850572211

pkg/components/components.go
RamLavi: +1, but that's an entirely different task.. 
AFAIU, this CRD implementation is the source of truth (i.e. what contoller-gen end up configuring), so until we refactor CNAO to use the builder again - this is what we should do

pkg/network/tlsSecurityProfile_test.go
RamLavi: >> Modern/Intermediate/Old Groups piped through CurveIDs

>I dont follow, could you re-phrase or give example?

The table already checks that SelectTLSSettings returns TLSProfiles[type].Groups. CurveIDs is tested with a hand-built slice. Nothing checks the combination the metrics server actually uses: CurveIDs(SelectTLSSettings(profile).Groups).

For Modern that would look like:
```
  tlsCfg := SelectTLSSettings(&ocpv1.TLSSecurityProfile{
      Type:   ocpv1.TLSProfileModernType,
      Modern: &ocpv1.ModernTLSProfile{},
  })
  Expect(CurveIDs(tlsCfg.Groups)).To(Equal([]tls.CurveID{
      tls.X25519MLKEM768,
      tls.X25519,
      tls.CurveP256,
      tls.CurveP384,
  }))
```
Same idea for Intermediate/Old.

>>Custom with Groups omitted (nil -> runtime defaults)

>Please note this is covered by other tests (who doesnt specify Groups)

The tests that omit Groups are the predefined profiles. Those always get TLSProfiles[type].Groups, which is non-empty. Custom currently sets Groups: {"mygroup"}. There is no Custom case with Groups left unset, which is the API "no opinion" path (nil Groups -> CurveIDs nil -> Go defaults).

pkg/components/components.go
RamLavi: To be clear, I am not asking to switch this CRD to controller-gen in this PR.

GetCrd() is what we install today. Please still set XListType: "set" here, same as MinItems/MaxItems, so the schema matches openshift/api until a generator rewrite happens.

pkg/network/tlsSecurityProfile.go
RamLavi: I asked on their slack

pkg/components/components.go
ormergi: Done
Added XListType: "set" to the CRD definition, matching what controller-gen produce.

ormergi:
[Changes](https://github.com/kubevirt/cluster-network-addons-operator/compare/d3ee240ed25e554729097e40bba63feea3c32f71..084d3d76ae06da7ffc33f0e75a28cbb659f9e835): resolve https://github.com/kubevirt/cluster-network-addons-operator/pull/2923#discussion_r3846238032

pkg/network/tlsSecurityProfile.go
ormergi: I am not sure about HCO, but from what I understand running on FIPS mode require building binaries with additional flags/env var so that for FIPS there would be dedicated binaries, ref [[1]](https://go.dev/doc/security/fips140).
On such binaries, built for FIPS, according to Go docs [[1]](https://go.dev/doc/security/fips140), non FIPS complaint TLS settings should be considered unkown and  ignored silently
```
The crypto/tls package will ignore and not negotiate any protocol version, cipher suite, signature algorithm, or key exchange mechanism that is not FIPS 140-3 approved. (This is equivalent to the legacy opt-in crypto/tls/fipsonly Go+BoringCrypto mechanism.)
```
I think we are safe.

[1] https://go.dev/doc/security/fips140

pkg/network/tlsSecurityProfile_test.go
ormergi: Good catch, done.

RamLavi (approved):
/lgtm
/approve

PR #2925, opened 2026-08-18
vendor: Bump go 1.26

ormergi:
This PR bump Go to 1.26

Along many improvements, newer Go version support additional TLS groups for post-quantum cryptography (PQC) readiness [1].
PQC readiness is crucial for our d/s product (Openshift).

The Go directive (go keyword in go.mod) is set to 1.26, omitting the path version, by the it set the minimal version of the project to 1.26.
The Go toolchain  directive (toolchain keyword in go.mod) is set and specifies the project new target version including the patch version.

Having go directive committing patch version together with toolchain directive, enable consumers use their own toolchain and patch versions.
Its safe because patch version skew should not break build, as long the consumer use minor version fitting the minimal version set (1.26).

[1] https://go.dev/doc/go1.26#:~:text=crypto/tls,post-quantum%20key%20exchanges

**Release note**:

```release-note
vendor: Bump go 1.26
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,8 @@
 module github.com/kubevirt/cluster-network-addons-operator
 
-go 1.25.7
+go 1.26
+
+toolchain go1.26.6
 
 tool (
 	github.com/github-release/github-release
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -120,7 +120,7 @@ make bump-kubevirtci       # Update kubevirtci version
 
 ## Build Toolchain
 
-- **Go version**: 1.25 (max allowed; auto-installed to `build/_output/bin/go/` via `hack/install-go.sh`)
+- **Go version**: 1.26 (max allowed; auto-installed to `build/_output/bin/go/` via `hack/install-go.sh`)
 - **Build flags**: `GOFLAGS=-mod=vendor GO111MODULE=on CGO_ENABLED=0`
 - **Multi-arch**: Builds for `linux/amd64`, `linux/arm64`, `linux/s390x`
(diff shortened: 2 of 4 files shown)


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/kubevirt/cluster-network-addons-operator
 
-go 1.25.7
+go 1.26.6
```
oshoval: this will break D/S
use please toolchain directive if you need 1.26.6
and here put 1.26.5

best if you can just use 1.26.5 meanwhile without the toolchain directive and wait for .6

if you end using toolchain, need to make sure the make vendor doesnt drop it, and the upstream scripts that read the version in order to know what to install, obey the toolchain if exists on higher priority

ormergi: Change to:
```
go 1.26

toolchain go1.26.6
```

The Go directive set the minimal version to 1.26, omitting the path version.
The Go toolchain directive is set and specifies the patch version for the project to work with,
Having the go toolchain allow consumers use their own toolchain and patch version.
Its safe because patch version skew should not break build, as long the consumer use minor version fitting the minimal version set (1.26).

oshoval: check lanes please, the scripts that read the lang need maybe 1.26.0 or to parse the toolchain instead, if exist (and now it does)

there might be more scripts, i dont remember (i did see you changed some)

iirc we have some logic that if toolchain exist take it, otherwise fall back to go line
so better to stick with it (it might be on the CNAO older branches etc)

EDIT - anyhow i saw you fixed it while we speak,
it might be good enough this way

oshoval:
Thanks

/lgtm
/approve


Ram is already mentioned, so fyi

PR #2938, opened 2026-08-24
Wire kubevirt-ipam-controller to the TLS secutiy profile's TLS groups

ormergi:
**Rebased on top #2923, see the last tree commits.**

**What this PR does / why we need it**:
This PR wire kubevirt-ipam-controller to the operator TLS security profile's TLS groups.

With this change TLS groups configured by `NetworkAddonsConfig` CR's `tlsSecurityProfile` passed down to kubevirt-ipam-controller app, via `tls-group-prefrences` flag. Similar to other TLS settings.

**Special notes for your reviewer**:

**Release note**:

```release-note
Wire kubevirt-ipam-controller to TLS security profile's TLS groups
```

PR #2941, opened 2026-08-26
Wire kubemacpool to the TLS security profile's TLS groups

RamLavi:
Pass NetworkAddonsConfig `tlsSecurityProfile` groups into kubemacpool as
`--tls-group-preferences`, the same way min TLS version and ciphers are already
wired. Needed so KMP webhook and metrics honor the selected TLS groups
(CNV-85518).

- `OCPTLSProfileTLSGroupNames` is also introduced in #2938. Whichever of these
  PRs merges second should drop the duplicate helper in conflict resolution.

**Release note**:
```release-note
Wire kubemacpool to TLS security profile TLS groups
```

### DIFF
--- a/pkg/network/tlsSecurityProfile_test.go
+++ b/pkg/network/tlsSecurityProfile_test.go
@@ -320,4 +320,38 @@ var _ = Describe("TLS Security Profile", func() {
 			Expect(ids).To(Equal([]tls.CurveID{tls.X25519}))
 		})
 	})
+
+	Context("OCPTLSProfileTLSGroupNames", func() {
+		It("should return nil for empty input", func() {
+			Expect(OCPTLSProfileTLSGroupNames(nil)).To(BeNil())
+			Expect(OCPTLSProfileTLSGroupNames([]ocpv1.TLSGroup{})).To(BeNil())
+		})
+		It("should convert OCP TLSGroup to names", func() {
+			names := OCPTLSProfileTLSGroupNames([]ocpv1.TLSGroup{
+				ocpv1.TLSGroupX25519MLKEM768,
+				ocpv1.TLSGroupX25519,
+				ocpv1.TLSGroupSecP256r1,
+				ocpv1.TLSGroupSecP384r1,
+				ocpv1.TLSGroupSecP521r1,
+				ocpv1.TLSGroupSecP256r1MLKEM768,
+				ocpv1.TLSGroupSecP384r1MLKEM1024,
+			})
+			Expect(names).To(Equal([]string{
+				"X25519MLKEM768",
+				"X25519",
+				"CurveP256",
+				"CurveP384",
+				"CurveP521",
+				"SecP256r1MLKEM768",
+				"SecP384r1MLKEM1024",
+			}))
+		})
+		It("should skip groups with no known crypto/tls.CurveID", func() {
+			names := OCPTLSProfileTLSGroupNames([]ocpv1.TLSGroup{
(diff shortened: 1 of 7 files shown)


ormergi (approved):
Thanks, 
LGTM

pkg/network/tlsSecurityProfile_test.go
```diff
@@ -320,4 +320,38 @@ var _ = Describe("TLS Security Profile", func() {
...
+				"X25519MLKEM768",
+				"X25519",
+				"CurveP256",
+				"CurveP384",
+				"CurveP521",
+				"SecP256r1MLKEM768",
+				"SecP384r1MLKEM1024",
+			}))
+		})
+		It("should skip groups with no known crypto/tls.CurveID", func() {
```
maiqueb, line 349: sure we want to silently skip ? wouldn't it be better to surface a warning to the user ?

maiqueb (changes requested):
few questions, requesting changes for visibility.

pkg/network/tlsSecurityProfile_test.go
RamLavi: @ormergi gave an explanation in a former PR. Or - can you remind me?

RamLavi: reason explained [here](https://github.com/k8snetworkplumbingwg/kubemacpool/pull/681#discussion_r3801950361)

maiqueb: Unfortunately, I don't understand that answer.

Could you elaborate ?

RamLavi: Sorry, that KMP comment was the wrong layer (Go's empty CurvePreferences default, not logging).

`OCPTLSProfileTLSGroupNames` is not parsing admin input. It maps CRD-validated OpenShift `TLSGroup` values to Go `crypto/tls` names. Unknown names cannot appear on a valid CR; the skip is for version skew (API enum grows before this mapping table does).

Same as ciphers in this file (`CipherSuiteIDs`), and what the API tells operators to do:

> Operators may remove entries their operands do not support.
> https://github.com/openshift/api/blob/master/config/v1/types_tlssecurityprofile.go#L207-L217

If some groups map, we pass only those. If none do, we omit the flag and Go uses defaults (API: omitted = no opinion). KMP applies the same skip on its side.

A warning would fire on a CNAO mapping gap the admin cannot fix without upgrading. Failing reconcile would also take the KMP webhook down. I'd keep this consistent with ciphers and with #2938.

maiqueb: All I'm saying is whenever you "If some groups map, we pass only those" or "If none do, we omit the flag and Go uses defaults" you're giving the user something which they didn't request (yes, you might be *correcting* their request). In those circumstances, I think we should at least log a warning, so the user can have a visible trail of WHY they have something which they didn't ask for, and also to help us debug scenarios where the user opens bugs based on "I asked for XYZ, but this is running ABC".

RamLavi: Fair enough. Added log

RamLavi:
[Change](https://github.com/kubevirt/cluster-network-addons-operator/compare/1c885c76fe06faf2911d665c6768e33a2e29bd22..b0485822da3f8e6e78fc1c78f0f41d1c6677aaad): Add log for omitted groups

maiqueb (approved):
/lgtm

Nice. Thanks for accommodating my request.

PR #2950, opened 2026-09-03
e2e/compliance,hack: shorten tls-compliance-operator scan/cleanup intervals

ormergi:
Recently we saw flakes on CI in the compliance e2e suite TLS tests, due to timeout of the assert on the tls-reports genereted by the tls-compliance-operator.
For example  https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_cluster-network-addons-operator/2948/pull-e2e-cluster-network-addons-operator-workflow-k8s/2095159128644128768

This PR reduce the tls-compliance-operator scan and cleanup intervals so tls-report CRs get updated more quickly.
This change should eliminate such flakes.

```release-note
NONE
```


oshoval (approved):
Thanks

RamLavi (approved):
/approve

